PR #135, opened 2022-12-06
feat: publish Hugo docs

Jcahilltorre:
### Proposed changes

Creates Hugo-based documentation for docs.nginx.com/nginx-agent based on the github readme.

Preview: https://deploy-preview-135--agent-public-docs.netlify.app/nginx-agent/


mtbChef (changes requested):
There's quite a bit here that's out of scope for my review. Now that the latest README PR has been merged, we ultimately want this content to mirror that version.

https://github.com/nginx/agent/blob/main/README.md

I did notice that some of the content is a bit stale. For example, the file docs/static/docs/nim-agent-00.png part of this PR is older. We are no longer using the NIM graphic. Please refer to the latest README for the updated Grafana graphic.

mtbChef (approved):

docs/content/dev-environment-setup.md
```diff
@@ -0,0 +1,54 @@
...
+```bash
+sudo apt install make
+```
+
+NGINX Agent is written in Go. To install Go, run:
+
+```bash
+sudo apt install golang-go
+```
+
```
Dean-Coakley, line 46: @oliveromahony Isn't this setup very limiting without
`make install-tools`? And maybe `sudo apt install protobuf-compiler` ?

mtbChef: This is what's in our readme. We also suggest downloading Go from the official website. Should we add the above packages there as well? It didn't throw an immediate flag for me because building and running from source worked in my vm.

PR #145, opened 2022-12-09
fix: detect nginx upgrade

nickchen:
### Proposed changes

In proc, if the process symlink path has `(deleted)`, it's an indication that the process was deleted. Since we already have binary change detection, this change deletes the "deleted" in the exec path, and trim out the "deleted" so the exec path is valid for the original binary, which could have an updated binary.

### Checklist

### DIFF
--- a/src/core/nginx.go
+++ b/src/core/nginx.go
@@ -132,12 +132,7 @@ func (n *NginxBinaryType) GetChildProcesses() map[string][]*proto.NginxDetails {
 }
 
 func (n *NginxBinaryType) GetNginxIDForProcess(nginxProcess Process) string {
-	defaulted := false
-	if nginxProcess.Path == "" {
-		nginxProcess.Path = defaultToNginxCommandForProcessPath()
-		defaulted = true
-	}
-
+	defaulted := n.sanitizeProcessPath(&nginxProcess)
 	info := n.getNginxInfoFrom(nginxProcess.Path)
 
 	// reset the process path from the default to what NGINX tells us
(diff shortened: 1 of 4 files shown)


yluf5:
I have tested the NAP upgrade case with this change and verified that it works.

dhurley:
Could you add a unit test for this?
Also could you run `make deps` as well?

src/core/nginx.go
```diff
@@ -160,13 +155,21 @@ func (n *NginxBinaryType) GetNginxDetailsByID(nginxID string) *proto.NginxDetail
...
 
-func (n *NginxBinaryType) GetNginxDetailsFromProcess(nginxProcess Process) *proto.NginxDetails {
+func (n *NginxBinaryType) sanitizeProcessPath(nginxProcess *Process) bool {
 	defaulted := false
 	if nginxProcess.Path == "" {
 		nginxProcess.Path = defaultToNginxCommandForProcessPath()
 		defaulted = true
 	}
+	if strings.Contains(nginxProcess.Path, execDeleted) {
+		log.Infof("nginx was upgraded (process), using new info")
```
dhurley, line 165: Could you remove this log line or change the log level here from info to debug? Its being repeated multiple times in the log file for the agent

nickchen:
> Could you add a unit test for this? Also could you run `make deps` as well?

done, added tests for sanitize function, did not expand the scope since it looks like require other setups.

dhurley (approved):

PR #147, opened 2022-12-12
Add install/uninstall integration test

Dean-Coakley:
### Proposed changes

* Convert install/uninstall to be an integration test
* Run install/uninstall test on push CI via `make integration-test` target
* Test no longer depends on `nginx` running
* Install/uninstall test now uses local package
* Test run against docker container. `ubuntu` is the only supported image at the moment.

### Checklist


Makefile
```diff
@@ -165,8 +166,13 @@ build-benchmark-docker: ## Build benchmark test docker image for NGINX Plus, nee
 		-f test/docker/Dockerfile .
 
 # Install tests
-test-install: ## Run agent install test
-	GOWORK=off CGO_ENABLED=0 go test -v ./test/install
+test-install: ## Run agent install/uninstall test for release
```
oliveromahony, line 169: this should probably roll up into an integration target, with multiple tests suites running against that and a new CI/CD stage

Makefile
```diff
@@ -165,8 +166,13 @@ build-benchmark-docker: ## Build benchmark test docker image for NGINX Plus, nee
...
+	AGENT_PACKAGE_FILE=${shell find ./build/packages/deb/*${OS_CODENAME}_${LOCAL_ARCH}.deb -type f -exec realpath {} \;} \
```
craigell, line 170: Think we should let the pipeline dictate where the AGENT_PACKAGE_FILE variable is?

.github/workflows/release-branch.yml
```diff
@@ -151,6 +151,8 @@ jobs:
         run: |
           echo "$INDIGO_GPG_AGENT" | base64 --decode > .key.asc
           make clean package
+      - name: Test install and uninstall
```
dhurley, line 154: This only needs to be added to the ci.yml file as that workflow will trigger the release.yml workflow

Makefile
```diff
@@ -180,11 +180,11 @@ performance-test: ## Run performance tests
 	$(CONTAINER_CLITOOL) run -v ${PWD}:/home/nginx/$(CONTAINER_VOLUME_FLAGS) --rm nginx-agent-benchmark:1.0.0
 
 integration-test: local-deb-package
-	PACKAGE=${PACKAGE_NAME} BASE_IMAGE=${BASE_IMAGE} go test ./test/integration/api
+	PACKAGE_NAME=${PACKAGE_NAME} BASE_IMAGE=${BASE_IMAGE} go test ./test/integration/install
+	PACKAGE_NAME=${PACKAGE_NAME} BASE_IMAGE=${BASE_IMAGE} go test ./test/integration/api
 
 test-bench: ## Run benchmark tests
 	cd test/performance && GOWORK=off CGO_ENABLED=0 go test -mod=vendor -count 5 -timeout 2m -bench=. -benchmem metrics_test.go
-	cd test/performance && GOWORK=off CGO_ENABLED=0 go test -mod=vendor -count 1 -bench=. -benchmem user_workflow_test.go
```
oliveromahony, line 187: why was this removed?

Dean-Coakley: Accidental. I think the install test used to have a similar filename a while back. 

Thanks for catching this. Reverting.

test/integration/install/install_uninstall_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+const (
+	osReleasePath  = "/etc/os-release"
+	maxFileSize    = int64(20000000)
+	maxInstallTime = 30 * time.Second
+)
+
+var (
+	AGENT_PACKAGE_FILENAME = os.Getenv("PACKAGE_NAME")
+	AGENT_PACKAGE_FILEPATH = "/agent/build/" + AGENT_PACKAGE_FILENAME + ".deb"
```
craigell, line 26: Should we hardcode the package extension? This might not always run on a Debian or Ubuntu environment.

test/docker/integration/install/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:22.04 as install
```
oliveromahony, line 1: can we reuse the OSS docker files and add in multi-stages, one for the operating system and NGINX. The second for the agent. You can target specific stages in the docker file through the cli: see the options` --stage ` https://docs.docker.com/engine/reference/commandline/build/#options and this can be done in compose in a similar way: https://stackoverflow.com/a/53101932/515970

test/integration/install/install_uninstall_test.go
Dean-Coakley: I should've stated in this PR that the scope was solely ubuntu as a starting point.

However this was easy to support so I decided to include in: https://github.com/nginx/agent/pull/147/commits/bb3607a7d4d2386822297270586274f800e8b4f7

aphralG (approved):

test/integration/install/install_uninstall_test.go
```diff
@@ -0,0 +1,217 @@
...
+		"AgentDynamicConfigFile": "/etc/nginx-agent/agent-dynamic.conf",
+	}
+
+	agentPackageFilePath := getPackagePath(string(osReleaseContent))
+
+	// Check the agent package is present
+	agentPkg, err := agentContainer.CopyFileFromContainer(context.Background(), agentPackageFilePath)
+	assert.NoError(t, err)
+
+	a, err := io.ReadAll(agentPkg)
```
dhurley, line 91: Could you give these variables better names instead of using single letters?

dhurley (commented):
Looks good just have a few minor comments

test/integration/install/install_uninstall_test.go
```diff
@@ -0,0 +1,217 @@
...
+	assert.NoError(t, err)
+
+	f, err := os.CreateTemp("", "tmpfile-")
```
oliveromahony, line 94: I don't get the need for lines 94-102? Should you be reading the package location and verifying that instead?

test/integration/install/install_uninstall_test.go
oliveromahony, line 91: Can you add a more descriptive variable name? Something that helps the code reader with the context

dhurley (approved):

oliveromahony (approved):

PR #157, opened 2022-12-19
feat: directory map actions

nickchen:
### Proposed changes

- use explicit action in the directory map to indicate deletes

### Checklist


src/core/nginx.go
```diff
@@ -406,6 +411,41 @@ func (n *NginxBinaryType) WriteConfig(config *proto.NginxConfig) (*sdk.ConfigApp
 	return configApply, nil
 }
 
+// generateDeleteFromDirectoryMap return a list of delete files from the directory map where Action File_delete is set.
+// This supports incremental upgrade if the files in the DirectoryMap doesn't have any action set to a non-default value,
+// in which the return bool will be false, to indicate explicit action is not set in the provided DirectoryMap.
+func generateDeleteFromDirectoryMap(
```
oliveromahony, line 417: This code is better moved to https://github.com/nginx/agent/blob/main/src/core/environment.go#L164 around here looks a good place: https://github.com/nginx/agent/blob/main/src/core/environment.go#L389 - this function probably should get renamed to SyncFiles then. This impacts the tests and Agent REST API, so changes will be required in those.

nickchen: i think that's significant more changes, since the file action is only populated in the directory map, and the passed in file is created from the zipped content. From what I can read, the flow seems to be: 
- files []*proto.File get created from zipped content https://github.com/nginx/agent/blob/main/src/core/nginx.go#L339
- then calls to the WriteFile

the directoryMap action doesn't distinguish between aux file or config file, so it's more appropriate to do the "delete" later. If we still want WriteFile to do the delete, I can do the generateDelete as it is, and pass in to the WriteFile? Should I do that instead?

oliveromahony: Ok, on review and based on your feedback we think that what you have now is less complicated, so we should stick with the original changes. We want to explicitly change this functionality in future and we have plans to do that. So to get this PR over the line please add more unit tests for the function you added.

nickchen: added tests

oliveromahony (approved):

dhurley (approved):

aphralG (approved):

PR #189, opened 2023-02-02
Create dedicated cache and upstream metrics reports

karlsassenberg:
Split Nginx Plus MetricsReport into four reports of the following types: CACHE_ZONE, UPSTREAMS, INSTANCE and SYSTEM. A single CACHE_ZONE report is created containing all StatsEntities with the dimension "cache_zone". A single UPSTREAMS report is created containing all StatsEntities with the dimension "upstream". The SYSTEM report contains all system metrics and the INSTANCE report contains all instance level metrics. Before only a single report was created of type SYSTEM which contained all metrics.

### Proposed changes

Split out plus.cache.* and plus.http.upstream.* metrics out from the generated MetricReport, each into their own dedicated MetricReport.
**Notes:**
Default value of metrics.mode in default.go was changed to "aggregated" instead of "aggregation" to match default nginx-agent.conf and internal code usages.
Empty MetricReports are suppressed.
Added two new types MetricsReport MetricsReport_CACHE_ZONE and MetricsReport_UPSTREAMS.

Refactored Existing tests and wrote new ones.
Manual testing done with the attached nginx conf nginx_caches_upstreams.conf.gz
where the logs were used to confirm the MetricReports were generated as expected and the analytics api was queried to confirm that the metrics were being persisted to clickhouse. Tested both streaming and aggregated metrics mode.

### Checklist


src/extensions/prometheus-metrics/prometheus_exporter_test.go
```diff
@@ -14,11 +15,11 @@ func TestExporter(t *testing.T) {
 
 	exporter := NewExporter(metricReport1)
 
-	assert.Equal(t, metricReport1, exporter.GetLatestMetricReport())
+	assert.Equal(t, metricReport1, exporter.GetLatestMetricReports()[0])
 
-	exporter.SetLatestMetricReport(metricReport2)
+	exporter.SetLatestMetricReport(&metrics.MetricsReportBundle{Data: []*proto.MetricsReport{metricReport2}})
```
dhurley, line 20: Can you update the unit test to have more than one report in the MetricsReportBundle?

test/integration/api/api_test.go
```diff
@@ -88,14 +88,13 @@ func TestAPI_Metrics(t *testing.T) {
 	resp, err := client.R().EnableTrace().Get(url)
 	assert.NoError(t, err)
 	assert.Equal(t, http.StatusOK, resp.StatusCode())
-	assert.Contains(t, resp.String(), "system_cpu_system")
+	//assert.Contains(t, resp.String(), "system_cpu_system")
```
dhurley, line 91: Can you add this back in?

test/component/agent_api_test.go
```diff
@@ -135,7 +136,7 @@ func TestMetrics(t *testing.T) {
 
 	agentAPI := plugins.NewAgentAPI(conf, mockEnvironment, mockNginxBinary)
 	agentAPI.Init(core.NewMockMessagePipe(context.TODO()))
-	agentAPI.Process(core.NewMessage(core.MetricReport, &proto.MetricsReport{
+	agentAPI.Process(core.NewMessage(core.MetricReport, &metrics.MetricsReportBundle{Data: []*proto.MetricsReport{{
```
dhurley, line 139: Can you update the component test to have more than one report in the MetricsReportBundle?

So like move the `system.cpu.idle` metric to a SYSTEM report and the `nginx.workers.count` to an INSTANCE report

src/core/metrics/metrics_util.go
```diff
@@ -292,3 +311,29 @@ func GetCalculationMap() map[string]string {
 		"container.mem.oom.kill":                             "avg",
 	}
 }
+
+func GenerateMetricsReports(entities []*StatsEntityWrapper) (bundles []core.Payload) {
```
dhurley, line 315: Should the return type here be `[]*MetricsReportBundle` ?

test/component/agent_api_test.go
karlsassenberg: wild

test/integration/api/api_test.go
karlsassenberg: will do

src/extensions/prometheus-metrics/prometheus_exporter_test.go
karlsassenberg: will do

src/core/config/defaults.go
```diff
@@ -71,7 +71,7 @@ var (
 			BulkSize:           20,
 			ReportInterval:     1 * time.Minute,
 			CollectionInterval: 15 * time.Second,
-			Mode:               "aggregation",
```
oliveromahony, line 74: this change seems to be unnecessary

karlsassenberg: In the code the keyword "aggregated", I thought only to bring the defaults into line with the code?

src/core/metrics/metrics_util.go
```diff
@@ -357,3 +376,29 @@ func GetCalculationMap() map[string]string {
...
```
dhurley, line 380: I think you can change the return type here to `bundle core.Payload`, since we always return an array with one payload in it.

Also a more descriptive name like `GenerateMetricsReportBundle` might be better.

test/integration/api/api_test.go
dhurley: think you may have forgot to add this back in

dhurley (commented):
Looks good. Just have a few minor comments

test/integration/api/api_test.go
karlsassenberg: thanks, done.

src/core/metrics/metrics_util.go
karlsassenberg: done.

dhurley (approved):
Looks good.
I verified that the metrics REST & gRPC APIs work as expected with this change.

craigell (approved):

PR #221, opened 2023-02-27
Fix plus docker files

aphralG:
### Proposed changes

Add `docker.io/` to `Dockerfiles` instead of in the `BASE_IMAGE` arg to stop Nginx Plus Dockerfiles from failing to build.

### Checklist

### DIFF
--- a/scripts/docker/nginx-plus/alpine/Dockerfile
+++ b/scripts/docker/nginx-plus/alpine/Dockerfile
@@ -1,5 +1,5 @@
 ARG BASE_IMAGE
-FROM ${BASE_IMAGE} as install
+FROM docker.io/${BASE_IMAGE} as install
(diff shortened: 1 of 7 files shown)


scripts/docker/nginx-plus/alpine/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BASE_IMAGE
-FROM ${BASE_IMAGE} as install
```
oliveromahony, line 2: I'm not sure if this change is correct looking at it. If we want to pull containers from a different registry we should be able to. I think the BASE_IMAGE variable in the Makefile should contain a registry variable and you can optionally default it to docker.io/

oliveromahony: or an additional CONTAINER_REGISTRY variable is ok too

PR #222, opened 2023-02-27
fix: uri and request parsing;

mohamed-gougam:
### Proposed changes

Right now we have a problem whenever the URI logged by NAP contains a comma `,`. That's due to the fact that we are separating logged values with comma `,`, so when splitting to get each value separately we fall in having erroneous values for both URI and request.

This fixes the problem as follow:

1. Under `/etc/app_protect/conf/log_default.json` add config to escape the comma `,` for each logged value with the URI encoding `%2C` (this represents standard URI encoding for `,` )
2. When parsing, for both `URI` and `request`, replace `%2C` with `,`.

### Checklist


src/plugins/nap_monitoring.go
```diff
@@ -55,7 +55,7 @@ func NewNAPMonitoring(env core.Environment, cfg *config.Config) (*NAPMonitoring,
 		log.Warnf("NAP Monitoring report count must be between %v and %v. Defaulting to %v",
 			minReportCountDelimiter,
 			maxReportCountDelimiter,
-			config.Defaults.NAPMonitoring.ReportInterval)
+			config.Defaults.NAPMonitoring.ReportCount)
```
sanathkumarbs, line 58: Good catch! Probably should add a unit test for things like this to catch and validate sooner.

sanathkumarbs (approved):
LGTM! We should make some Control Plane and Docs changes. I'll ping you about it.

src/plugins/nap_monitoring.go
olli-holmala: +1 on adding unit tests

src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -506,13 +505,13 @@ func setValue(napConfig *NAPConfig, key, value string, logger *logrus.Entry) err
...
 	case httpURI:
-		napConfig.HTTPURI = value
+		napConfig.HTTPURI = strings.ReplaceAll(value, encodedComma, decodedComma)
 	case isTruncated:
 		napConfig.IsTruncated = value
 	case policyName:
 		napConfig.PolicyName = value
 	case request:
-		napConfig.Request = value
+		napConfig.Request = strings.ReplaceAll(value, encodedComma, decodedComma)
```
olli-holmala, line 514: I have a question. So are the `HTTPURI` and `Request` fields fully URL encoded or just the commas? If they are fully encoded, should we be decoding them? The implementation we have now only decodes commas.
```go
request, err := url.QueryUnescape(value)
```

olli-holmala (approved):

src/extensions/nginx-app-protect/monitoring/processor/nap.go
mohamed-gougam: only the comma is encoded in `httpURI` and `request`

oliveromahony:
waiting on some unit tests before merging this

dhurley (approved):

oliveromahony (approved):

PR #241, opened 2023-03-08
deprecate-nginx-id-from-violation

mohamed-gougam:
### Proposed changes

We previously confirmed that it’s impossible to gather the NGINX ID for the NAP Violation being generated and sent to nms.

This follow up ticket is to deprecate the NGINX ID from both the violation struct being transferred.

### Checklist


sdk/proto/events/event.proto
```diff
@@ -115,7 +115,7 @@ message SecurityViolationEvent {
   string InstanceTags = 36 [(gogoproto.jsontag) = "instance_tags"];
   string InstanceGroup = 37 [(gogoproto.jsontag) = "instance_group"];
   string DisplayName = 38 [(gogoproto.jsontag) = "display_name"];
-  string NginxID = 39 [(gogoproto.jsontag) = "nginx_id"];
+  string NginxID = 39 [(gogoproto.jsontag) = "nginx_id", deprecated=true];
```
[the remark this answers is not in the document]
mohamed-gougam: applied 👍

olli-holmala (approved):

sdk/proto/events/event.proto
```diff
@@ -66,6 +66,10 @@ message ActivityEvent {
 
 // Represents a security violation that is emitted by the agent
 message SecurityViolationEvent {
+  // DEPRECATED
```
sanathkumarbs, line 69: question: was the protobuf bindings regenerated and the vendor too after the change from `deprecated` to the `reserved` field approach that you took?

sdk/proto/events/event.proto
```diff
@@ -66,6 +66,10 @@ message ActivityEvent {
...
+  reserved 39;
+  reserved "NginxID";
```
sanathkumarbs, line 71: question to @oliveromahony / @dhurley and the rest of the agent team - how do we add an RN note for this?

sanathkumarbs (approved):
LGTM, thank you for doing this!

sdk/proto/events/event.proto
mohamed-gougam: Yes, when used `deprecated` that reflected only documentation wise. When used `reserved` the field is not in the message anymore, can't be used or set. So yep the generated struct for `SecurityViolationEvent` did change.

sdk/proto/events/event.proto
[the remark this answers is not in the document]
sanathkumarbs: +1 to @dhurley's comment

sdk/proto/events/event.proto
sanathkumarbs: Thanks, @mohamed-gougam

sdk/proto/events/event.proto
sanathkumarbs: @mohamed-gougam, no need to wait on this to be resolved for the PR. You can follow up with a documentation update or an RN update. I'll resolve this thread for now, but let's figure this out as a follow on.

oliveromahony (approved):

dhurley (approved):

PR #265, opened 2023-03-22
refactor: extend security violation context parsing;

mohamed-gougam:
### Proposed changes

This extends the ability of NAP Monitoring extension in Agent to parse the XML Violations to gather the Context information and stream it to NGINX Management Suite.

### Checklist


src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -301,82 +320,128 @@ func (f *NAPConfig) getMetadata() (*models.Metadata, error) {
...
+				}
 				break
 			}
 
-			decodedName, err := base64.StdEncoding.DecodeString(v.Header.Name)
+			decodedName, err := base64.StdEncoding.DecodeString(name)
 			if err != nil {
-				logger.Errorf("could not decode the Header Name %s for %v", v.Header.Name, f.SupportID)
+				logger.Errorf("could not decode the Header Name %s for %v", name, f.SupportID)
 				break
```
olli-holmala, line 439: I know that this wasn't added in this PR, but are we sure that we want to `break` if we fail to base64 decode? I would think that we'd want to display the name and value even if there's a decoding issue, rather than returning an empty context?

mohamed-gougam: I don't think it's useful to display Base64 data that failed to be decoded, what do you think? It might create more confusion. The user might not even be aware that it's b64 encoded.

olli-holmala: Yeah I was thinking that there could be a case where it's got some other encoding that we haven't accounted for? But perhaps it's not a good a user experience to get encoded values in the front end 🤔

test/integration/advanced_metrics_test.go
```diff
@@ -12,12 +12,12 @@ import (
 	"testing"
 	"time"
 
+	conf "github.com/nginx/agent/test/integration/nginx"
+	"github.com/nginx/agent/test/integration/upstream"
+	"github.com/nginx/agent/test/integration/validator"
```
olli-holmala, line 17: I'm not quite sure why the path for these changed from `github.com/nginx/agent/v2` to `github.com/nginx/agent`. At least no tests broke as a result, and everything compiles normally.

olli-holmala (approved):
I think we are good to go. We've covered the violation categories that we have examples for in our unit and component tests.

dhurley (approved):

Dean-Coakley (approved):

craigell (approved):

oliveromahony (approved):

PR #288, opened 2023-04-12
add coverage to XML parsing in NAP Monitoring

mohamed-gougam:
### Proposed changes

This adds coverage to NAP Monitoring extension in Agent for parsing additional cases of the XML Violations to gather the Context information and stream it to NGINX Management Suite.

### Checklist


src/extensions/nginx-app-protect/monitoring/processor/processor_test.go
```diff
@@ -774,6 +774,222 @@ func TestNAPProcess(t *testing.T) {
 				ParentHostname: "",
 			},
 		},
+		{
```
olli-holmala, line 777: I cross-referenced our list of scoped violations against the violations found in this file and found that we are **missing** test cases for the following violations:

* `VIOL_PARAMETER_REPEATED`
* `VIOL_PARAMETER_MULTIPART_NULL_VALUE`
* `VIOL_PARAMETER_EMPTY_VALUE`
* `VIOL_URL`
* `VIOL_PARAMETER_VALUE_REGEXP`
* `VIOL_PARAMETER_NUMERIC_VALUE`
* `VIOL_PARAMETER_DATA_TYPE`
* `VIOL_PARAMETER_VALUE_LENGTH`
* `VIOL_PARAMETER_DYNAMIC_VALUE`
* `VIOL_PARAMETER_STATIC_VALUE`
* `VIOL_COOKIE_EXPIRED`
* `VIOL_PARAMETER_ARRAY_VALUE`
* `VIOL_PARAMETER_LOCATION`
* `VIOL_HEADER_REPEATED`

Let's please add all the XML examples from our table here as test cases so we can demonstrate that we parse them correctly.

olli-holmala:
I cross-referenced our list of scoped violations against the violations found in the `./test/component/nginx-app-protect/monitoring/testData/events-out` directory and found that we are missing test cases for the following violations:

* `VIOL_PARAMETER_REPEATED`
* `VIOL_PARAMETER_MULTIPART_NULL_VALUE`
* `VIOL_PARAMETER_EMPTY_VALUE`
* `VIOL_URL`
* `VIOL_PARAMETER_VALUE_REGEXP`
* `VIOL_PARAMETER_NUMERIC_VALUE`
* `VIOL_PARAMETER_DATA_TYPE`
* `VIOL_PARAMETER_VALUE_LENGTH`
* `VIOL_PARAMETER_DYNAMIC_VALUE`
* `VIOL_PARAMETER_STATIC_VALUE`
* `VIOL_COOKIE_EXPIRED`
* `VIOL_PARAMETER_ARRAY_VALUE`
* `VIOL_PARAMETER_LOCATION`

Let's please add all the XML examples from our table there as test cases so we can demonstrate that we parse them correctly.

src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -430,12 +437,25 @@ func (f *NAPConfig) getViolations(logger *logrus.Entry) []*models.ViolationData
...
+					value = v.Header.Value
+				} else {
+					value = v.Header.Text
+				}
 			} else if v.HeaderData != (Header{}) {
 				isB64Decoded = v.HeaderData.IsBase64Decoded
 				name = v.HeaderData.Name
 				value = v.HeaderData.Value
+			} else if v.HeaderLength != "" {
+				isB64Decoded = true
```
sanathkumarbs, line 451: question: is this a safe assumption to make that whenever we have HeaderLength, the Header name is B64 encoded?

src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -483,6 +503,15 @@ func (f *NAPConfig) getViolations(logger *logrus.Entry) []*models.ViolationData
 				}
 				name = v.SpecificDesc
 				value = string(decodedBuffer)
+			} else if v.CookieLength != "" {
+				isB64Decoded = true
```
sanathkumarbs, line 507: question: similar question as above

src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -515,13 +544,34 @@ func (f *NAPConfig) getViolations(logger *logrus.Entry) []*models.ViolationData
...
+				name = violationContextUri
 				value = v.Uri
 			} else if v.UriObjectData != (UriObjectData{}) {
+				name = violationContextUri
 				value = v.UriObjectData.Object
 			} else if v.UriLength != "" {
 				isB64Decoded = true
 				name = fmt.Sprintf("URI length: %s", v.UriLength)
 				value = fmt.Sprintf("URI length limit: %s", v.UriLengthLimit)
+			} else if v.HeaderData != (Header{}) {
```
sanathkumarbs, line 556: question: should this be Uri Data or Header Data? I ask this as it seems like we are parsing the URI Context information here.

test/component/nginx-app-protect/monitoring/testData/events-out/xml_violation_url_data.json
```diff
@@ -39,10 +39,18 @@
...
+                    "Name": "uri",
                     "Value": "/"
                 }
+            },
+            {
+                "Name": "VIOL_URL_CONTENT_TYPE",
+                "Context": "uri",
+                "ContextData": {
+                    "Name": "$",
+                    "Value": "actual header value: beni. matched header value: beni"
```
sanathkumarbs, line 51: question: it seems like we have a name containing the description at the time, for ex:
` "Name": "Header length: 42, exceeds Header length limit: 10",`

and, here we have the value containing the description:
`"Value": "actual header value: beni. matched header value: beni"`

Should we try to be consistent here?

src/extensions/nginx-app-protect/monitoring/processor/processor_test.go
mohamed-gougam: @olli-holmala These are just violation names that are missing. But all structures are covered, even structures of the listed names missing. Imo we should worry about structs variation for test cases not the names of violations (that would imply testing a different string values for violation name).

@sanathkumarbs thoughts please?

olli-holmala: Including the violations that are still missing would show:
* What we assumed, during time of implementation, the XML to look like for each violation
  * That way if something changes on the NAP side for XML structures (unlikely to happen) then we can easily see the delta between our assumed structure and actual structure
* That we are able to parse each violation and extract the correct key-value pair from each one
  * Even though we may be parsing key-value pairs from each violation correctly now, it's still important to demonstrate it
  * It would also show what we are **not** parsing for each violation
  * Demonstrating functionality via unit tests thoroughly is actually one of the PR requirements by the Agent team.

This XML parsing logic is essential to the violation contexts feature, that's why I think it deserves a bit more scrutiny.

src/extensions/nginx-app-protect/monitoring/processor/nap.go
mohamed-gougam: Based on all the examples we came across the assumption is correct.

src/extensions/nginx-app-protect/monitoring/processor/nap.go
mohamed-gougam: Based on all the examples we came across the assumption is correct. Param, URL, Cookie or Header is always b64 encoded.

test/component/nginx-app-protect/monitoring/testData/events-out/xml_violation_url_data.json
mohamed-gougam: Applied 👍

src/extensions/nginx-app-protect/monitoring/processor/processor_test.go
mohamed-gougam: I have added all the listed violation names listed above 👍

sanathkumarbs: I think both of you are right on different fronts; it's good to have a unit test for variations of the violation list, but it isn't critical (or a must) if they follow the same structure as the rest. 

The unit test should focus on parsing a set of combinations. However, to catch regressions and changes in XML implementations, I would highly recommend E2E tests, as we will likely have passing unit tests even if the XML structure changes underneath us. 

I don't disagree with the benefits listed above for having them as well.

sanathkumarbs (approved):
Overall everything is LGTM. Thanks for extending the XML support!

olli-holmala (approved):

oliveromahony (approved):

PR #300, opened 2023-04-21
Enable disable features

aphralG:
### Proposed changes

Added ability to enable and disable features 

### Checklist


Makefile
```diff
@@ -268,6 +268,7 @@ oss-image: ## Build agent container image for NGINX OSS
 	--build-arg BASE_IMAGE=${BASE_IMAGE} \
 	--build-arg ENTRY_POINT=./scripts/docker/entrypoint.sh
 
+
```
oliveromahony, line 271: this seems an unnecessary change

oliveromahony (approved):

src/core/pipe.go
```diff
@@ -89,6 +91,49 @@ func (p *MessagePipe) Register(size int, plugins []Plugin, extensionPlugins []Ex
 	return nil
 }
 
+func (p *MessagePipe) Deregister(pluginNames []string) error {
```
oliveromahony, line 94: DeRegister would be following our convention

src/plugins/features.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+func (f *Features) Process(msg *core.Message) {
+	log.Infof("--------> Process function in the features.go, %s %v", msg.Topic(), msg.Data())
+	switch data := msg.Data().(type) {
+	case string:
+		switch msg.Topic() {
+		case core.EnableFeature:
+			switch data {
```
oliveromahony, line 64: this could be refactored into a map

e.g. ... ```featureMap := map[string]func(){
	agent_config.FeatureMetrics: func() {
...```

then called like: 
```
if initFeature, ok := featureMap[data.(string)]; ok {
	initFeature()
}
```

src/plugins/features.go
```diff
@@ -0,0 +1,267 @@
...
+					nginx.Init(f.pipeline)
+
+				}
+			}
+
+		}
+	}
+}
+
+func (f *Features) isPluginAlreadyRegistered(pluginName string) bool {
```
oliveromahony, line 257: is this useful to add to pipe.go?

src/plugins/metrics_test.go
```diff
@@ -275,16 +290,6 @@ func TestMetrics_Process_AgentConfigChanged(t *testing.T) {
 	}
 }
 
-func TestMetrics_Process_RegistrationCompleted(t *testing.T) {
```
oliveromahony, line 278: was this removed intentionally? If so, should we move it rather than remove it?

src/plugins/metrics_test.go
```diff
@@ -295,16 +300,6 @@ func TestMetrics_Process_AgentCollectorsUpdate(t *testing.T) {
 	assert.True(t, pluginUnderTest.collectorsUpdate.Load())
 }
 
-func TestMetrics_Process_NginxPluginConfigured(t *testing.T) {
```
oliveromahony, line 298: was this removed intentionally? If so, should we move it rather than remove it?

dhurley: this was also intentionally removed. I removed the unnecessary dependency on the nginx plugin

src/plugins/features.go
aphralG: have the check in pipe.go instead ?

src/plugins/agent_api.go
```diff
@@ -271,10 +272,25 @@ func (h *NginxHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			response := AgentAPIConfigApplyStatusResponse{
+				CorrelationId: uuid.New().String(),
+				Message:       "unable to process NGINX config apply request as the nginx-config-async feature is disabled",
+				Status:        errorStatus,
+			}
+			err := writeObjectToResponseBody(w, response)
+			if err != nil {
+				log.Warn(err)
+			}
+			log.Error("Config Apply Feature Disabled")
```
oliveromahony, line 291: this should be a warning

src/plugins/commander.go
```diff
@@ -159,12 +162,60 @@ func (c *Commander) agentRegistered(cmd *proto.Command) {
...
+				}
+			}
 		}
+
 	default:
 		log.Debugf("unhandled command: %T", cmd.Data)
 	}
 }
 
+func (c *Commander) deregisterPlugin(data string) {
```
oliveromahony, line 196: deRegisterPlugin

src/plugins/features.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func (f *Features) Subscriptions() []string {
+	return []string{
+		core.EnableFeature,
+	}
+}
+
+func (f *Features) Process(msg *core.Message) {
+	log.Debugf("--------> Process function in the features.go, %s %v", msg.Topic(), msg.Data())
```
dhurley, line 270: Remove arrow in debug log

dhurley (approved):

src/plugins/features.go
oliveromahony: yeah, central place is good

PR #308, opened 2023-05-08
refactor: populate key-value depending on their content

mohamed-gougam:
### Proposed changes
**Context:** 
During the design phase of Violation Context backend APIs, a decision was made to improve the logic of how the violation contexts are populated in the Agent.

**Proposed solution:**
The UI design for the Violation Context tables contains the violation context key, violation context value and the count. If we look at the actual data for what these will contain, often there may be only the key or only the value (and not both) that are populated. Often in these cases fully understanding the key or value requires also the name of the violation itself. For example:

```
{
    "violation_name": "VIOL_HEADER_REPEATED",
    "violation_context": "header",
    "context_key": "Cookie",
    "context_value": ""
}
```

In this case, our API would return `context_key` = ‘Cookie’ and an empty `context_value` with `count` = 1. From the user's standpoint they are seeing that a cookie caused some violation, but it’s not directly apparent what was the actual failure - namely that the cookie was repeated. If `VIOL_HEADER_REPEATED` were returned in the response, then it would be fairly easy to understand.

Another example:

```
{
    "violation_name": "VIOL_PARAMETER_EMPTY_VALUE",
    "violation_context": "parameter",
    "context_key": "a",
    "context_value": ""
}
```

One way to fix the above is to always shift the key to `context_value` if `context_value` is empty and then populate `context_key` with the `violation_name`. So the Violation Context above would become:

```
{
    "context_key": "Parameter empty value",
    "context_value": "a",
    "count": 10
}
```

**Acceptance criteria:**

- Ensure that the `context_key` and `context_value` are actionable to the user when used together (the user will also have the `violation_context` available).
- Update/add unit and component tests

### Checklist


src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -336,38 +336,34 @@ func (f *NAPConfig) getMetadata() (*models.Metadata, error) {
 	return NewMetadata(t, f.SupportID)
 }
 
-func (f *NAPConfig) getViolationContext() string {
-	contexts := []string{}
+func (f *NAPConfig) extractViolationsContexts() {
 	if f.ViolationDetailsXML != nil {
 		for i, v := range f.ViolationDetailsXML.RequestViolations.Violations {
 			if v.Context != "" {
```
sanathkumarbs, line 342: minor: is this check still necessary?

sanathkumarbs (approved):
LGTM, thank you for adding the support to correctly populate the key-value and also for the additional test cases! 🙌

src/extensions/nginx-app-protect/monitoring/processor/nap.go
mohamed-gougam: Yep, it's still necessary as we don't want to change (extract from violation name) context if it's already been received within XML. Also as olli requested I have reverted back to concatenating and sending `violation_contexts`.

src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -624,6 +547,32 @@ func (f *NAPConfig) getViolations(logger *logrus.Entry) []*models.ViolationData
...
+	}
+	return
+}
+
+func violationNameToDataName(violationName string) string {
+	parts := strings.Split(strings.ToLower(violationName), string(violationNameSeparator))
+	if parts[0] != violationNamePrefix {
+		return strings.Join(parts, " ")
+	}
+	return strings.Join(parts[1:], " ")
```
olli-holmala, line 573: Should we convert these to title case? They would be slightly neater on the frontend imo.

```suggestion
	return strings.Title(strings.Join(parts[1:], " "))
```
The above only works if the violation name is in `strings.ToLower()` first.

olli-holmala (approved):

src/extensions/nginx-app-protect/monitoring/processor/nap.go
```diff
@@ -573,33 +491,40 @@ func (f *NAPConfig) getViolations(logger *logrus.Entry) []*models.ViolationData
...
+					logger.Errorf("could not decode the %s Name %s for %v", v.Context, name, f.SupportID)
+				} else {
+					decodedValue, err := base64.StdEncoding.DecodeString(value)
+					if err != nil {
+						logger.Errorf("could not decode the %s Value %s for %v", v.Context, value, f.SupportID)
+					} else {
+						name, value = populateNameValue(violation.Name, string(decodedName), string(decodedValue))
+						violation.ContextData = &models.ContextData{Name: name, Value: value}
+					}
+				}
```
Dean-Coakley, line 526: Is it possible to simplify this by flattening the error handling? (optional)

```suggestion
				decodedName, err := base64.StdEncoding.DecodeString(name)
				if err != nil {
					logger.Errorf("could not decode the %s Name %s for %v", v.Context, name, f.SupportID)
					return
				} 
				
				decodedValue, err := base64.StdEncoding.DecodeString(value)
				if err != nil {
					logger.Errorf("could not decode the %s Value %s for %v", v.Context, value, f.SupportID)
					return
				}
				
				name, value = populateNameValue(violation.Name, string(decodedName), string(decodedValue))
				violation.ContextData = &models.ContextData{Name: name, Value: value}
```

Dean-Coakley (approved):

src/extensions/nginx-app-protect/monitoring/processor/nap.go
Dean-Coakley: If you want to only decode value if the `name` can be decoded and need to `populateNameValues`/`ContextData` unconditionally this might not be possible. Feel free to ignore

mohamed-gougam: Thanks for the comment Dean, yep that's what I want to achieve `only decode value if the name can be decoded and need to populateNameValues/ContextData unconditionally`. I will resolve the issue

craigell (approved):

Issue #317, opened 2023-05-18
NGINX agent cannot find the default network interface on Linux hosts.


richbrowne:
Description:
The NGINX agent is incorrectly reporting the loopback interface as the default destination for network traffic.  
When the agent reads `/proc/net/route` it always skips over the first line of route data.  This causes it to miss the default route in many cases and it will report the default interface as loopback.

Reproduce:
Run the agent on any Linux host with debug logging level.  You will see the error message:
time="2023-05-18T05:25:01Z" level=debug msg="Error getting default network interface, interface with default destination not found"

Expected Result:
There is no error in the agent log about not being able to find the default destination and the agent reports the correct default interface. 

Actual Result:
There is an error message about not finding the default destination and the agent reports 'lo' as the default interface.

dekobon:
Looking at [`network.go`](https://github.com/nginx/agent/blob/main/src/core/network/network.go#L298), it appears that we skip the first two lines of `/proc/net/route`, when we should just be skipping a single line. See below.
```go
func parseToLinuxRouteStruct(output []byte) (routeStruct, error) {
	const (
		destinationField = 1 // field containing hex destination address
	)
	lineNumber := 0
	scanner := bufio.NewScanner(bytes.NewReader(output))

	// Skip header line
	if !scanner.Scan() {
		return routeStruct{}, errors.New("Invalid linux route file")
	}

	for scanner.Scan() {
		lineNumber++
		if lineNumber == 1 {
			// Skip header line.
			continue
		}
		row := scanner.Text()
...
```

PR #318, opened 2023-05-18
NGINX agent cannot find the default network interface on Linux hosts.…

richbrowne:
… (#317)

The parser was skipping over the first (header) line with the first invocation of scanner.Scan().  The second invocation of scanner.Scan() called in the for loop points to the first line of route data; however, this line was being skipped.  The tests were incorrect because they started with a newline before the header.

### Proposed changes
Remove the logic to skip over the first line of route data for Linux hosts and corrected the unit tests.

Tested the change on Ubuntu 20.04.5 LTS

### Checklist


richbrowne:
@nkashiv @achawla2012 Could you review too and get this pushed through?

achawla2012:
> @nkashiv @achawla2012 Could you review too and get this pushed through?

The default interface is the one that has 0 for both destination and mask.  Therefore we will also have to check if mask is 0 as well. https://github.com/jackpal/gateway/blob/master/gateway_parsers.go#L69

achawla2012:
LGTM

Dean-Coakley:
Fixed in https://github.com/nginx/agent/pull/331

PR #331, opened 2023-06-05
fix: fix Linux default network interface parsing

### LINKED ISSUE #317: NGINX agent cannot find the default network interface on Linux hosts.

dekobon:
Fixes #317

This fix resolves a problem where the header line of /proc/net/route is skipped twice when running on Linux. This change also adds error checking to the Scanner performing the parsing operation.

### Checklist

- [X] If applicable, I have updated any relevant documentation ([`README.md`](https://github.com/nginx/agent/blob/main/README.md))
- [X] If applicable, I have tested my cross-platform changes on Ubuntu 22, Redhat 8, SUSE 15 and FreeBSD 13

### DIFF
--- a/src/core/network/network.go
+++ b/src/core/network/network.go
@@ -292,20 +292,21 @@ func parseToLinuxRouteStruct(output []byte) (routeStruct, error) {
 	const (
 		destinationField = 1 // field containing hex destination address
 	)
-	lineNumber := 0
 	scanner := bufio.NewScanner(bytes.NewReader(output))
 
 	// Skip header line
 	if !scanner.Scan() {
-		return routeStruct{}, errors.New("Invalid linux route file")
+		if scanner.Err() == nil {
+			return routeStruct{}, errors.New("invalid linux route file: no header line")
+		} else {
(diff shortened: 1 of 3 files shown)


dekobon:
Apparently, I'm blind and I did not see this PR: https://github.com/nginx/agent/pull/318

aphralG (approved):

Dean-Coakley:
@dekobon Can you take a look at https://github.com/nginx/agent#installing-prerequisite-packages, then give `make deps` a run and commit the result?

Dean-Coakley (approved):

PR #338, opened 2023-06-07
Fix segmentation fault during agent registration

achawla2012:
### Proposed changes

- Add nil check around feature object.
- Add unit tests during agent registration

### Checklist

- [X] If applicable, I have updated any relevant documentation ([`README.md`](https://github.com/nginx/agent/blob/main/README.md))


src/plugins/commander.go
```diff
@@ -163,11 +163,14 @@ func (c *Commander) agentRegistered(cmd *proto.Command) {
 				}
 			}
 
-			for index, feature := range agtCfg.Details.Features {
-				agtCfg.Details.Features[index] = strings.Replace(feature, "features_", "", 1)
+			if agtCfg.Details != nil && agtCfg.Details.Features != nil {
```
ryepup, line 166: nit: it's safe to `range` over a nil slice ([demo](https://go.dev/play/p/dADYKu93P57)), we can omit one of these checks

```suggestion
			if agtCfg.Details != nil {
```

src/plugins/commander.go
```diff
@@ -163,11 +163,14 @@ func (c *Commander) agentRegistered(cmd *proto.Command) {
...
+				for index, feature := range agtCfg.Details.Features {
+					agtCfg.Details.Features[index] = strings.Replace(feature, "features_", "", 1)
+				}
+
+				sort.Strings(agtCfg.Details.Features)
 			}
 
-			sort.Strings(agtCfg.Details.Features)
 			sort.Strings(c.config.Features)
```
achawla2012, line 174: Need feedback on `sort.Strings(c.config.Features)`.

I have not added nil check around it with assumption `c.config.Features` won't be null.
Is my assumption true?

ryepup: I think we're OK on this line,`sort.Strings` works fine if given a `nil`: https://go.dev/play/p/91Jt0vRD9vv

The next few lines look like more potential nil pointer derefs if `agtCfg.Details` is nil. Maybe we make one larger `if agtCfg.Details != nil {` block and pull everything down to L189 into it? Seems like this code is all about synchronizing `agtCfg.Details.Features` and `c.config.Features`, and we don't need to do any of that work if we don't have any `agtCfg.Details`.

It's a little tempting to introduce a helper methods here, so we could have something simple in `agentRegistered`:

```go
if agtCfg.Details != nil && agtCfg.Details.Features != nil {
  c.synchronizeFeatures(agtCfg.Details.Features)
}
```

achawla2012: At first I thought the same, but agent uses `agentConfig` as a way to see if management and data plane are sync. It's likely if `agtCfg.Details.Features` is nil but `c.config.Features` not. Agent would flag such setting as un-synchronized. There fore take steps to synchronize.

If we go by above approach, synchronization would then only happen when `agtCfg.Details.Features` is non-nil.

cc @oliveromahony

ryepup: hrm, just noticed that L141 kinda looks like it's doing a lot of the same work?

achawla2012: Oh yeah, it's sth that is taken care of before this. I will refactor this a bit. Thanks for pointing this.

oliveromahony (approved):

Dean-Coakley (approved):

oliveromahony (approved):

dhurley (approved):

PR #346, opened 2023-06-12
Fix access log parsing for custom log format

nkashiv:
### Proposed changes

This PR aims to fix the bug where if a user has an access log format with a variable that is not present in the `logVarMap` https://github.com/nkashiv/agent/blob/main/src/core/metrics/sources/nginx_access_log.go#L33 then the access log parsing fails and no access log metric is collected. 

For example, the following log format will cause an error because it contains `$upstream_addr` which is not present in the map. A user can also have a custom variable defined which will break the parsing. 
```
log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for" '
                      '"$upstream_addr" "$upstream_response_time" "$upstream_status" '
                      '$http_x_amplify_id';
```
Added a unit test in `nginx_access_log_test.go` to demonstrate this fix.

### Checklist

### DIFF
--- a/src/core/metrics/sources/nginx_access_log.go
+++ b/src/core/metrics/sources/nginx_access_log.go
@@ -199,6 +199,19 @@ func (c *NginxAccessLog) logStats(ctx context.Context, logFile, logFormat string
 	for key, value := range logVarMap {
 		logPattern = strings.ReplaceAll(logPattern, key, value)
 	}
+	log.Debugf("LogPattern = %s", logPattern)
+	r, err := regexp.Compile(`[\$]([a-z_]+)`)
+	if err != nil {
+		log.Warnf("unable to compile access log regex: %v", err)
+	}
+
+	variables := r.FindAllString(logPattern, -1)
+
+	log.Debugf("LogPattern = %s Matched variables = %v", logPattern, variables)
+	for _, variable := range variables {
+		replacement := "%" + fmt.Sprintf("{DATA:%s}", strings.Trim(variable, "$"))
+		logPattern = strings.Replace(logPattern, variable, replacement, -1)
+	}
 
 	log.Debugf("Collecting from: %s using format: %s", logFile, logFormat)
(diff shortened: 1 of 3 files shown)


src/core/metrics/sources/nginx_access_log.go
```diff
@@ -622,3 +628,18 @@ func getDefaultCounters() (map[string]float64, map[string]float64, map[string]fl
 
 	return httpCounters, upstreamCounters, upstreamCacheCounters
 }
+
+func replaceCustomLogVars(logPattern string) string {
+	// For all the variables in the log format that are not present in the logVarMap
+	// replace them with the %{DATA:.*} format
```
achawla2012, line 634: Should these comment to move out at function level?

achawla2012 (approved):
LGTM

src/core/metrics/sources/nginx_access_log.go
```diff
@@ -56,6 +56,10 @@ var logVarMap = map[string]string{
 	"]":                         "\\]",
 }
 
+// Pattern to match all the variables that are mentioned in the access log
+// and are absent from logVarMap
+var logVarRegex = regexp.MustCompile(`\$([a-zA-Z]+[_[a-zA-Z]+]*)`)
```
ryepup, line 61: nit: the regex pattern itself doesn't exclude things from the `logVarMap`, that bit of docs might be misleading

```suggestion
// Pattern to match all the variables that are mentioned in the access log
// format
var logVarRegex = regexp.MustCompile(`\$([a-zA-Z]+[_[a-zA-Z]+]*)`)
```

ryepup (approved):

aphralG (approved):

oliveromahony (approved):

Dean-Coakley (approved):

craigell (approved):

PR #350, opened 2023-06-20
fix: use host architecture to determine FreeBSD ABI

defanator:
### Proposed changes

This change allows to get working FreeBSD agent packages on both x86 and arm64 hosts/runners when running `make local-txz-package` target.

### Checklist

- [x] If applicable, I have tested my cross-platform changes (verified manually on FreeBSD 13.2 amd64/aarch64)

### DIFF
--- a/scripts/packages/packager/local-entrypoint.sh
+++ b/scripts/packages/packager/local-entrypoint.sh
@@ -9,8 +9,6 @@ case "$(uname -m)" in
     arm64|aarch64) ABIARCH=aarch64 ;;
 esac
 
-FREEBSD_DISTROS="FreeBSD:12:${ABIARCH} FreeBSD:13:${ABIARCH}"
-
 cd /nginx-agent/
 
 mkdir -p /staging/usr/local/bin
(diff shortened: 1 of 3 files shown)


oliveromahony:
We don't support arm in FreeBSD in NGINX Plus see: https://docs.nginx.com/nginx/technical-specs/

Is this PR needed?

defanator:
@oliveromahony uhm not sure what's the relationship with N+ here; I was working on the fix in #352 and realized that NFPM produces broken .pkg/.txz on M1 macbook. This PR fixes this in a way when you'll get same arch package on any host.

oliveromahony:
We generally deploy on the same host/container as NGINX. We use Plus as the reference in our support matrices we test against. So for us, supporting arm64 isn't required for the packages we build. The local changes I'm ok with if you want to test on Mac M1/2 but we won't produce official arm64 packages for FreeBSD for the moment unless Plus supports that architecture

scripts/packages/packager/local-entrypoint.sh
```diff
@@ -4,6 +4,13 @@ set -e
 set -x
 set -euxo pipefail
 
+case "$(uname -m)" in
```
oliveromahony, line 7: we already have OSARCH variable in the root Makefile you can leverage. It should take from that value instead of adding more logic here

scripts/packages/packager/local-entrypoint.sh
```diff
@@ -24,17 +31,21 @@ chmod +x /staging/usr/local/etc/rc.d/nginx-agent
...
-    -p /staging/plist \
-    -o ./build/; \
-
-# Creating symbolic link from txz to pkg. In older versions of pkg the extension would represent the format of the file 
-# but since version 1.17.0 pkg will now always create a file with the extesion pkg no matter what the format is. 
-# See 1.17.0 release notes for more info: https://cgit.freebsd.org/ports/commit/?id=e497a16a286972bfcab908209b11ee6a13d99dc9
-cd build/
-ln -s nginx-agent-"$(git describe --match 'v[0-9]*' --abbrev=0 | tr -d 'v')"-SNAPSHOT-"$(git rev-parse --short HEAD)".pkg nginx-agent-"$(git describe --match 'v[0-9]*' --abbrev=0 | tr -d 'v')"-SNAPSHOT-"$(git rev-parse --short HEAD)".txz 
-cd ..
+for freebsd_abi in $FREEBSD_DISTROS; do
```
dhurley, line 34: For local packages, we typically build one of each type e.g. deb, rpm, apk, etc. We don't build all possible versions. Could we remove the for loop and just build one package but this time using the new `ABIARCH` variable you have defined?
i.e.
```
pkg -o ABI="FreeBSD:13:${ABIARCH}" create --format txz \
....
```
That way we don't need to add the additional `cp` command in the ci.yml file.

scripts/packages/packager/local-entrypoint.sh
defanator: Uhm I don't see any relation between Makefile arg(s), which are not even being exported, and entrypoint scripts. Moreover, with multiarch containers (e.g. buildx) you'll get broken entrypoint for archs not matching host's arch if you will be relying on OSARCH obtained on host. Finally, OSARCH is being weirdly transformed from `uname -m` (introduced here - https://github.com/nginx/agent/commit/24a158b319da1a9acd5bf52eb119d93a327810a5#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L36), so I wouldn't rely on that at all for the sake of simplicity.

scripts/packages/packager/local-entrypoint.sh
defanator: Makes sense. Will do shortly.

defanator: (I did it that way to reduce a diff between "local" and "signed" scripts - just to explain the rationale.)

defanator:
> but we won't produce official arm64 packages for FreeBSD for the moment unless Plus supports that architecture

@oliveromahony roger that - I'm not asking you to do so; but if/when N+ will ship aarch64 binaries for FreeBSD, agent changes will be minimal with this PR (basically, you'll just need to use full path with ABI/arch component for resulting .pkgs, that's all).

scripts/packages/packager/signed-entrypoint.sh
```diff
@@ -4,7 +4,12 @@ set -e
 set -x
 set -euxo pipefail
 
-FREEBSD_DISTROS="FreeBSD:12:amd64 FreeBSD:13:amd64"
+case "$(uname -m)" in
```
dhurley, line 7: Can we undo the changes here just for the signed-entrypoint.sh? 
The changes for the local-entrypoint.sh are fine but I think if are to support FreeBSD aarch64 packages in the future we will need to make updates here https://github.com/nginx/agent/blob/main/Makefile.packaging#L165 to the Makefile.packaging to run the docker container twice. Once for aarch64 and once for amd64. Passing in a different binary build each time.

defanator: @dhurley sure, will do.

(I'm struggling to understand how it would help though given that goreleaser/NFPM is able to produce multiarch binaries on any host arch, and you can reuse the same binaries in "signed" builder in exactly the same way as in "local" builder.)

defanator: @dhurley latest commit (https://github.com/nginx/agent/pull/350/commits/5e8c0917af5d27bf2b6b4b7f0bfc994b02e9cc41) reverted all functional changes in signed-entrypoint.sh back to its original state - isn't that that you requested?

dhurley (approved):

aphralG (approved):

scripts/packages/packager/signed-entrypoint.sh
dhurley: oh yes you're correct sorry I missed that. Looks good now.

oliveromahony (approved):

PR #352, opened 2023-06-21
fix: avoid stopping nginx-agent service on package upgrade

### LINKED ISSUE #303: Uninstalling agent duplicates messages
**Description:**
When uninstalling the agent, the output of the uninstall displays duplicate messages. 

**Reproduce:**
Uninstall agent using `sudo apt purge nginx-agent`

**Expected Result:**
Logs show no duplicate messages.

**Actual Result:**
Logs show duplicate messages, see output below.

Output of  `sudo apt purge nginx-agent`:
```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  nginx-agent*
0 upgraded, 0 newly installed, 1 to remove and 61 not upgraded.
After this operation, 26.2 MB disk space will be freed.
Do you want to continue? [Y/n] y
(Reading database ... 68488 files and directories currently installed.)
Removing nginx-agent (2.25.0) ...
Stop and disable nginx-agent service
Running daemon-reload
Removing run directory
(Reading database ... 68486 files and directories currently installed.)
Purging configuration files for nginx-agent (2.25.0) ...
Stop and disable nginx-agent service
Running daemon-reload
Removing run directory
dpkg: warning: while removing nginx-agent, directory '/var/log/nginx-agent' not empty so not removed
dpkg: warning: while removing nginx-agent, directory '/etc/nginx-agent' not empty so not removed
```

defanator:
### Proposed changes

This change introduces a number of improvements around packaging nginx-agent to ensure that the following behavior is maintained:

1. The `nginx-agent` service is called with stop argument on package removal.
2. The `nginx-agent` service is called with restart argument on package upgrade if a service is running at the moment of package upgrade (previously, a service has just been stopped without any notice).
3. On fresh installations, or on upgrades when `nginx-agent` service from older package is not running, it will not be started automatically.

While here, added OpenRC init scripts to make `service nginx-agent` command fully functional on Alpine Linux.

Tested on:
 - Debian 11 (.deb)
 - RHEL 8 (.rpm)
 - Alpine 3.18 (.apk)

Unfortunately, FreeBSD is failing to follow the 2nd rule (restart on upgrade if a service was running) due to limitation of the pkg package manager (see https://github.com/freebsd/pkg/pull/2128 for more details).

Fixes #303.

### Checklist


defanator:
Tagging @thresheek @fblr @oxpa for the sake of visibility.

dhurley (commented):
I think the addition of openrc for alpine looks good, but I'm not sure about changing the current behaviour of the agent during upgrade. We require the agent to be stopped by the user before upgrade and then started again after the upgrade by the user.

scripts/packages/nginx-agent.openrc
```diff
@@ -0,0 +1,15 @@
+#!/sbin/openrc-run
+
+description="nginx-agent"
```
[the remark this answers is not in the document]
defanator: @dhurley done in https://github.com/nginx/agent/pull/352/commits/797044d419a487aef3971f4cd0feb704bf75f2fb

scripts/packages/postinstall.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+# vim:sw=4:ts=4:et:
```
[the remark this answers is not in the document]
defanator: @dhurley done in https://github.com/nginx/agent/pull/352/commits/797044d419a487aef3971f4cd0feb704bf75f2fb

dhurley:
> > We require the agent to be stopped by the user before upgrade and then started again after the upgrade by the user.
> 
> @dhurley this is quite counterintuitive and stumbling behavior; what was the rationale behind that decision?
> 
> To elaborate on a context: we came across the situation when nginx-agent service silently stopped during the usual package upgrade process with the appropriate package manager (yum/dnf, apt, apk) on a large fleet of instances. There were no any errors from corresponding package scripts during the upgrade, and there were no any signs of broken agent's configuration etc - so from the administrator's point of view, upgrade went fine - except the fact that we ended up with stopped monitoring for all affected instances.
> 
> I'm wondering where that requirement of manual stopping and starting came from. Agent's process can be restarted gracefully, so why prevent it to do so?

Hi @defanator, actually it looks like our docs might be a bit out of date. Yes you actually shouldn't need to stop the agent before upgrading. When we upgrade the agent, the old version of the agent should continue to run and its not until the user restarts the agent that the new version of the agent is used. Sorry about the confusion.

scripts/packages/postremove.sh
```diff
@@ -1,21 +1,68 @@
 #!/bin/sh
+
 # Determine OS platform
 # shellcheck source=/dev/null
 . /etc/os-release
 
-if [ "$ID" = "freebsd" ]; then
-    echo "Stop and remove nginx-agent service"
+_stop_agent_freebsd() {
```
dhurley, line 7: Could you remove the underscore prefix from these function names so that it is consistent with the naming convention in other script.

scripts/packages/postupgrade.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
dhurley, line 1: Just trying to understand why we are doing things differently here when the OS is alpine?
Is there a way to keep the old version of the agent running and let the user restart the agent afterwards when they want to switch versions?

dhurley (commented):
Thanks for fixing the issue where the agent was being stopped during upgrades. I just have questions related to upgrading on alpine.

defanator:
> When we upgrade the agent, the old version of the agent should continue to run and its not until the user restarts the agent that the new version of the agent is used.

@dhurley such behavior will introduce an inconsistent setup when your running binary will no longer be available on disk, which is potentially just another _delayed_ point of failure (aka time bomb): imagine a situation when users won't restart service after upgrade till system reboot or another component(s) upgrade that will involve restarting services affected by particular changed shared library - which is a common scenario in modern Ubuntu/Debian - in this situation, in case of any unexpected issues with agent, it will be impossible to just restart the service without doing rollback to previous package version.

Restarting service on upgrade (and failing immediately if a restart was unsuccessful) solves this inconsistency + allows a user to be alarmed on possible issues as soon as possible.

We spent a while building similar patterns for other nginx products (just noticing - given the fact agent's trying to follow established principles / support matrix of nginx/nginx-plus); I'm happy to elaborate more on this if required.

scripts/packages/postupgrade.sh
defanator: @dhurley alpine is following exactly the same pattern here as RHEL/Debian; it just has slightly different schema of package scripts, hence some actions were moved to another specific stages. The way agent is using to build linux packages is quite confusing and obfuscating (yes, I mean NFPM) - I strongly recommend to consider moving pkg building to "natural" tooling (rpmbuild/debuild/abuild, etc), that will resolve many ambiguities that are currently in place with NFPM.

scripts/packages/postremove.sh
defanator: @dhurley done in https://github.com/nginx/agent/pull/352/commits/9bb401d7571bf798aeb6607b5d5b439de51f7364

bitcoinfinancier (approved):

defanator:
@dhurley @oliveromahony any updates on this one?

defanator:
TWIMC, rebased this against latest `main` and tested on these platforms:
 - Debian 11 "bullseye",
 - Alpine 3.18,
 - RHEL 9.3.

dhurley (approved):

oliveromahony (approved):

PR #353, opened 2023-06-23
Add Dockerfiles & docker-compose files for official OSS & Plus images

dhurley:
### Proposed changes

Add official Dockerfiles for NGINX OSS & Plus images that have the NGINX Agent installed on them as well.

### Checklist

### DIFF
--- a/scripts/docker/official/nginx-oss-with-nginx-agent/alpine/Dockerfile
+++ b/scripts/docker/official/nginx-oss-with-nginx-agent/alpine/Dockerfile
@@ -1,0 +1,21 @@
+ARG NGINX_VERSION=1.25.1-alpine
+FROM nginx:${NGINX_VERSION}
+LABEL maintainer="NGINX Agent Maintainers <agent@nginx.com>"
+
+RUN set -x \
+    && apk add curl bash iproute2
+
+RUN printf "%s%s\n" \
+    "http://packages.nginx.org/nginx-agent/alpine/v3.16" \
+    "/main" | tee -a /etc/apk/repositories
+
+RUN curl -o /etc/apk/keys/nginx_signing.rsa.pub https://nginx.org/keys/nginx_signing.rsa.pub
+
+RUN apk add nginx-agent
+
+COPY ./scripts/docker/entrypoint.sh /agent/entrypoint.sh
+RUN chmod +x /agent/entrypoint.sh
+
+STOPSIGNAL SIGTERM
+
+ENTRYPOINT ["/agent/entrypoint.sh"]
--- a/scripts/docker/official/nginx-oss-with-nginx-agent/alpine/Dockerfile
+++ b/scripts/docker/official/nginx-oss-with-nginx-agent/alpine/Dockerfile
@@ -2,6 +2,10 @@ ARG NGINX_VERSION=1.25.1-alpine
 FROM nginx:${NGINX_VERSION}
 LABEL maintainer="NGINX Agent Maintainers <agent@nginx.com>"
 
+# Unlink symbolic links for access and error logs as the NGINX Agent 
+# requires both files to be able to monitor NGINX and provide metrics.
+RUN unlink /var/log/nginx/access.log && unlink /var/log/nginx/error.log
+
 RUN set -x \
     && apk add curl bash iproute2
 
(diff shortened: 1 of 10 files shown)


scripts/docker/official/nginx-oss-with-nginx-agent/alpine/Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG NGINX_VERSION=1.25.1-alpine
+FROM nginx:${NGINX_VERSION}
+LABEL maintainer="NGINX Agent Maintainers <agent@nginx.com>"
+
+RUN set -x \
+    && apk add curl bash iproute2
+
+RUN printf "%s%s\n" \
+    "http://packages.nginx.org/nginx-agent/alpine/v3.16" \
+    "/main" | tee -a /etc/apk/repositories
```
Dean-Coakley, line 10: Makes sense to do something like the following? 

So when `ARG NGINX_VERSION` is modified and includes a alpine version change, the correct agent version will always be installed?

```suggestion
RUN printf "%s%s%s\n" \
    "http://packages.nginx.org/nginx-agent/alpine/v" \
    `grep -o -E '^[0-9]+\.[0-9]+' /etc/alpine-release` \
    "/main" | tee -a /etc/apk/repositories
```

Dean-Coakley (approved):

scripts/docker/official/nginx-plus-with-nginx-agent/alpine/Dockerfile
```diff
@@ -0,0 +1,68 @@
+FROM alpine:3.17
```
defanator, line 1: Any reason not to use official nginx/foss alpine image as a base here? It should be pretty straightforward and would make future maintenance easier.

dhurley: This Dockerfile is for NGINX Plus where there isn't an official image. We used the reference here to build an example Dockerfile that people can use if they are using NGINX Plus. https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-docker/#creating-nginx-plus-docker-image 

We do use the docker image for NGINX OSS alright here for this other Dockerfile we have https://github.com/nginx/agent/blob/fc8606e8ad47b83a830a77c067d52116d86f7bae/scripts/docker/official/nginx-oss-with-nginx-agent/alpine/Dockerfile

scripts/docker/official/nginx-plus-with-nginx-agent/alpine/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+        echo "key verification succeeded!"; \
+        mv /tmp/nginx_signing.rsa.pub /etc/apk/keys/; \
+    else \
+        echo "key verification failed!"; \
+        exit 1; \
+    fi \
+    && cat cert.pem > /etc/apk/cert.pem \
+    && cat cert.key > /etc/apk/cert.key \
+    && apk add -X "https://pkgs.nginx.com/plus/alpine/v$(egrep -o '^[0-9]+\.[0-9]+' /etc/alpine-release)/main" --no-cache $nginxPackages \
+    && apk add -X "https://pkgs.nginx.com/nginx-agent/alpine/v3.16/main" --no-cache nginx-agent \
```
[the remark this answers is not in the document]
dhurley: Yeah at the moment v3.17 is missing for the nginx-agent. Once that is made available in pkgs.nginx.com, I'll update the Dockerfile.

scripts/docker/official/nginx-plus-with-nginx-agent/alpine/Dockerfile
defanator: I'm trying to say that the only difference between nginx/foss image and N+ image is nginx package itself; it's much easier to take the same base image (foss one) and just replace the package from nginx to nginx-plus. That way you won't need to keep and maintain separate Dockerfile. YMMV.

thresheek: The size of the resulting image would probably be larger if the package is replaced.

craigell (approved):

scripts/docker/official/nginx-plus-with-nginx-agent/alpine/Dockerfile
dhurley: Yes the size of the image would be bigger. Also there isn't mush difference in terms of number of lines in the Dockerfile with either approach. I would prefer to follow the dockerfile provided by nginx-plus to be consistent.

aphralG (approved):

oliveromahony (approved):

PR #382, opened 2023-07-07
Fix backoff settings

dhurley:
### Proposed changes

Fix backoff settings

### Checklist

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -27,9 +27,9 @@ const (
 )
 
 var DefaultBackoffSettings = backoff.BackoffSettings{
-	InitialInterval: 10 * time.Second,
-	MaxInterval:     60 * time.Second,
-	MaxElapsedTime:  2 * time.Minute,
+	InitialInterval: 100 * time.Millisecond,
+	MaxInterval:     1 * time.Minute,
+	MaxElapsedTime:  0,
 	Jitter:          backoff.BACKOFF_JITTER,
 	Multiplier:      backoff.BACKOFF_MULTIPLIER,
 }
--- a/src/plugins/commander.go
+++ b/src/plugins/commander.go
@@ -88,8 +88,23 @@ func (c *Commander) Process(msg *core.Message) {
 
 func (c *Commander) agentBackoff(agentConfig *proto.AgentConfig) {
 	log.Debugf("update commander client configuration to %+v", agentConfig)
-	backOffSettings := sdk.ConvertBackOffSettings(agentConfig.Details.Server.GetBackoff())
+
+	if agentConfig.GetDetails() == nil || agentConfig.GetDetails().GetServer() == nil || agentConfig.GetDetails().GetServer().GetBackoff() == nil {
+		log.Debug("not updating commander client configuration as new Agent backoff settings is nil")
+		return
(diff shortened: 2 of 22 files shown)


src/plugins/commander.go
```diff
@@ -88,8 +88,23 @@ func (c *Commander) Process(msg *core.Message) {
 
 func (c *Commander) agentBackoff(agentConfig *proto.AgentConfig) {
 	log.Debugf("update commander client configuration to %+v", agentConfig)
-	backOffSettings := sdk.ConvertBackOffSettings(agentConfig.Details.Server.GetBackoff())
+
+	if agentConfig.GetDetails() == nil || agentConfig.GetDetails().GetServer() == nil || agentConfig.GetDetails().GetServer().GetBackoff() == nil {
+		log.Debug("not updating commander client configuration as new Agent backoff settings is nil")
```
achawla2012, line 93: Once server is out of planned outage mode, where should we reset agent backoff settings back? We do not want to have settings that server provided when it went in outage mode.

sdk/client/client.go
```diff
@@ -27,9 +27,9 @@ const (
 )
 
 var DefaultBackoffSettings = backoff.BackoffSettings{
-	InitialInterval: 10 * time.Second,
-	MaxInterval:     60 * time.Second,
-	MaxElapsedTime:  2 * time.Minute,
+	InitialInterval: 100 * time.Millisecond,
+	MaxInterval:     1 * time.Minute,
+	MaxElapsedTime:  0,
```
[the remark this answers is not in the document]
dhurley: Yes this is intention. By default we want to keep retrying to connect to the server when there is an outage.
But if a user doesn't want this they can always change these setting by updating their nginx-agent.conf with whatever backoff settings they prefer.

src/plugins/commander.go
dhurley: The agent is stateless, so it's up to the server to tell us what backoff settings they want. If you change the backoff settings and then want to update them again back to the default backoff setting, it is the server that need to send us the default backoff settings again in the agent config command.

src/plugins/metrics_sender.go
```diff
@@ -107,8 +107,23 @@ func (r *MetricsSender) Process(msg *core.Message) {
 
 func (r *MetricsSender) metricSenderBackoff(agentConfig *proto.AgentConfig) {
 	log.Debugf("update metric reporter client configuration to %+v", agentConfig)
-	backOffSettings := sdk.ConvertBackOffSettings(agentConfig.Details.Server.GetBackoff())
+
+	if agentConfig.GetDetails() == nil || agentConfig.GetDetails().GetServer() == nil || agentConfig.GetDetails().GetServer().GetBackoff() == nil {
```
aphralG, line 111: Hey, are we now no longer setting the defaults when the agentConfig Details etc are nil ?

dhurley: The agent is stateless, so it's up to the server to tell us what backoff settings they want. If you change the backoff settings and then want to update them again back to the default backoff setting, it is the server that needs to send us the default backoff settings again in the agent config command.

aphralG (approved):

PR #412, opened 2023-07-19
Unit test failures

oliveromahony:
### Proposed changes

Investigation into unit test failures. Changes to tidy some resources

### Checklist


src/plugins/commander.go
```diff
@@ -96,16 +96,7 @@ func (c *Commander) agentBackoff(agentConfig *proto.AgentConfig) {
 	}
 
 	backOffSettings := sdk.ConvertBackOffSettings(agentConfig.GetDetails().GetServer().GetBackoff())
-
 	c.cmdr.WithBackoffSettings(backOffSettings)
-	err := c.cmdr.Close()
```
aphralG, line 101: Hey, why was the close removed from here and metrics_sender ?

dhurley: Some of these changes are from my PR here https://github.com/nginx/agent/pull/401
The reconnect of grpc clients after the backoff settings are updated isn't actually necessary so we can stop doing this.

aphralG: Aww right, that makes sense got confused as to what change in this PR caused it to not be needed. Thank you!

aphralG (approved):

oliveromahony:
Waiting for https://github.com/nginx/agent/pull/401 before I merge

Dean-Coakley (approved):

dhurley (approved):

PR #426, opened 2023-08-04
Send error to UI when config async is disabled 

aphralG:
### Proposed changes

An error was not being sent to the UI when config async was disabled causing it to just eventually time out.

When config async is disabled instead of DeRegistering the `NginxBinary` plugin and then registering it, now the `isFeatureNginxConfigEnabled` is updated in `syncAgentConfigChange()` 

### Checklist


dhurley (approved):

oliveromahony (approved):

src/plugins/events.go
```diff
@@ -254,9 +254,13 @@ func (a *Events) sendConfigApplyEvent(msg *core.Message) {
 			command.Meta.GetMessageId(),
 		)
 	} else if nginxConfigResponse.Status.Status == proto.CommandStatusResponse_CMD_ERROR {
+		nginxId := ""
+		if nginxConfigResponse.GetConfigData() != nil {
+			nginxId = nginxConfigResponse.GetConfigData().NginxId
```
oliveromahony, line 259: .GetNginxId()

we should use the functions rather than the properties directly

oliveromahony (approved):

craigell (approved):

PR #437, opened 2023-08-15
Race condition reloading

oliveromahony:
### Proposed changes
Reload status url based on reload so that it doesn't cause a race condition. 

### Checklist


src/core/nginx.go
```diff
@@ -116,6 +119,11 @@ func (n *NginxBinaryType) UpdateNginxDetailsFromProcesses(nginxProcesses []*Proc
 	defer n.workersMapMutex.Unlock()
 	n.nginxWorkersMap = map[string][]*proto.NginxDetails{}
 
+	n.statusUrlMutex.Lock()
+	defer n.workersMapMutex.Unlock()
```
dhurley, line 123: this is already called on line 119 `defer n.workersMapMutex.Unlock()`

oliveromahony: yeah I have the local change but didn't push

Dean-Coakley (approved):

aphralG (approved):

src/plugins/dataplane_status.go
```diff
@@ -228,31 +225,19 @@ func (dps *DataPlaneStatus) detailsForProcess(processes []*core.Process, send bo
...
-		detail := dps.binary.GetNginxDetailsFromProcess(p)
-		if dps.statusUrls[detail.NginxId] != detail.StatusUrl {
-			log.Infof("NGINX status API updated.  Old status API: %v, new status API: %v", dps.statusUrls[detail.NginxId], detail.StatusUrl)
-			dps.statusUrls[detail.NginxId] = detail.StatusUrl
-			statusAPIUpdated = true
-		}
-		details = append(details, detail)
+		details = append(details, dps.binary.GetNginxDetailsFromProcess(p))
 		// spec says process CreateTime is unix UTC in MS
 		if time.Unix(p.CreateTime/1000, 0).After(dps.lastSendDetails) {
```
nickchen, line 235: nit, use time.UnixMilli https://pkg.go.dev/time#UnixMilli instead.

src/core/nginx.go
```diff
@@ -203,20 +210,30 @@ func (n *NginxBinaryType) GetNginxDetailsFromProcess(nginxProcess *Process) *pro
 		nginxDetailsFacade.ConfPath = path
 	}
 
-	stubStatusApiUrl, err := sdk.GetStubStatusApiUrl(nginxDetailsFacade.ConfPath, n.config.IgnoreDirectives)
-	if err != nil {
-		log.Tracef("Unable to get Stub Status API URL from the configuration: NGINX OSS metrics will be unavailable for this system. please configure aStub Status API to get NGINX OSS metrics: %v", err)
-	}
+	if len(n.statusUrls[nginxID]) == 0 || n.statusUrls[nginxID] == "" {
```
nickchen, line 213: i think the first check should be sufficient? they seems to be checking the same thing.

nickchen: also, potential race condition. The assignment for the `nginxID` in line 231 is mutex protected, but here it's not. I think a common pattern would be accessor functions that's mutex protected. Something like:

```
func (n ..) updateStatusURL(...) {
    n.statusUrlMutex.Lock()
    defer n.statusUrlMutex.Unlock(()
    ...
}
```

oliveromahony: there's no assignment here?

oliveromahony: updated to use the RLock

craigell (approved):

PR #454, opened 2023-08-25
Determine php fpm process status

achawla2012:
### Proposed changes

This change determines status of php-fpm (this would be included as part of php-fpm health check). 

### Testing
Tested via unit/Shell command

### Checklist


src/core/environment.go
```diff
@@ -676,14 +676,32 @@ type Shell interface {
 	Exec(cmd string, arg ...string) ([]byte, error)
 }
 
-type execShellCommand struct{}
+type ExecShellCommand struct{}
 
-func (e execShellCommand) Exec(cmd string, arg ...string) ([]byte, error) {
+// FakeShell mocks shell command output and errors
+type FakeShell struct {
```
ryepup, line 682: It's a little unusual to include test mocks in the production code.

Looks like this project already has some packages dedicated to test code, how about we move `FakeShell` into https://github.com/nginx/agent/blob/main/test/utils/environment.go, and then use that package from our tests?

src/extensions/php-fpm-metrics/pkg/phpfpm/status.go
```diff
@@ -0,0 +1,69 @@
...
+		return "INSTALLED"
+	case RUNNING:
+		return "RUNNING"
+	}
+	return "UNKNOWN"
+}
+
+// GetStatus returns phpfpm process status
+func GetStatus(pid, version string) Status {
+	output, err := shell.Exec("ps xao pid,ppid,command | grep 'php-fpm[:]'")
```
ryepup, line 35: this repo has a few features that depend on reading processes, could we re-use something like https://github.com/nginx/agent/blob/181b995a87a7c2905915f20acd0d2053c553ce64/src/core/process.go#L20C12-L20C12, or something else from <https://pkg.go.dev/github.com/shirou/gopsutil/v3/process>?

It's nice to avoid a shell execs if we can; they tend to be slower and can run into portability problems. A lot of times we can use a data structure directly instead of parsing stdout.

src/extensions/php-fpm-metrics/pkg/phpfpm/status.go
```diff
@@ -0,0 +1,69 @@
...
+	case INSTALLED:
+		return "INSTALLED"
+	case RUNNING:
+		return "RUNNING"
+	}
+	return "UNKNOWN"
+}
+
+// GetStatus returns phpfpm process status
+func GetStatus(pid, version string) Status {
```
ryepup, line 34: recommend returning `(Status, error)`, and letting the caller decide what to do with errors. This ends up being a more maintainable approach in general, as you can add error handling without touching a broad swath of the codebase.

src/extensions/php-fpm-metrics/pkg/phpfpm/status.go
```diff
@@ -0,0 +1,69 @@
...
+		if strings.Contains(l, "master process") {
+			parsed := strings.Fields(l)
+			if parsed[1] == pid {
+				return RUNNING
+			}
+		}
+	}
+
+	// not running; maybe it's installed
+	output, err = shell.Exec("ls", fmt.Sprintf("/etc/php/%s", version))
```
ryepup, line 57: [`os.ReadDir`](https://pkg.go.dev/os#ReadDir) is a good choice for reading the filesystem. If you want to keep it unit-testable, you can use it with [`fs.FS`](https://pkg.go.dev/io/fs#FS)

src/extensions/php-fpm-metrics/pkg/phpfpm/status_test.go
```diff
@@ -0,0 +1,92 @@
+package phpfpm
```
ryepup, line 1: in general I highly recommend keeping tests in a separate `*_tests` package.

From [The Zen of Go](https://the-zen-of-go.netlify.app/):

> *Write tests to lock in the behaviour of your package’s API*
>
> Test first or test later, if you shoot for 100% test coverage or are happy with less, regardless your package’s API is your contract with its users. Tests are the guarantees that those contracts are written in. Make sure you test for the behaviour that users can observe and rely on.

When our tests are in the same package as the code, we're able to use things outside the packages API, and potentially miss things.

src/core/environment.go
achawla2012: Sure. I can move to utils package.

src/extensions/php-fpm-metrics/pkg/phpfpm/status.go
achawla2012: In initial few MRs, I am putting forth on parity & functionality aspect by atomically adding components. These components at the moment are unit tested.
But I do plan to replace shell commands with `go-psutil` in future MRs where I would be integrating above atomic components, and in addition be testing on sandbox env.

src/extensions/php-fpm-metrics/pkg/phpfpm/status_test.go
ryepup: This usually requires design changes to add dependency injection or the functional options pattern to support mocking.

src/extensions/php-fpm-metrics/pkg/phpfpm/status.go
ryepup: > I am putting forth on parity & functionality aspect

ok, so something like porting the python to go, then as another pass making it more efficient/idiomatic? That kind of context is good to put in commit messages and PR descriptions. As a code reviewer my assumption is that the proposed code is complete.

Maybe we leave some TODOs in code to remind us that this is temporary?

Dean-Coakley (approved):

oliveromahony (approved):

aphralG (approved):

achawla2012:
Closing this as this is ported to https://github.com/nginx/agent/pull/460

PR #461, opened 2023-08-31
Add worker conn metrics

Dean-Coakley:
### Proposed changes

* Add [worker metrics](https://nginx.org/en/docs/http/ngx_http_api_module.html#workers_) from R30 release
* Change: detect latest api version and use that rather than defaulting to `6`. 
  * Now falls back to version `7` if apiVersion can not be determined
* Improve nginx plus collector tests. Reduced footprint and increase coverage

### Checklist


src/core/metrics/sources/nginx_plus.go
```diff
@@ -53,31 +59,39 @@ func NewNginxPlus(baseDimensions *metrics.CommonDim, nginxNamespace, plusNamespa
 func (c *NginxPlus) Collect(ctx context.Context, wg *sync.WaitGroup, m chan<- *metrics.StatsEntityWrapper) {
 	defer wg.Done()
 	c.init.Do(func() {
-		cl, err := plusclient.NewNginxClientWithVersion(&http.Client{}, c.plusAPI, c.clientVersion)
+		latestAPIVersion, err := getLatestAPIVersion(ctx, &http.Client{}, c.plusAPI)
+		if err != nil {
+			c.logger.Log(fmt.Sprintf("Failed to check available api versions: %v", err))
```
oliveromahony, line 64: if there is an error, can you try use the default client version? https://github.com/nginx/agent/blob/main/src/core/config/defaults.go#L64 this should be 7

src/core/metrics/sources/nginx_plus.go
```diff
@@ -53,31 +59,39 @@ func NewNginxPlus(baseDimensions *metrics.CommonDim, nginxNamespace, plusNamespa
...
-			c.logger.Log(fmt.Sprintf("Failed to retrieve plus metrics, %v", err))
+			c.logger.Log(fmt.Sprintf("Failed to retrieve plus metrics: %v", err))
 			SendNginxDownStatus(ctx, c.baseDimensions.ToDimensions(), m)
 			c.prevStats = nil
 			return
 		}
 	})
 
-	cl, err := plusclient.NewNginxClientWithVersion(&http.Client{}, c.plusAPI, c.clientVersion)
+	cl, err := plusclient.NewNginxClient(&http.Client{}, c.plusAPI)
```
oliveromahony, line 85: why was this changed? we should control the version sent to the client if we calculate it?

src/core/metrics/sources/nginx_plus.go
```diff
@@ -725,7 +738,7 @@ func (c *NginxPlus) cacheMetrics(stats, prevStats *plusclient.Stats) []*metrics.
 	return zoneMetrics
 }
 
-func (c *NginxPlus) slabMetrics(stats, prevStats *plusclient.Stats) []*metrics.StatsEntityWrapper {
```
dhurley, line 728: Can you remove the prevStats argument here? Looks like it isn't used in the function.

src/core/metrics/sources/nginx_plus.go
```diff
@@ -827,10 +877,48 @@ func createStreamPeerMap(peers []plusclient.StreamPeer) map[string]plusclient.St
...
-	if mybool {
+func boolToFloat64(myBool bool) float64 {
+	if myBool {
 		return valueFloat64One
 	} else {
 		return valueFloat64Zero
 	}
 }
+
+func (c *NginxPlus) getLatestAPIVersion(ctx context.Context, httpClient *http.Client, endpoint string) (int, error) {
```
dhurley, line 888: Do you need to pass httpClient as an argument? Could you create it at the start of the function instead?

src/core/metrics/sources/nginx_plus.go
Dean-Coakley: Marking as resolved as this is already done as far as I can see. Let me know if anything else is required.

dhurley (approved):

oliveromahony (approved):

aphralG (approved):

craigell (approved):

PR #464, opened 2023-09-04
Add support for file actions during a config apply

dhurley:
### Proposed changes

With the addition of file actions, we can now support partial updates to NGINX configuration. Before all NGINX config files needed to be sent to perform a config apply. Now you only need to send the files that you have made updates to.

### Checklist


src/core/environment.go
```diff
@@ -236,13 +238,69 @@ func (env *EnvironmentType) WriteFiles(backup ConfigApplyMarker, files []*proto.
...
+		if err = env.WriteFile(backup, file, confPath); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+// WriteFile writes the provided file content to disk. If the file.GetName() returns an absolute path, it'll be written
+// to the path. Otherwise, it'll be written to the path relative to the provided confPath.
+func (env *EnvironmentType) WriteFile(backup ConfigApplyMarker, file *proto.File, confPath string) error {
```
oliveromahony, line 250: can you add benchmark tests for these new functions? https://github.com/nginx/agent/blob/main/test/performance/environment_test.go#L88

dhurley: sure

Dean-Coakley (approved):

src/core/nginx.go
```diff
@@ -399,81 +399,115 @@ func (n *NginxBinaryType) WriteConfig(config *proto.NginxConfig) (*sdk.ConfigApp
...
+	} else {
+		configApply, err = sdk.NewConfigApplyWithIgnoreDirectives("", n.config.AllowedDirectoriesMap, n.config.IgnoreDirectives)
+		if err != nil {
+			log.Warnf("config_apply error: %s", err)
+			return nil, err
 		}
-		if err = os.Remove(file); err != nil {
-			return configApply, err
+
+		for _, file := range confFiles {
```
nginx-nickc, line 481: instead of two loops, i think you can simplify it with `for _, file := range append(confFiles, auxFiles...)`, I think it looks like both loop logic are the same.

src/core/nginx.go
```diff
@@ -399,81 +399,115 @@ func (n *NginxBinaryType) WriteConfig(config *proto.NginxConfig) (*sdk.ConfigApp
...
+			if _, found := filesToUpdate[file.Name]; !found {
+				continue
+			}
+
+			if err := n.env.WriteFile(configApply, file, rootDirectoryPath); err != nil {
+				log.Warnf("configuration write failed: %s", err)
+				return configApply, err
+			}
+		}
+
```
nginx-nickc, line 504: should we delete from `filesToUpdate` and warn if there are left over files after?

src/core/nginx.go
```diff
@@ -514,6 +548,51 @@ func generateDeleteFromDirectoryMap(
...
+			if !filepath.IsAbs(path) {
+				continue
+			}
+
+			if !allowedFile(path, allowedDirectory) {
+				continue
+			}
+
+			if f.Action == proto.File_unset {
+				allFilesHaveAnAction = false
```
nginx-nickc, line 579: if feels like we should then able to return here?

src/core/nginx.go
```diff
@@ -514,6 +548,51 @@ func generateDeleteFromDirectoryMap(
...
+) {
+	allFilesHaveAnAction = true
+	filesToUpdate = make(map[string]proto.File_Action)
+	filesToDelete = make(map[string]proto.File_Action)
+
+	for _, dir := range directoryMap.Directories {
+		for _, f := range dir.Files {
+			path := filepath.Join(dir.Name, f.Name)
+
+			log.Tracef("file %s has action %v", path, f.Action)
```
nginx-nickc, line 567: nit, debug?

src/core/nginx.go
dhurley: Unfortunately I can't as there is one slight difference in the two for loops. The way we determine the `rootDirectoryPath` is different depending on if it's a config or aux file.

src/core/nginx.go
```diff
@@ -399,81 +399,115 @@ func (n *NginxBinaryType) WriteConfig(config *proto.NginxConfig) (*sdk.ConfigApp
...
-	if err != nil {
-		log.Warnf("configuration write failed: %s", err)
-		n.writeBackup(config, confFiles, auxFiles)
-		return configApply, err
-	}
+	filesToUpdate, filesToDelete, allFilesHaveAnAction := generateActionMaps(config.DirectoryMap, n.config.AllowedDirectoriesMap)
+
+	// If the file action is unset for all files then all files are written to disk and a diff is performed to determine what files need to be deleted
+	if !allFilesHaveAnAction {
+		log.Debug("all files in directory map have no action set")
```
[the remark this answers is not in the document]
dhurley: Updated

src/core/nginx.go
dhurley: Updated

oliveromahony (approved):

PR #500, opened 2023-10-12
fix: avoid re-use of context by commander for Download and Upload requests

CodeMonkeyF5:
### Proposed changes

When agent reconnects to NIM after a disconnect, config apply doesn’t work as the DPM handler that was established with the original context is no longer valid. To resolve this the gRPC client Download and Upload needs to establish new contexts for each request.

### Checklist


Dean-Coakley (approved):

CodeMonkeyF5:
Issue was partially mitigated in DPM.  Change to agent needs to be done to ensure multiple `createClient` are not called during retry to return the number of connect requests.

PR #537, opened 2024-01-11
New project structure

dhurley:
### Proposed changes

- Created new project structure
- Implemented basic get instances REST API endpoint

### Checklist


olli-holmala:
The `util` package name is specifically called out as an example of a bad package name [here](https://go.dev/blog/package-names).

image

internal/model/instances/instances.proto
```diff
@@ -0,0 +1,34 @@
+/**
```
olli-holmala, line 1: Would it make sense to flatten the directory structure in `internal/model`?

olli-holmala: Or were these split up because there name collisions in the generated code?

internal/api/http/dataplane/client.gen.go
```diff
@@ -0,0 +1,236 @@
+// Package dataplane provides primitives to interact with the openapi HTTP API.
```
olli-holmala, line 1: We've now introduced a dataplane client for our own internal use (it's private to this project). Are we sure we don't want to export this for external use?

internal/model/os/os.proto
```diff
@@ -0,0 +1,19 @@
...
+ * This source code is licensed under the Apache License, Version 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+syntax = "proto3";
+package f5.nginx.agent.internal.model.os;
+
+option go_package = "github.com/nginx/agent/v3/internal/model/os";
+
+message Process {
```
olli-holmala, line 13: `proto` files are typically contract definitions, are `os.proto` and `instances.proto` both not meant for usage outside the project?

internal/plugin/dataplane_server.go
```diff
@@ -0,0 +1,153 @@
...
+		address         string
+		logger          *slog.Logger
+		instanceService service.InstanceServiceInterface
+		messagePipe     *bus.MessagePipe
+		server          net.Listener
+		processes       []*os.Process
+	}
+)
+
+type DataplaneServerParameters struct {
```
olli-holmala, line 46: nit: move this to the `type` block above.

internal/plugin/dataplane_server.go
```diff
@@ -0,0 +1,153 @@
...
+	Address         string
+	Logger          *slog.Logger
+	instanceService service.InstanceServiceInterface
+}
+
+func NewDataplaneServer(dataplaneServerParameters *DataplaneServerParameters) *DataplaneServer {
```
olli-holmala, line 52: Why is `DataplaneServer` a plugin?

Makefile
```diff
@@ -0,0 +1,74 @@
...
+no-local-changes:
+	git diff --quiet || { echo "Dependency changes detected. Please commit these before pushing." >&2; exit 1; }
+
+build: ## Build agent executable
+	@mkdir -p $@
+	@$(GOBUILD) -o $(BINARY) $(PROJECT_DIR)/${PROJECT_FILE}
+	@echo "📦 Build Done"
+
+lint: ## Run linter
+	GOWORK=off go vet ./...
```
olli-holmala, line 36: nit: we don't need all these `GOWORK=off` variables in this project.

Makefile
```diff
@@ -0,0 +1,74 @@
...
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\033[36m\033[0m\n"} /^[$$()% 0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-24s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+clean: ## Remove build directory
+	@rm -rf ${BUILD_DIR}
+	@echo "🌀 Cleaning Done"
+
+no-local-changes:
+	git diff --quiet || { echo "Dependency changes detected. Please commit these before pushing." >&2; exit 1; }
+
+build: ## Build agent executable
```
olli-holmala, line 30:
```suggestion
$(BUILD_DIR): ## Build agent executable
```

Makefile
```diff
@@ -0,0 +1,74 @@
...
+	GOWORK=off go vet ./...
+	GOWORK=off go run github.com/golangci/golangci-lint/cmd/golangci-lint run -c ./scripts/.golangci.yml
+
+format: ## Format code
+	@$(GORUN) mvdan.cc/gofumpt -l -w .
+	@echo "🏯 Format Done"
+
+install-tools: ## Install dependencies in tools.go using vendored version see https://www.jvt.me/posts/2023/06/19/go-install-from-mod/
+	@echo "Installing Tools"
+	@grep _ ./tools/tools.go | awk '{print $$2}' | xargs -tI % env GOBIN=$$(git rev-parse --show-toplevel)/bin GOWORK=off go get %
```
olli-holmala, line 45: I'm assuming that this line gets the vendored version, which we no longer have, so we can delete it.

internal/app.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/nginx/agent/v3/internal/plugin"
+)
+
+type App struct{}
+
+func NewApp() *App {
+	return &App{}
+}
+
+func (*App) Run() {
```
olli-holmala, line 26: This should likely return an error but that's probably out of scope for this PR (focus should be on directory structure).

internal/datasource/nginx/instances.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Nginx struct {
+	parameters NginxParameters
+}
+
+type NginxParameters struct {
+	GetInfo GetInfo
+}
+
+func NewNginx(parameters NginxParameters) *Nginx {
```
olli-holmala, line 34: nit: https://google.github.io/styleguide/go/decisions.html#package-vs-exported-symbol-name
```suggestion
func New(parameters NginxParameters) *Nginx {
```
So when the caller inits this data source, it'll look like this
```
nginx := nginx.New(params)
```
instead of
```
nginx := nginx.NewNginx(params)
```

olli-holmala: After looking at `process.go`, it looks like it should be something like `NewInstance` or something like that.

dhurley:
> The `util` package name is specifically called out as an example of a bad package name [here](https://go.dev/blog/package-names).
> 
> image

Ah yes I was meant to come back to that package. I'll see if I can call it something better that just `util`

internal/model/instances/instances.proto
dhurley: yeah I put them in separate directories because I wanted to avoid name collisions in the future. 
For example, messages named `Meta` or `Type` might be used in other messages in the future but have different contents.

internal/plugin/dataplane_server.go
dhurley: gives us the flexibility of disabling the API in future depending on what plugins you specify to enable in your agent config

Makefile
dhurley: I have another commit coming soon that will change how we handle installing tools so that we install a specific version each time

Makefile
```diff
@@ -0,0 +1,74 @@
...
+GORUN	= ${GOCMD} run
+
+BUILD_DIR		:= build
+TEST_BUILD_DIR  := build/test
+BINARY_NAME		:= nginx-agent
+BINARY			:= $(BUILD_DIR)/$(BINARY_NAME)
+PROJECT_DIR		= cmd/agent
+PROJECT_FILE	= main.go
+IMPORT_MAPPING 	:= ../common/common.yaml:github.com/nginx/agent/v3/internal/api/http/common
+
```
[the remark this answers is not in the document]
olli-holmala: Had a Zoom discussion about this with Donal. I'll add the same explanation that I gave there here too for the purpose of a `.PHONY`.

So a `.PHONY` is meant to let `make` know that the target does not produce a directory or file. Essentially every target that is defined in a Makefile is by default supposed to produce a directory or file that is the same name as the target. 

e.g. the target `clean` is by default expected to produce a directory called `clean`. Now when you add `clean` to `.PHONY`, you are saying that `clean` is essentially a function that does something instead of producing the `clean` directory.

In case my rambling explanation wasn't clear, here's maybe a better one: https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile

internal/datasource/nginx/instances.go
dhurley: Thanks for the feedback. I'll take a look at refactoring this.

internal/api/http/dataplane/client.gen.go
dhurley: Moving to root directory of project

internal/model/os/os.proto
dhurley: Moving proto files to `/api/grpc`

Makefile.tools
```diff
@@ -0,0 +1,16 @@
+OAPICODEGEN	  = github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen@v2.0.0
+LEFTHOOK	  = github.com/evilmartians/lefthook@v1.5.5
+GOLANGCILINT  = github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
+PROTOCGENGO	  = google.golang.org/protobuf/cmd/protoc-gen-go@v1.32.0
+FOFUMPT	      = mvdan.cc/gofumpt@v0.5.0
+COUNTERFEITER = github.com/maxbrunsfeld/counterfeiter/v6@v6.7.0
+
+install-tools: ## Install dependencies in tools.go using vendored version see https://www.jvt.me/posts/2023/06/19/go-install-from-mod/
```
olli-holmala, line 8:
```suggestion
install-tools:
```

internal/plugin/dataplane_server.go
olli-holmala: Missed this one 😄

internal/service/instance.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/nginx/agent/v3/api/grpc/instances"
+	"github.com/nginx/agent/v3/internal/datasource/nginx"
+	"github.com/nginx/agent/v3/internal/model/os"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.7.0 -generate
+//counterfeiter:generate -o mock_instance.go . InstanceServiceInterface
+//go:generate sh -c "grep -v github.com/nginx/agent/v3/internal/service mock_instance.go | sed -e s\\/service\\\\.\\/\\/g > mock_instance_fixed.go"
+//go:generate mv mock_instance_fixed.go mock_instance.go
+type InstanceServiceInterface interface {
```
olli-holmala, line 20: I'll just raise this point to make sure that we are making a conscious decision on this: **Do we want to propagate the convention of suffixing interfaces with** `-Interface`?

There seem to be several ways of doing this: https://medium.com/@dotronglong/interface-naming-convention-in-golang-f53d9f471593

In short:
1. Prepend the letter "I"  => `IInstanceService` - turns out this is a perfect example of a corner case where it doesn't really work too well
2. Use the `-er` suffix => `InstanceServicer` - typically used for one-method interfaces (per [Effective Go](https://go.dev/doc/effective_go#interface-names))
3. Keep it as is with the `Interface` suffix.

I don't really like any of the three tbh but I'd maybe propose suffixing `-Interface` when there's multiple methods and using `-er` if there's one method?

olli-holmala (approved):
LGTM, thanks @dhurley for being patient with all my feedback! 🙏

Makefile
```diff
@@ -0,0 +1,72 @@
+# Go parameters
```
oliveromahony, line 1: might be worth putting these in their own makefile also?

go.mod
```diff
@@ -1,3 +1,60 @@
 module github.com/nginx/agent/v3
 
 go 1.21.4
```
oliveromahony, line 3: go 1.21.6 is out

internal/app.go
```diff
@@ -0,0 +1,53 @@
...
+	ctx := context.Background()
+
+	handler := slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelDebug})
+	logger := slog.New(handler)
+	slog.SetDefault(logger)
+
+	slog.Info("starting NGINX Agent")
+
+	processMonitor := plugin.NewProcessMonitor(&plugin.ProcessMonitorParameters{
+		MonitoringFrequency: 1 * time.Minute,
```
oliveromahony, line 36: should we add configuration for reference?

dhurley: Didn't want to do too much in this PR. So I might make a separate PR where we read in the agent config.

oliveromahony: the configuration should probably get passed to the app then

dhurley: yeah the `app.go` would read the config and pass the relevant info to the plugins

internal/service/instance.go
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) F5, Inc.
+ *
+ * This source code is licensed under the Apache License, Version 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package service
```
oliveromahony, line 8: Is this the package doing the heavy lifting? Is the plan for this Service layer to be the brains of the Agent? Seems like the plugin layer is in this implementation?

internal/service/instance.go
oliveromahony: whatever we decide, we should start documenting decisions like these

internal/app.go
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (c) F5, Inc.
+ *
+ * This source code is licensed under the Apache License, Version 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package internal
```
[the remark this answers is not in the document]
dhurley: Lets keep it internal for now. We can always move it if needed

internal/app_test.go
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright (c) F5, Inc.
+ *
+ * This source code is licensed under the Apache License, Version 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package internal
```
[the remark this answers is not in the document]
dhurley: Not sure how to test the App yet in its current state. I'll add tests when I add the agent config logic in here as it will be easier to test then.

internal/service/instance.go
dhurley: Yes the service layer should brains of the agent. At the moment its not doing much since its just getting NGINX instance data. 
The plugin layer will delegate requests to the service layer which will determine what to do with the requests and call out to the appropriate datasource.

Makefile
dhurley: don't think so at the moment. If the Makefile gets too big in the future we can probably take a look at it then

oliveromahony (approved):

PR #539, opened 2024-01-16
Add agent configuration functionality

dhurley:
### Proposed changes

Add the ability to configure the following:

- Change log level and log file path
- Change process monitoring frequency
- Change dataplane API host & port

### Checklist


internal/app.go
```diff
@@ -10,44 +10,56 @@ package internal
...
 
-type App struct{}
+type App struct {
+	commit  string
+	version string
+}
 
-func NewApp() *App {
-	return &App{}
+func NewApp(commit string, version string) *App {
```
olli-holmala, line 26:
```suggestion
func NewApp(commit, version string) *App {
```

internal/logger/looger_test.go
```diff
@@ -0,0 +1,105 @@
+/**
```
olli-holmala, line 1: File name `looger` => `logger`

internal/logger/logger.go
```diff
@@ -0,0 +1,71 @@
...
+	)
+
+	return slog.New(handler)
+}
+
+func getLogLevel(level string) slog.Level {
+	if level == "" {
+		return slog.LevelInfo
+	}
+	return logLevels[strings.ToLower(level)]
```
olli-holmala, line 46: What about if the log level is not found from the `logLevels` map?

dhurley: It returns level info by default if not found

internal/config/flags.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"time"
+
+	flag "github.com/spf13/pflag"
+)
+
+// Registrable registers a pflag
+type Registrable interface {
+	register(*flag.FlagSet)
+}
```
olli-holmala, line 19: I know this is likely copied code from v2 but per Go's naming guide this should be called `Registerer`: https://go.dev/doc/effective_go#interface-names

oliveromahony: Maybe name it something more appropriate with the -er suffix then?

internal/config/flags.go
```diff
@@ -0,0 +1,79 @@
...
+
+type StringFlag struct {
```
olli-holmala, line 21: nit: do we wanna wrap these under the same `type` block? i.e.
```go
type (
  StringFlag struct {
    Name         string
    Usage        string
    DefaultValue string
  }

  StringSliceFlag struct {
    Name         string
    Usage        string
    DefaultValue []string
  }

  ...
)
```

internal/config/types.go
```diff
@@ -0,0 +1,32 @@
...
+	DataplaneAPI   DataplaneAPI   `mapstructure:"dataplane_api" yaml:"-"`
+}
+
+type Log struct {
+	Level string `mapstructure:"level" yaml:"-"`
+	Path  string `mapstructure:"path" yaml:"-"`
+}
+
+type ProcessMonitor struct {
+	MonitoringFrequency time.Duration `mapstructure:"monitoring_frequency" yaml:"-"`
```
olli-holmala, line 26: Should `MonitoringFrequency` also be defined here in this file?

internal/logger/logger.go
```diff
@@ -0,0 +1,71 @@
...
+		if fileInfo.IsDir() {
+			logPath = path.Join(logPath, defaultLogFile)
+		}
+
+		logFileHandle, err := os.OpenFile(logPath, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o600)
+		if err != nil {
+			slog.Error("failed to open log file, proceeding to log only to stdout/stderr", "error", err)
+			return os.Stderr
+		}
+		return logFileHandle
```
olli-holmala: Since the `logFile` argument comes from user input, this code will likely return an error from our static analyzer for Path Traversal Attacks: https://owasp.org/www-community/attacks/Path_Traversal

We'd need to resolve symlinks and ensure that we are working in a trusted path to avoid that error. Here's a good Go guide for avoiding it: https://www.stackhawk.com/blog/golang-path-traversal-guide-examples-and-prevention/

internal/config/flags.go
dhurley: Removing Registrable as its not needed anymore

internal/config/flags.go
dhurley: Removing flags.go

internal/app.go
```diff
@@ -10,44 +10,56 @@ package internal
...
 
-func NewApp() *App {
-	return &App{}
+func NewApp(commit, version string) *App {
+	return &App{commit, version}
 }
 
-func (*App) Run() error {
-	ctx := context.Background()
+func (a *App) Run() error {
```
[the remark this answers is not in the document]
olli-holmala: The one letter method receiver is, unfortunately, idiomatic to Go: https://go.dev/wiki/CodeReviewComments#receiver-names

But I'm okay with breaking convention for this case, since `app` is such a short word anyway.

olli-holmala (approved):

internal/logger/logger.go
dhurley: For log files I think allowing the user to freely specify where they want their logs to be written to is ok, but we will be adding an `allowedDirectories` map in another PR that we can use to validate config file paths for NGINX, Unit, etc.

internal/app.go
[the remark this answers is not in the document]
dhurley: I'd be more inclined to follow the convention in Go just to be consistent.

nginx-agent.conf
```diff
@@ -0,0 +1,24 @@
...
+  # set log path. if empty, don't log to file.
+  path: /var/log/nginx-agent/
+
+dataplane_api:
+  # The port at which NGINX Agent accepts remote connections
+  # The API address and port allow for remote management of NGINX and NGINX Agent
+  #
+  # ~~~ WARNING ~~~ 
+  # Set API address to allow remote management
+  # host: 127.0.0.1
```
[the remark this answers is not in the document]
dhurley: By default, the agent won't enable the dataplane API until the user updates your nginx-agent.conf. We will add documentation later on how to install the agent and configure it for different use cases.

aphralG (approved):

oliveromahony (approved):

PR #540, opened 2024-01-16
Config Download

aphralG:
### Proposed changes

Added: 
 - Download File using GetFilesMetaData() & GetFile()
 - Cache to be used to check if file needs to be downloaded or is current version on disk the latest 
 

### Checklist


internal/datasource/os/files_test.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func createProtoTime(timeString string) (*timestamppb.Timestamp, error) {
+	time, err := time.Parse(time.RFC3339, timeString)
+	protoTime := timestamppb.New(time)
+
+	return protoTime, err
+}
+
+func createCacheFile(t *testing.T, cachePath string) map[string]*instances.File {
```
dhurley, line 196: remove *testing.T from the function parameters. The asserts should only be done in the actual tests and not helper functions. This function should return an error of the test that calls it should assert the error.

internal/client/http_config_client.go
```diff
@@ -0,0 +1,120 @@
...
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.7.0 -generate
+//counterfeiter:generate -o mock_http_config_downloader.go . HttpConfigDownloaderInterface
+//go:generate sh -c "grep -v github.com/nginx/agent/v3/internal/client mock_http_config_downloader.go | sed -e s\\/client\\\\.\\/\\/g > mock_http_config_downloader_fixed.go"
+//go:generate mv mock_http_config_downloader_fixed.go mock_http_config_downloader.go
+type HttpConfigDownloaderInterface interface {
+	GetFilesMetadata(filesUrl string, tenantID uuid.UUID) (*instances.Files, error)
+	GetFile(file *instances.File, filesUrl string, tenantID uuid.UUID) (*instances.FileDownloadResponse, error)
+}
+
+type HttpConfigDownloader struct {
```
dhurley, line 32:
```suggestion
type HttpConfigClient struct {
```

internal/datasource/os/files.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	return &FileSource{
+		configDownloader: fileSourceParameters.configDownloader,
+	}
+}
+
+func WriteFile(fileContent []byte, filePath string) error {
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		slog.Debug("File does not exist, creating")
```
dhurley, line 46:
```suggestion
		slog.Debug("File does not exist, creating new file", "file", filePath)
```

internal/datasource/os/files_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	err = os.Remove(filePath)
+	assert.NoError(t, err)
+	assert.NoFileExists(t, filePath)
+}
+
+func TestReadCache(t *testing.T) {
+	instanceId, err := uuid.Parse("aecea348-62c1-4e3d-b848-6d6cdeb1cb9c")
+	assert.NoError(t, err)
+	cachePath := fmt.Sprintf("/tmp/%v/cache.json", instanceId.String())
```
olli-holmala, line 45:
```suggestion
	cachePath := fmt.Sprintf("/tmp/%s/cache.json", instanceId.String())
```

internal/datasource/os/files_test.go
```diff
@@ -0,0 +1,154 @@
...
+	time, err := time.Parse(time.RFC3339, timeString)
+	protoTime := timestamppb.New(time)
+
+	return protoTime, err
+}
+
+func createCacheFile(cachePath string) (map[string]*instances.File, error) {
+	timeFile1, err := createProtoTime("2024-01-08T13:22:23Z")
+	if err != nil {
+		return nil, fmt.Errorf("error creating time, error: %v", err)
```
olli-holmala, line 108:
```suggestion
		return nil, fmt.Errorf("error creating time, error: %w", err)
```

internal/datasource/os/files_test.go
olli-holmala: More context for cryptic percent string formatting: https://gobyexample.com/string-formatting

internal/datasource/os/files_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func createProtoTime(timeString string) (*timestamppb.Timestamp, error) {
+	time, err := time.Parse(time.RFC3339, timeString)
+	protoTime := timestamppb.New(time)
+
+	return protoTime, err
+}
+
+func createCacheFile(cachePath string) (map[string]*instances.File, error) {
+	timeFile1, err := createProtoTime("2024-01-08T13:22:23Z")
```
olli-holmala, line 106: nit:
```suggestion
	fileTime1, err := createProtoTime("2024-01-08T13:22:23Z")
```
`timeFile` sounds like the type is a `*os.File` instead of a `time.Time`.

internal/datasource/os/files.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/nginx/agent/v3/api/grpc/instances"
+)
+
+func WriteFile(fileContent []byte, filePath string) error {
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		slog.Debug("File does not exist, creating new file", "file", filePath)
+		err = os.MkdirAll(path.Dir(filePath), 0o750)
+		if err != nil {
+			return fmt.Errorf("error creating directory, directory: %v, error: %v", path.Dir(filePath), err)
```
olli-holmala, line 25: Maybe lower the verbosity of the error messages slightly (leads to shorter log entries later)?
```suggestion
			return fmt.Errorf("error creating directory %s: %w", path.Dir(filePath), err)
```
Same applies to the error on line 31.

internal/datasource/os/files.go
```diff
@@ -0,0 +1,69 @@
...
+	err := os.WriteFile(filePath, fileContent, 0o644)
+	if err != nil {
+		return fmt.Errorf("error writing to file, filePath: %v, error:%v", filePath, err)
+	}
+	slog.Debug("Content written to file", "filePath", filePath)
+
+	return nil
+}
+
+func ReadInstanceCache(cachePath string) (map[string]*instances.File, error) {
```
olli-holmala, line 38: What's the reason why we are using a `map[string]*instances.File` instead of introducing a `struct` with `path` and `file` fields/attributes? As a caller of this code, I won't know that the key to this map represents a file path without looking at the unit tests for this file.

internal/datasource/config_writer.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return currentConfigApply, skippedFiles, err
+}
+
+func isFilePathValid(filePath string) bool {
+	if filePath != "" && !strings.HasSuffix(filePath, "/") {
+		return true
+	}
+	return false
```
olli-holmala, line 78:
```suggestion
	return filePath != "" && !strings.HasSuffix(filePath, "/")
```

olli-holmala: Might need parentheses, not sure 😅

internal/datasource/config_writer.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func doesFileRequireUpdate(lastConfigApply map[string]*instances.File, fileData *instances.File) (latest bool) {
+	if lastConfigApply != nil && len(lastConfigApply) > 0 {
+		fileOnSystem, ok := lastConfigApply[fileData.Path]
+		if ok && !fileData.LastModified.AsTime().After(fileOnSystem.LastModified.AsTime()) {
+			return false
+		}
+		return true
```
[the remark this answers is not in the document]
olli-holmala: Side note: let's never use `!fileData.LastModified.AsTime().After(fileOnSystem.LastModified.AsTime())` in the future, when we could just use `fileData.LastModified.AsTime().Before(fileOnSystem.LastModified.AsTime())`

internal/datasource/config_writer.go
```diff
@@ -0,0 +1,90 @@
...
+func (cw *ConfigWriter) Write(lastConfigApply map[string]*instances.File, filesUrl string, tenantID uuid.UUID) (currentConfigApply map[string]*instances.File, skippedFiles map[string]struct{}, err error) {
+	currentConfigApply = make(map[string]*instances.File)
+	skippedFiles = make(map[string]struct{})
+
+	filesMetaData, err := cw.configDownloader.GetFilesMetadata(filesUrl, tenantID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error getting files metadata, filesUrl: %v, error: %v", filesUrl, err)
+	}
+
+filesLoop:
```
olli-holmala, line 43: Are we sure that `filesLoop` is necessary, as there is only one loop? My understanding was that named loops are used to break or continue an outer loop when in an inner loop 🤔

internal/datasource/os/files_test.go
aphralG: I think I got them all let me know if I missed some

internal/datasource/os/files_test.go
aphralG: I think I fixed all the errors let me know if not

internal/datasource/os/files.go
aphralG: Discussed this with Olli and Donal and decided to use an Alias for `map[string]*instances.File ` so its clearer what the map is

internal/datasource/config_writer.go
aphralG: Think this was left in by accident, Thank you!

internal/datasource/config_writer.go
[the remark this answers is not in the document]
aphralG: I've made the change with the return but want to see can I make it better as its hard to read at the moment with the `not after`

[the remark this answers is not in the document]
aphralG: `fileOnSystem.LastModified.AsTime().Before(fileData.LastModified.AsTime())` is used now instead

internal/client/http_config_client.go
```diff
@@ -0,0 +1,114 @@
...
+	GetFile(file *instances.File, filesUrl string, tenantID uuid.UUID) (*instances.FileDownloadResponse, error)
+}
+
+type HttpConfigClient struct {
+	httpClient http.Client
+}
+
+func NewHttpConfigClient() *HttpConfigClient {
+	httpClient := http.Client{
+		Timeout: time.Second * 10,
```
oliveromahony, line 38: can you move this to a config setting?

internal/datasource/os/files.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"log/slog"
+	"os"
+	"path"
+
+	"github.com/nginx/agent/v3/api/grpc/instances"
+)
+
+// map of files with filepath as key
+type FileCache = map[string]*instances.File
```
oliveromahony, line 21: should this type be visible outside of this file?

aphralG: Yes as its needed in config writer

oliveromahony (approved):

internal/datasource/config_writer.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type (
+	Client struct {
+		Timeout time.Duration
+	}
+
+	ConfigWriterParameters struct {
+		configDownloader client.HttpConfigClientInterface
+		client           Client
```
dhurley, line 29:
```suggestion
		Client           Client
```

dhurley (approved):

PR #542, opened 2024-01-18
Add update instance config endpoint

dhurley:
### Proposed changes
- Add update instance config endpoint.
- Add instance monitor plugin.

### Checklist


aphralG (approved):

Dean-Coakley (approved):

internal/service/instance.go
```diff
@@ -8,34 +8,54 @@
...
-	is.processes = newProcesses
+func (is *InstanceService) GetInstances() []*instances.Instance {
+	return is.instances
 }
 
-func (is *InstanceService) GetInstances() ([]*instances.Instance, error) {
-	n := nginx.New(nginx.NginxParameters{})
-	instances, err := n.GetInstances(is.processes)
-	return instances, err
+func (is *InstanceService) UpdateInstanceConfiguration(instanceId string, location string) (string, error) {
```
olli-holmala, line 53: nit: do we wanna use named return values to clearly denote what the `string` is supposed to represent? I know it's obvious from reading the function itself, but you'd see it immediately based on just the function signature with named return values 🤔 
```suggestion
func (is *InstanceService) UpdateInstanceConfiguration(instanceId string, location string) (correlationID string, err error) {
```

olli-holmala (approved):
LGTM!

internal/service/instance.go
aphralG: +1, think it would be nice to immediately tell what functions are returning

dhurley: updated

PR #544, opened 2024-01-21
feat: Introduce metrics plugin

olli-holmala:
### Proposed changes

Adds the Metrics plugin with one data source, the Prometheus data source.

Other additions:
* A Metrics config
* Modified `Plugin` contract to return errors
* Several OTel functions for reuse with other data sources

TODO items:
* Several unit tests are missing because I could not mock the OTel interfaces due to each of them having embedded interfaces, which `counterfeiter` does not seem to understand.
* Still looking into how to nicely mock gRPC connections, which were required e.g. for the Metrics plugin's `Init()` method unit test.
* TLS support for the gRPC connection (via Metrics or Agent config).
* Dynamic discovery of data sources (integration with Message Pipe sources)
  * How does the Agent discover that there is a Prometheus endpoint for it to target?
* Proper integration with `MessagePipe`, such that new data sources are automatically discovered once a new process/instance event comes in.

### Checklist


oliveromahony:
can we pull in https://github.com/nginx/agent/blob/main/test/performance/plugins_test.go and something similar to https://github.com/nginx/agent/blob/main/src/core/mock_pipe.go because they will be useful (don't mind if they are in a separate PR)

internal/bus/message_pipe.go
```diff
@@ -9,39 +9,52 @@ package bus
...
-	plugins        []Plugin
-	mu             sync.RWMutex
-	ctx            context.Context
-	cancel         context.CancelFunc
-}
+	Plugin interface {
+		Init(MessagePipeInterface) error
+		Close() error
+		Info() *Info
+		Process(*Message) error
```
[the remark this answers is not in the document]
oliveromahony: if it fails to process should we capture that?

internal/datasource/prometheus/prometheus.go
```diff
@@ -0,0 +1,271 @@
...
+	Scraper struct {
+		meter                       metric.Meter
+		producer                    *metrics.MetricsProducer
+		previousCounterMetricValues map[string]DataPoint
+		cancelFunc                  context.CancelFunc
+	}
+)
+
+func NewScraper(meterProvider metric.MeterProvider, producer *metrics.MetricsProducer) *Scraper {
+	meter := meterProvider.Meter("nginx-agent-prometheus-scraper", metric.WithInstrumentationVersion("v0.1"))
```
oliveromahony, line 62: nginx-agent-prometheus-scraper is probably a const as well

internal/bus/message_pipe.go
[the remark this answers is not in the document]
dhurley: We can't handle the error since its the message bus library that is calling it

internal/datasource/metric/provider.go
```diff
@@ -0,0 +1,61 @@
...
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+)
+
+const (
+	// tenantID is static for now.
+	tenantID         = "7332d596-d2e6-4d1e-9e75-70f91ef9bd0e"
+	serviceNamespace = "nginx"
+	readInterval     = 10 * time.Second
```
oliveromahony: can we add this to the centralized configuration?

internal/datasource/metric/producer.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Dumps the latest recorded metrics and reinitializes the metrics array.
+func (hp *MetricsProducer) Produce(context.Context) ([]metricdata.ScopeMetrics, error) {
+	hp.metricsLock.Lock()
+	defer hp.metricsLock.Unlock()
+
+	scopeMetrics := []metricdata.ScopeMetrics{
+		{
+			Scope: instrumentation.Scope{
+				Name:    "github.com/agent/v3",
```
oliveromahony, line 59: maybe worth adding this to the Centralized config as well, under meta

olli-holmala:
> Just wondering as well should metric be under datasource? Or should it be under internal?

Yeah it might be a good idea actually 🤔 I'll move it to `internal`

internal/bus/message_pipe.go
```diff
@@ -163,8 +176,16 @@ func (p *MessagePipe) GetPlugins() []Plugin {
 }
 
 func (p *MessagePipe) initPlugins() {
+	var err error
 	for _, r := range p.plugins {
-		r.Init(p)
+		initErr := r.Init(p)
+		if initErr != nil {
```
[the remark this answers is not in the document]
olli-holmala: Sure, I'll switch to individual logging 👍

internal/datasource/metric/exporter.go
```diff
@@ -0,0 +1,83 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create grpc connection: %v", err)
+	}
+
+	return otlpmetricgrpc.New(ctx, otlpmetricgrpc.WithGRPCConn(conn), otlpmetricgrpc.WithTimeout(7*time.Second))
+}
+
+// Returns a OTel exporter that transmits via HTTP.
+func NewHTTPExporter(ctx context.Context, conf config.Metrics) (*otlpmetrichttp.Exporter, error) {
```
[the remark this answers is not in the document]
olli-holmala: We don't currently, but I suppose we want to support HTTP exporters too later. I'm gonna try to make the Exporter side of things more generic and enable using this one too.

internal/bus/topics.go
```diff
@@ -10,4 +10,5 @@ package bus
 const (
 	OS_PROCESSES_TOPIC = "os-processes"
 	INSTANCES_TOPIC    = "instances"
+	METRICS_TOPIC      = "metrics"
```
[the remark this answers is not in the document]
olli-holmala: We discussed this at the office and I'm gonna refactor by moving OTel exporting logic from `prometheus.go` to the metrics plugin whenever a `metrics` event is received.

internal/bus/message_pipe.go
```diff
@@ -163,8 +176,16 @@ func (p *MessagePipe) GetPlugins() []Plugin {
...
+			err = errors.Join(err, initErr)
+		}
+	}
+
+	if err != nil {
+		slog.Error("plugin failures occurred", "errors", err)
```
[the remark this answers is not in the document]
olli-holmala: I'll set up a logging conventions page.

internal/datasource/metric/provider.go
olli-holmala: This is already added to the config, this `const` is actually unused. The code below also already uses the value from the config 😄 

Removed the `const`!

internal/datasource/metric/producer.go
olli-holmala: I don't think this should be configurable. This is the signature of `instrumentation.Scope`:
```go
type Scope struct {
	// Name is the name of the instrumentation scope. This should be the
	// Go package name of that scope.
	Name string
	// Version is the version of the instrumentation scope.
	Version string
	// SchemaURL of the telemetry emitted by the scope.
	SchemaURL string
}
```

olli-holmala:
> can we pull in https://github.com/nginx/agent/blob/main/test/performance/plugins_test.go and something similar to https://github.com/nginx/agent/blob/main/src/core/mock_pipe.go because they will be useful (don't mind if they are in a separate PR)

I'll get this done in a follow-up PR 👌

internal/metrics/model.go
```diff
@@ -0,0 +1,28 @@
...
+		Description string
+		Values      []DataPoint
+	}
+
+	// A single data point for an entry. An entry can have multiple points.
+	DataPoint struct {
+		Name   string
+		Labels map[string]string
+		Value  float64
+	}
```
olli-holmala, line 20: I'm not quite happy with having these intermediate types that we very quickly transform into OTel data in `metrics.go`. I'm looking into getting rid of these altogether, and using `metricdata.DataPoint`s directly instead, which we could then push out in `metrics.go` to the appropriate meters.

internal/plugin/metrics.go
```diff
@@ -0,0 +1,303 @@
...
+
+	DataSource interface {
+		// Starts the data source and begins data collection at a regular interval. Metrics should be pushed to the
+		// MessagePipe using the metrics topic, as they will be handled in the metrics plugin.
+		Start(ctx context.Context, updateInterval time.Duration) error
+		// Stops the data source from collecting data.
+		Stop()
+		// Getter for the type of data source.
+		Type() string
+	}
```
olli-holmala, line 44: I would also like to potentially change this contract, as it does not clearly define how a data source should behave (in this case, push metrics to the metrics topic at the configured update interval). I also want to make sure that having a push model is better than a pull model for our use cases.

dhurley:
Might be good to have 2 packages under `internal/metrics/` called `source` and `exporter`. So prometheus can go under source and the OTel code can go under exporter.

internal/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+	Process(...*bus.Message)
+	Run()
+	Context() context.Context
+	GetPlugins() []bus.Plugin
+	IsPluginAlreadyRegistered(string) bool
+}
+
+// Used for deserializion of parsed Prometheus data.
+type Scraper struct {
+	bus bus.MessagePipeInterface
```
dhurley, line 49: not sure if we should be passing the bus in here. I think the bus should only be used by plugins and not passed to other places in the codebase.

internal/plugin/metrics.go
```diff
@@ -0,0 +1,303 @@
...
+
+const otelServiceName = "nginx-agent"
+
+type (
+	shutdownFunc func()
+
+	DataSource interface {
+		// Starts the data source and begins data collection at a regular interval. Metrics should be pushed to the
+		// MessagePipe using the metrics topic, as they will be handled in the metrics plugin.
+		Start(ctx context.Context, updateInterval time.Duration) error
```
dhurley, line 39: Instead of a Start function would a Collect function be better. So instead of each DataSource having its own ticker, the metrics plugin have its own ticker that would call the Collect function for each DataSource. The Collect function would return an array of DataEntry which the metrics plugin could be aggregate together and send to the exporter.

internal/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (s *Scraper) Type() string {
+	return DataSourceType
+}
+
+func scrapeEndpoint(target string) []metrics.DataEntry {
+	resp, err := http.Get(target)
+	if err != nil {
+		slog.Error("GET to Prometheus HTTP scrape endpoint failed", "endpoint", target)
```
dhurley, line 111:
```suggestion
		slog.Error("GET request to Prometheus HTTP scrape endpoint failed", "endpoint", target)
```

internal/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+		targets = strings.Split(value, ",")
+	}
+
+	for _, target := range targets {
+		resp, err := http.DefaultClient.Get(strings.TrimSpace(target))
+		if err != nil {
+			return fmt.Errorf("could not find prometheus endpoint %s: %w", target, err)
+		}
+
+		log.Printf("found prometheus endpoint %v", resp.Request.URL)
```
dhurley, line 79:
```suggestion
		slog.Debug("Found prometheus endpoint", "endpoint", resp.Request.URL)
```

internal/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+					msgs = append(msgs, e.ToBusMessage())
+				}
+				s.bus.Process(msgs...)
+			}
+		}
+	}
+}
+
+func (s *Scraper) Stop() {
+	slog.Info("prometheus data source shutting down")
```
dhurley, line 100:
```suggestion
	slog.Info("Prometheus data source shutting down")
```

internal/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+		} else if line != "" {
+			labels := map[string]string{}
+			splitMetric := strings.SplitN(line, " ", 3)
+			nameAndLabels := splitMetric[0]
+			splitNameAndLabels := strings.Split(nameAndLabels, "{")
+
+			pointName := splitNameAndLabels[0]
+			var pointErr error
+			if len(splitNameAndLabels) > 1 {
+				labelsJson := strings.ReplaceAll("{\""+splitNameAndLabels[1], "=", "\":")
```
dhurley, line 150: Maybe we can extract the contents of this if statement into its own function. Something like `parseLabels()` just to make the code more readable.

internal/plugin/metrics.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return fmt.Errorf("data source discovery failed: %w", err)
+	}
+
+	return m.startDataSources()
+}
+
+// Stops the metrics plugin. Required for the `Plugin` interface.
+func (m *Metrics) Close() error {
+	slog.Info("stopping all metrics data sources")
```
dhurley, line 113:
```suggestion
	slog.Info("Stopping all metrics data sources")
```

internal/plugin/metrics.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+}
+
+// Processes an incoming Message Bus message in the plugin. Required for the `Plugin` interface.
+func (m *Metrics) Process(msg *bus.Message) {
+	switch msg.Topic {
+	case bus.METRICS_TOPIC:
+		entry, ok := msg.Data.(metrics.DataEntry)
+		if !ok {
+			slog.Warn("failed to convert metrics event to a data entry")
```
dhurley, line 138: If the datatype is wrong we should return at this stage
```suggestion
			slog.Warn("Failed to convert metrics event to a data entry")
			return
```

internal/plugin/metrics.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// Calls Start() on each data source.
+func (m *Metrics) startDataSources() error {
+	for _, metricsSrc := range m.Sources {
+		go func(ds DataSource) {
+			err := ds.Start(m.pipe.Context(), m.conf.Metrics.ReportInterval)
+			if err != nil {
+				// Probably need to figure out how to handle these errors better (e.g. with a `chan error`).
+				slog.Error("failed to start metrics source", "error", err)
```
dhurley, line 194:
```suggestion
				slog.Error("Failed to start metrics source", "error", err)
```

internal/plugin/metrics.go
```diff
@@ -0,0 +1,303 @@
...
+	gauge := metrics.NewFloat64Gauge()
+
+	for _, point := range de.Values {
+		_, err := meter.Float64ObservableGauge(
+			de.Name,
+			metric.WithDescription(de.Description),
+			metric.WithFloat64Callback(gauge.Callback),
+		)
+		if err != nil {
+			slog.Error("failed to initialize OTel gauge", "error", err)
```
dhurley, line 211:
```suggestion
			slog.Error("Failed to initialize OTel gauge", "error", err)
```

internal/plugin/plugin_manager.go
```diff
@@ -23,6 +23,14 @@ func LoadPlugins(agentConfig *config.Config, slogger *slog.Logger) []bus.Plugin
 
 	instanceMonitor := NewInstanceMonitor(&InstanceMonitorParameters{})
 
+	if agentConfig.Metrics != nil {
+		metrics, err := NewMetrics(*agentConfig)
+		if err != nil {
+			slogger.Error("failed to initialize metrics plugin")
```
dhurley, line 29:
```suggestion
			slogger.Error("Failed to initialize metrics plugin", "error", err)
```

internal/config/types.go
```diff
@@ -16,6 +16,7 @@ type Config struct {
 	ProcessMonitor ProcessMonitor `mapstructure:"process_monitor" yaml:"-"`
 	DataplaneAPI   DataplaneAPI   `mapstructure:"dataplane_api" yaml:"-"`
 	Client         Client         `mapstructure:"client" yaml:"-"`
+	Metrics        *Metrics       `mapstructure:"metrics" yaml:"-"`
```
dhurley, line 19: Should this be a pointer? Seems inconsistent with the other parameters

olli-holmala: It is inconsistent. The idea is that you could omit the `Metrics` config in the config file, which would mean that the metrics plugin would not start. We could use this same practice for the other plugins, but it sorta depends on how we want to go about this.

I suppose the alternative would be to have a list of plugins that we activate in the configuration. My instinct is to go with option #1, but you guys have more experience with this, what do you think? @dhurley, @oliveromahony

internal/plugin/metrics.go
olli-holmala: Sure, we can do that. That would just mean that we'd have a common refresh interval for all sources instead of a configurable refresh interval per data source but that's likely fine because we are exporting everything at the same rate anyway.

I'll get to work on this!

internal/metrics/model.go
olli-holmala: @dhurley made the fair point that we will likely want to support exporting to other types than OTel in the future, therefore having tight coupling to OTel might not be a great idea. So we will support this internal model, and then perform transformations to the correct target types based on the target data sink.

internal/metrics/prometheus/prometheus.go
olli-holmala: Yeah, this will be removed after we change the data source contract as you suggested [here](https://github.com/nginx/agent/pull/544#discussion_r1469848053).

internal/metrics/prometheus/prometheus.go
Dean-Coakley: Surely we should also log `err` value here?

(OR similar?)
```suggestion
		slog.Error("GET request to Prometheus HTTP scrape endpoint failed", "endpoint", target, "error", err)
```

internal/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+
+func scrapeEndpoint(target string) []metrics.DataEntry {
+	resp, err := http.Get(target)
+	if err != nil {
+		slog.Error("GET to Prometheus HTTP scrape endpoint failed", "endpoint", target)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		slog.Warn("failed to parse Prometheus scrape endpoint response body")
```
Dean-Coakley, line 116: Should include `err` value in log and maybe return early? Nothing to try to parse if it cannot be scraped

```suggestion
		slog.Warn("failed to parse Prometheus scrape endpoint response body", "error", err)
		return
```

internal/metrics/prometheus/prometheus.go
olli-holmala: The MessagePipe is now only used by the Metrics Plugin, so this is now resolved.

internal/plugin/metrics.go
olli-holmala: `DataSource` was renamed into a `MetricsProducer`, which can `Produce()` metrics in the way that you describe `Collect` here. I chose `Producer` to make it perhaps more clear that it's following the Producer-Consumer pattern.

internal/metrics/prometheus/prometheus.go
olli-holmala: This code has been radically refactored, will have to scan through again to see if I've missed something like this.

internal/metrics/sources/prometheus/exporter.go
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) F5, Inc.
```
dhurley, line 1: I'd combine and move `exporter.go` and `otel.go` into their own folder like `metrics/exporter/otel`

internal/plugin/metrics_plugin.go
```diff
@@ -0,0 +1,224 @@
...
+	// Data sources can also be passed in as options on plugin init.
+	if _, ok := m.producers[model.Prometheus]; !ok {
+		// Will always use the env var for the time being.
+		m.producers[model.Prometheus] = prometheus.NewScraper(make([]string, 0))
+	}
+}
+
+func (m *Metrics) startExporters(ctx context.Context) error {
+	for srcType := range m.producers {
+		// How do we know which exporter is used by which source?
```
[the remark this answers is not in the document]
olli-holmala: Oh yeah, you're right, we would want each Exporter type to be a singleton, I'll get that fixed. But I suppose we will have other exporters than OTel exporters, right?

internal/metrics/sources/prometheus/exporter.go
olli-holmala: Yeah true, that should be pulled out of the `prometheus` dir. I'm not a huge fan of adding that much nesting though, if we can avoid it.

internal/metrics/prometheus/prometheus.go
olli-holmala: Fixed this in the refactor, good catch!

internal/metrics/sources/prometheus/exporter.go
dhurley: putting it under just `metrics/exporter` probably makes more sense

internal/metrics/otel.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return otlpmetrichttp.New(
+		ctx,
+		otlpmetrichttp.WithInsecure(),
+		otlpmetrichttp.WithEndpoint(target),
+		// WithTimeout sets the max amount of time the Exporter will attempt an
+		// export.
+		otlpmetrichttp.WithTimeout(7*time.Second),
+		otlpmetrichttp.WithRetry(otlpmetrichttp.RetryConfig{
```
oliveromahony, line 70: we probably should externalize these configurations

internal/metrics/otel_test.go
```diff
@@ -0,0 +1,35 @@
...
+func TestGRPCExporter_Constructor(t *testing.T) {
+	conf := config.Metrics{
+		OTelExporterTarget: "dummy-target",
+		ReportInterval:     1 * time.Second,
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	_, err := NewGRPCExporter(ctx, conf)
```
oliveromahony, line 25: can you assert on anything in otlpmetricgrpc.Exporter return?

internal/metrics/sources/prometheus/convert.go
```diff
@@ -0,0 +1,238 @@
...
+			Value:      value,
+		})
+	}
+
+	return &metricdata.Gauge[N]{
+		DataPoints: dataPoints,
+	}
+}
+
+const (
```
oliveromahony, line 132: prefer consts to be used near the top of the file

internal/metrics/sources/prometheus/prometheus.go
```diff
@@ -0,0 +1,265 @@
...
+	// If entry already has content, then we add it to the result slice and reset.
+	if builder.CanBuild() {
+		result = append(result, builder.Build())
+		*builder = model.NewEntryBuilder(model.WithSourceType(model.Prometheus))
+	}
+
+	return result
+}
+
+const (
```
oliveromahony, line 184: same comments about the consts here, probably more readable at the top of the file

internal/model/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+	return eb.entry
+}
+
+func (eb *EntryBuilder) CanBuild() bool {
+	return eb.entry.Name != "" && eb.entry.Description != "" && eb.entry.SourceType != 0 && eb.entry.Type != Unknown
+}
+
+type MetricsSourceType int
+
+const (
```
oliveromahony, line 126: ideally would like one type block and one const block with both grouped together near the top of the file. Seems to be a lot mixed around this file

internal/model/metrics.go
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) F5, Inc.
+//
+// This source code is licensed under the Apache License, Version 2.0 license found in the
+// LICENSE file in the root directory of this source tree.
```
oliveromahony, line 4: would like to see unit tests for some of the functions in this file, even if you can't test everything easily you can test some of it

internal/metrics/otel.go
olli-holmala: When you say "externalize", do you mean read them from our config file?

internal/model/metrics.go
olli-holmala: Yeah I'll group em' together 👍

internal/metrics/sources/prometheus/prometheus.go
olli-holmala: https://github.com/nginx/agent/pull/544#discussion_r1483071548

internal/metrics/sources/prometheus/convert.go
olli-holmala: https://github.com/nginx/agent/pull/544#discussion_r1483071548

internal/model/metrics.go
olli-holmala: Adding `metrics_test.go`, which has 100% coverage

image

internal/metrics/otel_test.go
olli-holmala: The `Exporter` struct only has private fields, so can't really assert against them.
```go
// Exporter is a OpenTelemetry metric Exporter using gRPC.
type Exporter struct {
	// Ensure synchronous access to the client across all functionality.
	clientMu sync.Mutex
	client   interface {
		UploadMetrics(context.Context, *metricpb.ResourceMetrics) error
		Shutdown(context.Context) error
	}

	temporalitySelector metric.TemporalitySelector
	aggregationSelector metric.AggregationSelector

	shutdownOnce sync.Once
}
```
I suppose I can add a nil check, since it returns a pointer to the `Exporter`.

internal/metrics/otel.go
olli-holmala: The gRPC configurations are now taken from the Agent's config file, but the HTTP ones are not, as we don't currently support HTTP OTel connections.

internal/plugin/metrics_plugin.go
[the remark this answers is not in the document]
olli-holmala: Discussed this at the office and the answer was 'yes', we want to support more exporters but each exporter of a given type should be a singleton, instead of instantiating a new exporter per source. This is now fixed.

internal/plugin/metrics_plugin.go
```diff
@@ -0,0 +1,257 @@
...
+
+		m.exporters[exporter.Type()] = exporter
+	}
+
+	return nil
+}
+
+func (m *Metrics) startExporters(ctx context.Context) {
+	for expType, exp := range m.exporters {
+		slog.Info("Starting export", "producer_type", expType.String())
```
dhurley, line 176:
```suggestion
		slog.Info("Starting export", "exporter_type", expType.String())
```

internal/plugin/metrics_plugin.go
```diff
@@ -0,0 +1,257 @@
...
+	for i, e := range entries {
+		busMsgs[i] = e.ToBusMessage()
+	}
+
+	m.pipe.Process(busMsgs...)
+
+	return failedAttempts
+}
+
+//##########################
```
dhurley, line 243: Not sure about having these MetricsOptions comments here

scripts/testing/docker-compose_otel-collector.yaml
```diff
@@ -0,0 +1,73 @@
+version: '3.9'
```
dhurley, line 1: should the files in `scripts/testing/` be in `test/integration/metrics/`?

test/integration/metrics/prometheus_test.go
```diff
@@ -0,0 +1,974 @@
+// Copyright (c) F5, Inc.
```
dhurley, line 1: I think we have done integrations two different ways if you compare these tests to the ones here https://github.com/nginx/agent/blob/v3/test/integration/dataplane_api_test.go. Think we need to have a discussion on what we want define as unit test or an integration tests.

internal/config/config.go
```diff
@@ -99,29 +92,63 @@ func registerFlags() {
...
-	fs.String(ConfigDirectoriesConfigKey, "/etc/nginx:/usr/local/etc/nginx:/usr/share/nginx/modules",
+	fs.String(DataplaneAPIHostKey, "", "The host used by the Dataplane API.")
+	fs.Int(DataplaneAPIPortKey, 0, "The desired port to use for NGINX Agent to expose for HTTP traffic.")
+	fs.Duration(ClientTimeoutKey, time.Minute, "Client timeout")
+	fs.String(ConfigDirectoriesKey, "/etc/nginx:/usr/local/etc/nginx:/usr/share/nginx/modules",
 		"Defines the paths that you want to grant nginx-agent read/write access to."+
 			" This key is formatted as a string and follows Unix PATH format")
+	fs.Duration(
+		MetricsProduceIntervalKey, DefMetricsProduceInterval,
+		"The interval for how often NGINX Agent queries metrics from its sources.",
```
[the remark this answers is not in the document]
olli-holmala: My version is actually gramatically correct 😅 "its" is the genetive form of "it" whereas "it's" is a contraction for "it is": https://www.merriam-webster.com/grammar/when-to-use-its-vs-its

internal/config/config.go
```diff
@@ -99,29 +92,63 @@ func registerFlags() {
...
+		OTelExporterBufferLengthKey, DefOTelExporterBufferLength,
+		"The length of the OTel Exporter's buffer for metrics.",
+	)
+	fs.Int(
+		OTelExporterExportRetryCountKey, DefOTelExporterExportRetryCount,
+		"How many times an OTel Export is retried in the event of failure.",
+	)
+	fs.Duration(
+		OTelExporterExportIntervalKey, DefOTelExporterExportInterval,
+		"The interval for how often NGINX Agent attempts to send the contents of its OTel Exporter's buffer.",
```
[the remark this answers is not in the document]
olli-holmala: Ditto: https://github.com/nginx/agent/pull/544#discussion_r1499014020

internal/plugin/metrics_plugin.go
```diff
@@ -0,0 +1,257 @@
...
+
+	m.discoverSources()
+
+	err := m.createExporters(ctx)
+	if err != nil {
+		return fmt.Errorf("could not start exporters: %w", err)
+	}
+
+	if len(m.producers) != len(m.exporters) {
+		slog.Warn("number of metrics producers and exporters not equal",
```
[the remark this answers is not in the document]
olli-holmala: True, this check was part of an older implementation and should no longer be there. Removing ❌

internal/plugin/metrics_plugin.go
olli-holmala: True, removing

test/integration/metrics/prometheus_test.go
olli-holmala: Yeah I think you're right, I always get confused about the testing terminology 😅 Now that I think about it, maybe my tests are more like component tests rather than integretaion tests 🤔 

We'll discuss it at the office anyway 👌

internal/metrics/export/otel.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return otlpmetricgrpc.New(
+		ctx, otlpmetricgrpc.WithGRPCConn(conn), otlpmetricgrpc.WithTimeout(conf.ConnTimeout),
+	)
+}
+
+// NewHTTPExporter returns a OTel export that transmits via HTTP.
+// nolint: gomnd
+func NewHTTPExporter(ctx context.Context) (*otlpmetrichttp.Exporter, error) {
```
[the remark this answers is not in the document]
olli-holmala: Yeah I'll just remove it for now, won't be difficult to add it back if we need it later.

olli-holmala:
TODO: 

- [ ] Rebase commits before merge.

Makefile
```diff
@@ -132,3 +132,8 @@ local-rpm-package: ## Create local rpm package
 	@CGO_ENABLED=0 GOARCH=${OSARCH} GOOS=linux $(GOBUILD) -o $(BUILD_DIR)/$(BINARY_NAME) -ldflags=${LDFLAGS} $(PROJECT_DIR)/${PROJECT_FILE}
 	ARCH=$(OSARCH) VERSION=$(shell echo $(VERSION) | tr -d 'v') $(GORUN) $(NFPM) pkg --config ./scripts/packages/.local-nfpm.yaml --packager rpm --target ./build/$(PACKAGE_PREFIX)-$(shell echo ${VERSION} | tr -d 'v')-SNAPSHOT-$(COMMIT).rpm;
 
+integration: integration-metrics
```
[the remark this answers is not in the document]
olli-holmala: Removed this target altogether as a result of our discussion at the office.

Makefile
```diff
@@ -132,3 +132,8 @@ local-rpm-package: ## Create local rpm package
...
+
+integration-metrics:
```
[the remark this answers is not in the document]
olli-holmala: See: https://github.com/nginx/agent/pull/544#discussion_r1499091280

internal/config/config.go
```diff
@@ -185,30 +212,70 @@ func normalizeFunc(f *flag.FlagSet, name string) flag.NormalizedName {
...
+		m.OTelExporter = &otelExp
+	}
+
+	if viperInstance.IsSet(PrometheusSrcKey) {
+		var prometheusSrc PrometheusSource
+		err := viperInstance.UnmarshalKey(PrometheusSrcKey, &prometheusSrc)
+		if err == nil {
+			m.PrometheusSource = &prometheusSrc
+		} else {
+			slog.Info("metrics configuration: no Prometheus source configured", "error", err)
```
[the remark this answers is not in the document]
olli-holmala: Yeah, changing to an error log.

internal/metrics/source/prometheus/convert.go
```diff
@@ -0,0 +1,254 @@
+// Copyright (c) F5, Inc.
```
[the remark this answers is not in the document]
olli-holmala: That would ideally be the case, but I'm not sure that's true? For example a Prometheus histogram is represented as a series of datapoints where:
* Datapoints whose name is suffixed with `_bucket` represent a 'bucket' data point
* Datapoints whose name is suffixed with `_sum` represent a 'sum' data point
* Datapoints whose name is suffixed with `_count` represent a 'count' data point

`convert.go` assumes the above for its parsing. 

Will we force the other data sources to transform their histograms to conform to the model above? If yes, then I would agree that we move `convert.go` to `internal/metrics`, otherwise I would say each exporter will need its own `convert.go`.

internal/metrics/source/prometheus/convert.go
```diff
@@ -0,0 +1,254 @@
...
+const (
+	bucketSuffix = "_bucket"
+	sumSuffix    = "_sum"
+	countSuffix  = "_count"
+)
+
+type InstrumentPointConverter[N float64 | int64] func(input model.DataEntry) metricdata.Aggregation
+
+// ConvertPrometheus converts metrics data points from our internal model to OTel format.
+func ConvertPrometheus(input model.DataEntry) (metricdata.Metrics, error) {
```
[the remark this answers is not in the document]
olli-holmala: This rename depends on: https://github.com/nginx/agent/pull/544#discussion_r1499114417

test/integration/metrics/prometheus_test.go
olli-holmala: We agreed that I will move the content of these tests to the unit tests, as integration tests should deploy in Docker both the Agent and the app it is integrating with.

scripts/testing/docker-compose_otel-collector.yaml
olli-holmala: I'll leave these where they are for now, since there are no tests yet that use em'. As discussed at the office, I'll get to work on proper metrics integration tests in a separate PR, and I'll get these moved to `test/integration/metrics/` then.

internal/metrics/otel.go
olli-holmala: The HTTPExporter was removed per: https://github.com/nginx/agent/pull/544#discussion_r1497688634

internal/datasource/metric/exporter.go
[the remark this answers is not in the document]
olli-holmala: This was deleted per: https://github.com/nginx/agent/pull/544#discussion_r1497688634 and this comment.

internal/metrics/source/prometheus/convert.go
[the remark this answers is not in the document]
olli-holmala: Then there's the issue of how the temporality of metrics for the metrics source. Different sources may use different temporality, which means we would need to at least be able to somehow parameterize that in the converter.

I would be inclined to leave this as is for now and then look at what a generic converter would look like when we add our next source of metrics.

scripts/testing/docker-compose_otel-collector.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - "1888:1888"   # pprof extension
+      - "13133:13133" # health_check extension
+      - "4317:4317"   # OTLP/2 gRPC receiver
+      - "4318:4318"   # OTLP/2 http receiver
+      - "55679:55679" # zpages extension
+      - "8889:8889"   # prometheus exporter
+    networks:
+      - monitoring
+
+  # grafana:
```
oliveromahony, line 29: is this required? If not, we could add a different compose yaml so we can run it separately?

aphralG (approved):

scripts/testing/docker-compose_otel-collector.yaml
olli-holmala: I removed it for now, I'll add a separate docker compose file later (once I'm sure it works).

internal/metrics/source/prometheus/convert.go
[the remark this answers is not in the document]
olli-holmala: We discussed this at the office and decided that we'll keep it as is for now and then refactor to a common converter if it makes sense when we add the next data source.

internal/config/types.go
olli-holmala: We will have a separate discussion on this and align all the configurations to a similar style in a follow-up PR.

internal/metrics/export/otel.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	res, err := resource.New(ctx,
+		// Keep the default detectors
+		resource.WithTelemetrySDK(),
+		// Add your own custom attributes to identify your application
+		resource.WithAttributes(
+			semconv.ServiceNamespaceKey.String(serviceNamespace),
+			semconv.ServiceNameKey.String(serviceName),
+			semconv.ServiceInstanceIDKey.String(tenantID),
```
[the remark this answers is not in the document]
olli-holmala: Discussed this at the office, decided to generate a unique UID for now so that each Agent at least has a unique identifier. The UUID will not survive an Agent restart, so we will need to figure out a way to make it persistent in the future.

internal/metrics/export/otel.go
```diff
@@ -0,0 +1,242 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/nginx/agent/v3/internal/config"
+	"github.com/nginx/agent/v3/internal/model"
+	"github.com/sagikazarmark/slog-shim"
```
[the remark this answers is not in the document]
olli-holmala: Great catch!

dhurley (approved):

PR #545, opened 2024-01-25
Validate config & reload nginx

aphralG:
### Proposed changes

1. Add ConfigInterface to mock 
2. Add Config Writer
   - Write Config & Cache 
   - Reads Cache
   - Set DataplaneConfig (NginxConfig in this case)
3. Add NginxConfig 
   - Reload Nginx 
   - Validate Config 
4. Change process.go to set the exePath in nginxInfo for use in Reload() 
5. Unit Tests for the above changes

### Checklist


internal/service/instance_test.go
```diff
@@ -32,3 +38,24 @@ func TestInstanceService_GetInstances(t *testing.T) {
...
+
+	return tenantId, instanceId, err
+}
+
+func createProtoTime(timeString string) (*timestamppb.Timestamp, error) {
+	time, err := time.Parse(time.RFC3339, timeString)
+	protoTime := timestamppb.New(time)
+
+	return protoTime, err
+}
```
olli-holmala, line 61: Are these used anywhere?

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,190 @@
...
+		cachePath          string
+		dataplaneConfig    nginx.DataplaneConfigInterface
+	}
+
+	// map of files with filepath as key
+	FileCache = map[string]*instances.File
+)
+
+func NewConfigWriter(configWriterParameters *ConfigWriterParameters, instanceId string) *ConfigWriter {
+	cachePath := fmt.Sprintf("/var/lib/nginx-agent/config/%v/cache.json", instanceId)
```
olli-holmala, line 57: nit: maybe make `"/var/lib/nginx-agent/config/%v/cache.json"` a `const`?

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return previousFileCache, err
+}
+
+func isFilePathValid(filePath string) (validPath bool) {
+	return filePath != "" && !strings.HasSuffix(filePath, "/")
+}
+
+func doesFileRequireUpdate(previousFileCache FileCache, fileData *instances.File) (updateRequired bool) {
+	if previousFileCache != nil && len(previousFileCache) > 0 {
```
olli-holmala, line 185: I think my IDE complained about this earlier, as a `nil` `map` is the same as `len(map) == 0`, so you could just simplify this to:
```suggestion
	if len(previousFileCache) > 0 {
```

internal/datasource/config/config_writer_test.go
```diff
@@ -17,23 +19,20 @@ import (
...
 	"github.com/nginx/agent/v3/api/grpc/instances"
 	"github.com/nginx/agent/v3/internal/client"
-	datasource_os "github.com/nginx/agent/v3/internal/datasource/os"
+	"github.com/nginx/agent/v3/internal/datasource/nginx"
 
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-func TestUpdateInstanceConfig(t *testing.T) {
+func TestWriteConfig(t *testing.T) {
```
olli-holmala, line 27: Just checked out to your branch and ran the unit tests, looks like the `Write()` function is not being tested for cases where latest version is not already on disk. Think we should make sure we cover those too.

Also, the code coverage percentage (see bottom of screenshot) is skewed, because the coverage calculation now includes our mock code, which we obviously aren't going to unit test.

image

internal/service/instance_test.go
aphralG: Nope, forgot to remove them when we changed where everything was

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/google/uuid"
+	"github.com/nginx/agent/v3/api/grpc/instances"
+	"github.com/nginx/agent/v3/internal/client"
+	"github.com/nginx/agent/v3/internal/datasource/nginx"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.7.0 -generate
+//counterfeiter:generate -o mock_config_writer.go . ConfigWriterInterface
+//go:generate sh -c "grep -v github.com/nginx/agent/v3/internal/datasource/config mock_config_writer.go | sed -e s\\/config\\\\.\\/\\/g > mock_config_writer_fixed.go"
+//go:generate mv mock_config_writer_fixed.go mock_config_writer.go
```
olli-holmala, line 28: Probably should do this the same way as Donal is in his PR: https://github.com/nginx/agent/pull/547

olli-holmala (approved):

internal/datasource/config/config_writer.go
aphralG: I think I might need the other changes Donal made for that to work so might wait till I that is merged in. Think I need the change from `mock` to `fake` could be wrong though

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return fmt.Errorf("error writing cache to %s: %w", cw.cachePath, err)
+	}
+
+	return err
+}
+
+func (cw *ConfigWriter) SetDataplaneConfig(dataplaneConfig nginx.DataplaneConfigInterface) error {
+	cw.dataplaneConfig = dataplaneConfig
+	return nil
```
dhurley, line 139: Do you need the return here if its always nil?

internal/datasource/nginx/config.go
```diff
@@ -0,0 +1,41 @@
...
+func NewNginxConfig() *NginxConfig {
+	return &NginxConfig{}
+}
+
+func (nc NginxConfig) Validate() error {
+	out, err := exec.Command("nginx", "-t").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("NGINX config test failed %w: %s", err, out)
+	}
+	slog.Info("NGINX config tested", "output", out)
```
dhurley, line 29: Need to validate the response from the command as well like what we did here https://github.com/nginx/agent/blob/main/src/core/nginx.go#L314

aphralG: Done but I left out the TreatWarnings as errors part of that function

PR #546, opened 2024-01-26
refactor: data sources generate OTel data

olli-holmala:
### Proposed changes

This is a refactor PR, which presents an alternative way to produce OTel data for our data sources.

In this approach, the data source transforms its source metrics into OTel data points depending on the type of source content. OTel metrics are then propagated forward to the OTel Producer using a callback, which then periodically sends it out to our exporter (in this case via gRPC).

Using this method we don't need to keep track of e.g. previous counter values, but instead define the temporality of our data as `Delta` (instead of `Cumulative`), so OTel should understand that they should not be added to previously received values.

Usage of our own internal message bus is completely removed in this implementation.

I will still test this out to ensure that it works the way we want it to. We can opt to not go down this route too, just wanted to make sure we think about our options before pulling the trigger.

### Checklist


olli-holmala:
We opted to go with another approach than what was proposed here (although some elements were incorporated into PR-544), so I'll close this PR.

PR #547, opened 2024-01-26
Add config plugin for parsing dataplane configuration.

dhurley:
### Proposed changes

- Add config plugin for parsing dataplane configuration.
- Refactored project directory structure.
- Refactored how fake mocks are generated.

### Checklist


internal/plugin/dataplane_server.go
```diff
@@ -113,25 +110,26 @@ func (dps *DataplaneServer) run(ctx context.Context) {
...
+	slog.Debug("got instances", "instances", response)
+
 	ctx.JSON(http.StatusOK, response)
 }
 
-// (PUT /instances/{instanceId}/configurations)
+// PUT /instances/{instanceId}/configurations
 func (dps *DataplaneServer) UpdateInstanceConfiguration(ctx *gin.Context, instanceId string) {
-	slog.Debug("update instance configuration request", "instanceId", instanceId)
+	correlationId := uuid.New().String()
```
olli-holmala, line 131: I suppose in the future we'll need to extract the correlation ID from the request context, depending on how we will propagate correlation IDs.

internal/plugin/dataplane_server_test.go
```diff
@@ -136,44 +135,47 @@ func TestDataplaneServer_UpdateInstanceConfiguration(t *testing.T) {
...
-	assert.NoError(t, err)
-	assert.Equal(t, "Internal Server Error", result3.Message)
+	for _, test := range tests {
+		t.Run(test.name, func(tt *testing.T) {
+			res, err := performPutRequest(dataplaneServer, test.instanceId, data)
+			assert.NoError(tt, err)
+			assert.Equal(tt, test.expectedStatusCode, res.StatusCode)
+
+			resBody, err := io.ReadAll(res.Body)
+			assert.NoError(t, err)
```
olli-holmala, line 164:
```suggestion
			assert.NoError(tt, err)
```

internal/service/config/config.go
```diff
@@ -0,0 +1,26 @@
...
+	ParseConfig(instance *instances.Instance) (any, error)
+}
+
+type Config[C DataplaneConfig] struct {
+	DataplaneConfig C
+}
+
+func (c *Config[C]) ParseConfig(instance *instances.Instance) (any, error) {
+	return c.DataplaneConfig.ParseConfig(instance)
+}
```
olli-holmala, line 26: I don't fully understand what we accomplish here with generics. Are we trying to use [the bridge pattern](https://refactoring.guru/design-patterns/bridge)?

olli-holmala: In its current state, I'm not sure what the type parameter adds to this. I think using a simple `interface` has the same effect.
```suggestion
type DataplaneConfig interface {
	ParseConfig(instance *instances.Instance) (any, error)
}

type Config struct {
	DataplaneConfig DataplaneConfig
}

func (c *Config) ParseConfig(instance *instances.Instance) (any, error) {
	return c.DataplaneConfig.ParseConfig(instance)
}
```

Unless we want to parameterize the return type via type parameter, so something like this:
```suggestion
type ConfigType interface {
    InstanceConfig | MetricsConfig | K8sConfig // Add your config types here.
}

type DataplaneConfig[T ConfigType] interface {
	ParseConfig(instance *instances.Instance) (T, error)
}

type Config[T ConfigType] struct {
	DataplaneConfig DataplaneConfig[T]
}

func (c *Config[T]) ParseConfig(instance *instances.Instance) (T, error) {
	return c.DataplaneConfig.ParseConfig(instance)
}
```

internal/bus/fake_message_pipe.go
```diff
@@ -0,0 +1,135 @@
...
+		t.Fail()
+	}
+	return messagePipe
+}
+
+func ValidateMessages(t *testing.T, messagePipe *FakeMessagePipe, msgTopics []string) {
+	processedMessages := messagePipe.GetProcessedMessages()
+	if len(processedMessages) != len(msgTopics) {
+		t.Fatalf("expected %d messages, received %d: %+v", len(msgTopics), len(processedMessages), processedMessages)
+	}
```
olli-holmala, line 39: Do asserts not work here? Or does the test not fail cause it's in a function?
```suggestion
	assert.Equal(t, len(msgTopics), len(processedMessages))
```

internal/bus/fake_message_pipe.go
```diff
@@ -0,0 +1,135 @@
...
+	for idx, msg := range processedMessages {
+		if msgTopics[idx] != msg.Topic {
+			t.Errorf("unexpected message topic: %s :: should have been: %s", msg.Topic, msgTopics[idx])
+		}
```
olli-holmala, line 43: Ditto here
```suggestion
		assert.Equal(t, msgTopics[idx], msg.Topic, "unexpected message topic")
```

api/grpc/instances/instances.proto
```diff
@@ -23,12 +23,12 @@ message Meta {
     }
 }
 message NginxMeta {
-    string loadable_modules = 1;
```
oliveromahony, line 26: not sure about removing these?

internal/service/servicefakes/fake_config_service_interface.go
```diff
@@ -0,0 +1,236 @@
+// Code generated by counterfeiter. DO NOT EDIT.
+package servicefakes
```
oliveromahony, line 2: fakes not sufficient package name? service/servicefakes looks incorrect

oliveromahony: seems to be a counterfeiter convention :/

internal/datasource/nginx/process.go
```diff
@@ -0,0 +1,57 @@
+/**
```
oliveromahony, line 1: can you add a test for this file? At least  defaultToNginxCommandForProcessPath and sanitizeExeDeletedPath functions

internal/datasource/os/exec/exec.go
```diff
@@ -13,9 +13,7 @@ import (
 )
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.7.0 -generate
-//counterfeiter:generate -o mock_exec.go . ExecInterface
-//go:generate sh -c "grep -v github.com/nginx/agent/v3/internal/datasource/os/exec mock_exec.go | sed -e s\\/exec\\\\.\\/\\/g > mock_exec_fixed.go"
-//go:generate mv mock_exec_fixed.go mock_exec.go
+//counterfeiter:generate . ExecInterface
 type ExecInterface interface {
```
oliveromahony, line 17: can you add tests for exec.go?

internal/service/servicefakes/fake_config_service_interface.go
dhurley: we were doing some weird file parsing before so I changed it to the default that counterfeiter had because it was causing issues. Hence the weird package name

internal/datasource/os/exec/exec.go
dhurley: Don't know how I can test this one. Do you know of a way of mocking the `os/exec` package?

internal/service/config/config.go
oliveromahony: I think it should be the latter

api/grpc/instances/instances.proto
dhurley: we weren't populating them at the moment so I said I'd remove them for the time being until we start populating them with values. I can add them back in if you want.

internal/plugin/config.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	parsedConfig, err := c.configServices[request.Instance.GetInstanceId()].ParseInstanceConfiguration(request.CorrelationId, request.Instance)
+	if err != nil {
+		slog.Error("Unable to parse instance configuration", "correlationId", request.CorrelationId, "instanceId", request.Instance.InstanceId, "error", err)
+	} else {
+		switch config := parsedConfig.(type) {
+		case model.NginxConfigContext:
+			c.configServices[request.Instance.GetInstanceId()].SetConfigContext(config)
+		}
```
olli-holmala, line 78: I assume that a type switch is used here because we will be adding more contexts soon? Otherwise it could just be a type cast:
```suggestion
		pc, ok := parsedConfig.(model.NginxConfigContext)
		if !ok {
		   slog.Errorf("unknown config context type")
		} else {
		    c.configServices[request.Instance.GetInstanceId()].SetConfigContext(pc)
		}
```

Also, should there be a `default` case, which logs an unknown config type?

internal/bus/fake_message_pipe.go
dhurley: Actually this function isn't used so removing it.

internal/bus/fake_message_pipe.go
dhurley: Removing as well

internal/plugin/config.go
dhurley: yes there will be other contexts in the future. I'll add a default

internal/service/instance/nginx.go
```diff
@@ -0,0 +1,203 @@
...
+	re     = regexp.MustCompile(`(?P<name>\S+)/(?P<version>\S+)`)
+	plusre = regexp.MustCompile(`(?P<name>\S+)/(?P<version>\S+).\((?P<plus>\S+plus\S+)\)`)
+)
+
+type Info struct {
+	Version       string
+	PlusVersion   string
+	Prefix        string
+	ConfPath      string
+	Cfgf          map[string]interface{}
```
olli-holmala, line 35: I think this could use a comment on what this is, cuz I have no idea without reading how `Info` is used.

internal/service/instance/nginx.go
```diff
@@ -0,0 +1,203 @@
...
+
+	nginxProcesses := make(map[int32]*model.Process)
+	for _, p := range processes {
+		if isNginxProcess(p.Name, p.Cmd) {
+			nginxProcesses[p.Pid] = p
+		}
+	}
+
+	for _, nginxProcess := range nginxProcesses {
+		_, ok := nginxProcesses[nginxProcess.Ppid]
```
olli-holmala, line 67: We are iterating through `nginxProcesses`, so why are we checking whether `nginxProcess` is in `nginxProcesses`?

dhurley: we are checking that the parent of the nginx process is in the map or not to determine if its a worker or master process

internal/service/instance/nginx.go
```diff
@@ -0,0 +1,203 @@
...
+		if !ok {
+			if nginxProcess.Exe == "" {
+				nginxProcess.Exe = process.New(n.executer).GetExe()
+			}
+
+			if nginxProcess.Exe == "" {
+				slog.Debug("Unable to find NGINX exe", "pid", nginxProcess.Pid)
+				continue
+			}
```
olli-holmala, line 76: This `if` is the same as the one above?

dhurley: The `GetExe()` can also return an empty string so we need to check a second time

dhurley: probably should be moved into the first if statement to make it clearer

internal/service/instance/nginx.go
olli-holmala: Ah, sorry I missed the extra 'P' of `Ppid`. Was super confused there 😅

olli-holmala (approved):

aphralG (approved):

internal/datasource/os/exec/exec.go
oliveromahony: does this work? https://npf.io/2015/06/testing-exec-command/

oliveromahony (approved):

PR #551, opened 2024-01-31
validate config & reload nginx

aphralG:
### Proposed changes

1. Add ConfigInterface to mock 
2. Add Config Writer
   - Write Config & Cache 
   - Reads Cache
   - Set DataplaneConfig (NginxConfig in this case)
3. Add NginxConfig 
   - Reload Nginx 
   - Validate Config 
4. Change process.go to set the exePath in nginxInfo for use in Reload() 
5. Unit Tests for the above changes

### Checklist


api/grpc/instances/instances.proto
```diff
@@ -24,6 +24,7 @@ message Meta {
 }
 message NginxMeta {
     string config_path = 3;
```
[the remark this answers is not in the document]
olli-holmala: Should indexing start from 0?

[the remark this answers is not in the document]
dhurley: in proto I think it starts at 1 if i remember corectly

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/nginx/agent/v3/internal/service/config"
+
+	"github.com/google/uuid"
+	"github.com/nginx/agent/v3/api/grpc/instances"
+	"github.com/nginx/agent/v3/internal/client"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.7.0 -generate
+//counterfeiter:generate . ConfigWriterInterface
+type ConfigWriterInterface interface {
```
olli-holmala, line 28: nit: this could go in with all the other `type`s.

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,195 @@
...
+		currentFileCache   FileCache
+		cachePath          string
+		dataplaneConfig    config.DataplaneConfig
+	}
+
+	// map of files with filepath as key
+	FileCache = map[string]*instances.File
+)
+
+func NewConfigWriter(configWriterParameters *ConfigWriterParameters, instanceId string) *ConfigWriter {
```
olli-holmala, line 60:
```suggestion
func NewConfigWriter(configWriterParameters *ConfigWriterParameters, instanceID string) *ConfigWriter {
```

internal/datasource/config/config_writer.go
```diff
@@ -0,0 +1,195 @@
...
+	if configWriterParameters.cachePath == "" {
```
olli-holmala, line 61: I think it might be a good idea to add a `nil` check for `configWriterParameters`, as otherwise this is can be a `nil` pointer dereference.

api/grpc/instances/instances.proto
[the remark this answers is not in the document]
aphralG: ya it starts at 1 I believe

olli-holmala (approved):

dhurley (approved):

internal/datasource/config/config_writer.go
oliveromahony: should we default the values in the configWriterParameters so it is never nil and the you can remove the cachePath default value from here? Whichever one is better (this or whatever I suggested)

internal/service/config/nginx.go
```diff
@@ -191,3 +230,7 @@ func traverse(root *crossplane.Directive, callback crossplaneTraverseCallback, s
 	}
 	return nil
 }
+
+func getPermissions(fileMode os.FileMode) string {
+	return fmt.Sprintf("%#o", fileMode.Perm())
```
oliveromahony, line 235: not sure if this is the best place for this function

aphralG: Ya, I meant to ask about that. Donal had added it to `files.go` which I had deleted in my branch. I just put it there temporarily as it was only being used there. Forgot to ask where to put it, I can add files.go back in ?

aphralG: Added file_helpers.go which has GetPermissions()

PR #563, opened 2024-02-06
Add allowed dir list

aphralG:
### Proposed changes

Add allowed directories list to agent config
Write now checks if filepath is in allowed directories list 

### Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -78,6 +78,8 @@ func GetConfig() *Config {
 	for _, dir := range strings.Split(config.ConfigDir, ":") {
 		if dir != "" && filepath.IsAbs(dir) {
 			config.AllowedDirectories = append(config.AllowedDirectories, dir)
+		} else {
+			slog.Warn("Invalid directory: ", "dir", dir)
 		}
 	}
 	slog.Debug("Agent config", "config", config)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -76,7 +76,7 @@ func GetConfig() *Config {
 	}
 
 	for _, dir := range strings.Split(config.ConfigDir, ":") {
-		if dir != "" {
+		if dir != "" && filepath.IsAbs(dir) {
 			config.AllowedDirectories = append(config.AllowedDirectories, dir)
 		}
 	}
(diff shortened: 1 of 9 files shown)


internal/config/config.go
```diff
@@ -65,13 +66,20 @@ func RegisterConfigFile() error {
...
+		ProcessMonitor:     getProcessMonitor(),
+		DataplaneAPI:       getDataplaneAPI(),
+		Client:             getClient(),
+		ConfigDir:          getConfigDir(),
+		AllowedDirectories: []string{},
 	}
 
+	for _, dir := range strings.Split(config.ConfigDir, ":") {
+		if dir != "" {
+			config.AllowedDirectories = append(config.AllowedDirectories, dir)
```
dhurley, line 80: probably should validate that the directory path is an absolute path before adding it to the array. If its not it shouldn't be added and a warning message should be logged.

internal/datasource/config/config_writer.go
```diff
@@ -36,46 +37,40 @@ type (
...
 	}
 
 	// map of files with filepath as key
 	FileCache = map[string]*instances.File
 )
 
-func NewConfigWriter(configWriterParameters *ConfigWriterParameters, instanceID string) *ConfigWriter {
-	if configWriterParameters.cachePath == "" {
-		configWriterParameters.cachePath = fmt.Sprintf(cacheLocation, instanceID)
+func NewConfigWriter(configClient client.HTTPConfigClientInterface, cachePath string,
```
dhurley, line 53: Might be out of scope of this PR but we probably should rename `HTTPConfigClientInterface` to `ConfigClient` to make it more generic

dhurley, line 53: Do you need to pass the cachePath in here as a parameter? Could you create the cachePath in the function using the instanceID?

aphralG: I can make the change in this PR, I don't mind

aphralG: Ya I can, I wasn't sure what we were doing with the cache path and if it would always be at that location so left it as a parameter. I'll remove it

Dean-Coakley (approved):

oliveromahony (approved):

dhurley (approved):

PR #566, opened 2024-02-09
Refactor plugins to remove unnecessary paramaters

dhurley:
### Proposed changes

Refactor plugins to remove unnecessary paramaters

### Checklist


oliveromahony (approved):

internal/plugin/dataplane_server.go
```diff
@@ -29,32 +29,19 @@ type (
...
-	if dataplaneServerParameters.instanceService == nil {
-		dataplaneServerParameters.instanceService = service.NewInstanceService()
-	}
-
+func NewDataplaneServer(agentConfig *config.Config, logger *slog.Logger) *DataplaneServer {
 	return &DataplaneServer{
-		address:         fmt.Sprintf("%s:%d", dataplaneServerParameters.Host, dataplaneServerParameters.Port),
-		logger:          dataplaneServerParameters.Logger,
-		instanceService: dataplaneServerParameters.instanceService,
+		address: fmt.Sprintf("%s:%d", agentConfig.DataplaneAPI.Host, agentConfig.DataplaneAPI.Port),
```
olli-holmala, line 43: Should we use `net.JoinHostPort()` as it can deal with IPv6 hosts? https://pkg.go.dev/net#JoinHostPort
```suggestion
		address: net.JoinHostPort(agentConfig.DataplaneAPI.Host, strconv.FormatInt(agentConfig.DataplaneAPI.Port), 10),
```

internal/plugin/plugin_manager.go
```diff
@@ -15,22 +15,15 @@ import (
...
 
 	plugins = append(plugins, processMonitor, instanceMonitor, configPlugin)
 
 	if agentConfig.DataplaneAPI.Host != "" && agentConfig.DataplaneAPI.Port != 0 {
-		dataplaneServer := NewDataplaneServer(&DataplaneServerParameters{
-			Host:   agentConfig.DataplaneAPI.Host,
-			Port:   agentConfig.DataplaneAPI.Port,
-			Logger: slogger,
-		})
+		dataplaneServer := NewDataplaneServer(agentConfig, slogger)
```
olli-holmala, line 26: What's the difference between passing `slogger` around as a parameter and calling e.g. `slog.Info()` like we've done elsewhere?

internal/plugin/process_monitor.go
```diff
@@ -11,31 +11,25 @@ import (
...
-
+func NewProcessMonitor(agentConfig *config.Config) *ProcessMonitor {
 	return &ProcessMonitor{
-		params:    params,
-		processes: []*model.Process{},
+		monitoringFrequency: agentConfig.ProcessMonitor.MonitoringFrequency,
+		processes:           []*model.Process{},
+		getProcessesFunc:    host.GetProcesses,
 	}
 }
```
olli-holmala, line 34: Extremely off-topic, feel free to ignore:

Going back to the "Updating a config endpoint" scenario that we discussed at the office, would it not be better if the plugin just saved a pointer to `agentConfig` and always read from that? That way the underlying `config.Config` could be updated by our "Config update service" we have and we could just use the same config pointer here.

internal/plugin/process_monitor.go
```diff
@@ -11,31 +11,25 @@ import (
...
-}
-
 type ProcessMonitor struct {
-	params      *ProcessMonitorParameters
-	processes   []*model.Process
-	messagePipe bus.MessagePipeInterface
+	monitoringFrequency time.Duration
+	processes           []*model.Process
+	messagePipe         bus.MessagePipeInterface
+	getProcessesFunc    GetProcessesFunc
```
olli-holmala, line 25: I forget, what is the purpose of the `GetProcessesFunc`, why don't we just call `host.GetProcesses()` directly? Is the idea to pass a different `GetProcessesFunc` for K8s?

olli-holmala (approved):

internal/plugin/plugin_manager.go
dhurley: I only need the logger here for this plugin as I need it to configure the http server with the same logger created in main.go. See line 83.
`server.Use(sloggin.NewWithConfig(dps.logger, sloggin.Config{DefaultLevel: slog.LevelDebug}))`

internal/plugin/process_monitor.go
dhurley: The only purpose of the `GetProcessesFunc` is so that I can mock the call in the unit tests

internal/plugin/process_monitor.go
dhurley: We will still have to notify plugins and services of the config changes but lets discuss this again when we start looking at that work.

PR #568, opened 2024-02-12
Add new unit test coverage

oliveromahony:
### Proposed changes

Tidied up some testing and made the code more readable

### Checklist


test/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func CreateProtoTime(t *testing.T, timeString string) *timestamppb.Timestamp {
+	t.Helper()
+	newTime, err := time.Parse(time.RFC3339, timeString)
+	protoTime := timestamppb.New(newTime)
+	if err != nil {
+		assert.Fail(t, fmt.Sprintf("failed on creating timestamp %s", protoTime))
+	}
```
dhurley, line 41:
```suggestion
        require.NoError(t, err)
	protoTime := timestamppb.New(newTime)
```

test/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func CreateFileWithErrorCheck(t *testing.T, dir, fileName string) *os.File {
```
dhurley, line 19: The `CreateFileWithErrorCheck` and `RemoveFileWithErrorCheck` functions seem a bit of an overkill to me. Don't think we need to make our own functions for creating files and directories.

dhurley (commented):
I see that in some places you have renamed `dataplane` to `dataPlane`. I think you have missed a few other places in the codebase where the name would need to be changed also to be consistent.

test/helpers.go
oliveromahony: it simplifies 2 lines of code that appears numerous times in the tests

oliveromahony:
> I see that in some places you have renamed `dataplane` to `dataPlane`. I think you have missed a few other places in the codebase where the name would need to be changed also to be consistent.

I just did it in some places and not all to make the PR smaller. I tried to do it in the files I touched mostly

aphralG (approved):

test/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	testConf, err := os.CreateTemp(dir, fileName)
+	require.NoError(t, err)
+
+	return testConf
+}
+
+func RemoveFileWithErrorCheck(t *testing.T, fileName string) {
+	t.Helper()
+	err := os.Remove(fileName)
+	if err != nil {
```
dhurley, line 30: instead of the if statement would `require.NoError(err)` be better?

test/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func CreateProtoTime(t *testing.T, timeString string) *timestamppb.Timestamp {
+	t.Helper()
+	newTime, err := time.Parse(time.RFC3339, timeString)
+	require.NoError(t, err)
+
+	protoTime := timestamppb.New(newTime)
+	if err != nil {
```
dhurley, line 41: You can remove this if statement

test/helpers.go
dhurley: I guess this is used a far bit alright in the test code. Just want to be careful that we dont overuse this pattern and end up with a lot of wrapper test functions that only check errors returned from another function.

internal/plugin/dataplane_server_test.go
```diff
@@ -26,39 +26,68 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestDataplaneServer_Init(t *testing.T) {
-	dataplaneServer := NewDataplaneServer(&config.Config{}, slog.Default())
+func TestDataPlaneServer_Init(t *testing.T) {
+	dataPlaneServer := NewDataplaneServer(&config.Config{}, slog.Default())
```
dhurley, line 30: Should `NewDataplaneServer` be `NewDataPlaneServer` now?

internal/plugin/dataplane_server.go
```diff
@@ -88,10 +88,10 @@ func (dps *DataplaneServer) run(_ context.Context) {
 	server.Use(sloggin.NewWithConfig(dps.logger, sloggin.Config{DefaultLevel: slog.LevelDebug}))
```
dhurley, line 88: Should the file be called `data_plane_server.go` now?

test/data.go
```diff
@@ -0,0 +1,65 @@
...
+		FilePath:    path,
+		InstanceId:  instanceID,
+		FileContent: content,
+	}
+}
+
+func CreateProtoTime(timeString string) *timestamppb.Timestamp {
+	newTime, err := time.Parse(time.RFC3339, timeString)
+	if err != nil {
+		slog.Error("failed to parse time")
```
dhurley, line 56: Should we not return an error here?

test/helpers.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) F5, Inc.
```
dhurley, line 1: We are using `require.NoError`, `t.Fail` and `assert.Fail` in this file when checking errors. We should pick one way and be consistent.

oliveromahony: that was updated based on previous comments

olli-holmala (approved):

dhurley (approved):

PR #569, opened 2024-02-13
Add integration tests for get instances endpoint

dhurley:
### Proposed changes

Add integration tests for get instances endpoint

### Checklist


scripts/docker/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+nginx_pid=$!
+
+SECONDS=0
+
+while ! ps -ef | grep "nginx: master process" | grep -v grep; do 
+    if (( SECONDS > 5 )); then 
+        echo "couldn't find nginx master process"
+        exit 1
+    fi
+done
```
olli-holmala, line 29: Should `SECONDS` be incremented?

dhurley: No the `SECONDS` variable is a special built in variable in bash that increments itself as time passes.

From [bash man page](https://man7.org/linux/man-pages/man1/bash.1.html):

SECONDS

Each time this parameter is referenced, the number of seconds since shell invocation is returned. If a value is assigned to SECONDS, the value returned upon subsequent references is the number of seconds since the assignment plus the value assigned. If SECONDS is unset, it loses its special properties, even if it is subsequently reset.

.github/workflows/ci.yml
```diff
@@ -58,9 +58,34 @@ jobs:
...
+      - name: Download Packages
+        uses: actions/download-artifact@eaceaf801fd36c7dee90939fad912460b18a1ffe # v4.1.2
+        with:
+          name: nginx-agent-unsigned-snapshots
+          path: build
+      - name: Run Integration Tests
+        run: |
+          go install github.com/goreleaser/nfpm/v2/cmd/nfpm@${{ env.NFPM_VERSION }}
+          OS_RELEASE="${{ matrix.container.image }}" OS_VERSION="${{ matrix.container.version }}" \
+            make integration-test
```
aphralG, line 91: Would it be worth capturing the logs if they fail like we have done previously https://github.com/nginx/agent/blob/82c7a23f0a0a70e8f91e2bf3a114aee155082c8f/.github/workflows/ci.yml#L160

dhurley: Not sure if we still need to do this anymore. In the test teardown function I log the container logs now every time.
https://github.com/nginx/agent/blob/add-integration-tests/test/integration/dataplane_api_test.go#L63

test/integration/utils/test_containers_utils.go
```diff
@@ -0,0 +1,71 @@
...
+			Context:       "../../",
+			Dockerfile:    fmt.Sprintf("./scripts/docker/nginx-oss/%s/Dockerfile", os.Getenv("CONTAINER_OS_TYPE")),
+			KeepImage:     false,
+			PrintBuildLog: true,
+			BuildArgs: map[string]*string{
+				"PACKAGE_NAME":  toPtr(os.Getenv("PACKAGE_NAME")),
+				"PACKAGES_REPO": toPtr(os.Getenv("PACKAGES_REPO")),
+				"BASE_IMAGE":    toPtr(os.Getenv("BASE_IMAGE")),
+				"OS_RELEASE":    toPtr(os.Getenv("OS_RELEASE")),
+				"OS_VERSION":    toPtr(os.Getenv("OS_VERSION")),
```
olli-holmala, line 37: Should we assert that the env vars aren't empty before passing them to the testcontainer?

olli-holmala (approved):

.github/workflows/ci.yml
aphralG: Aww right perfect, thought that was only the agent logs and not all of the logs misread it.

aphralG (approved):

test/integration/utils/test_containers_utils.go
dhurley: yeah thats a good point. Let me make that update.

test/integration/dataplane_api_test.go
craigell, line 1: Will there be a config apply integration test?

dhurley: Yeah there will be a separate PR for that. I just wanted to keep this PR small so I only added one simple test

PR #573, opened 2024-02-20
Add mock management plane API

dhurley:
### Proposed changes

- Add mock management plane API that can be used for integration testing.
- Centralised NGINX test config files.

### Checklist


oliveromahony (approved):

test/mock/mock-management-plane-api.yaml
```diff
@@ -0,0 +1,170 @@
...
+          required: true
+          description: Instance ID
+          example: "aecea348-62c1-4e3d-b848-6d6cdeb1cb9c"
+        - in: path
+          name: filePath
+          schema:
+            type: string
+          required: true
+          description: file location
+          example: "/etc/nginx/nginx.conf"
```
olli-holmala, line 57: I believe I brought this up before but what happens if we pass an absolute path to `filePath`? Since we are passing it in the URL, does the leading slash get omitted?

test/mock/mock-management-plane-api.yaml
```diff
@@ -0,0 +1,170 @@
...
+      required:
+        - type
+      properties:
+        instanceId:
+          type: string
+          example: "aecea348-62c1-4e3d-b848-6d6cdeb1cb9c"
+        files:
+          type: array
+          items:
+            $ref: '#/components/schemas/File'
```
olli-holmala, line 102: We've listed the `type` property as required but we haven't defined a `type` property.

test/mock/mock-management-plane-api.yaml
```diff
@@ -0,0 +1,170 @@
...
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/DownloadResponse'
+                  - $ref: '#/components/schemas/DownloadResponseEncoded'
+                discriminator:
+                  propertyName: type
+
+components:
```
olli-holmala, line 83: Just pointing out that we haven't marked many of our components' fields as `required`, meaning that the generated structs' fields will be pointers. I think it's good practice to mark fields that we expect to be populated as `required` to reduce the need for nil checks in our code.

test/mock/mock-management-plane-api.yaml
dhurley: no the leading slash is still ther. Here is an example with the file path encoded in the url:
`api/v1/instances/32/files/%2Fetc%2Fnginx%2Fnginx.conf?encoded=true`

olli-holmala: Should we add note in the description to URL encode the `filePath`?

dhurley: sure let me add it there

test/mock/mock_management_plane_server.go
```diff
@@ -0,0 +1,159 @@
...
+				InstanceId:  instanceID,
+				Type:        DOWNLOADRESPONSE,
+			}
+
+			ctx.JSON(http.StatusOK, response)
+
+			return
+		} else {
+			ctx.JSON(http.StatusNotFound, "")
+		}
```
olli-holmala, line 127: Just checking: so we are expecting the first file in `files` to always be the one we are looking for, otherwise we return a Not Found response?

olli-holmala (approved):

test/mock/mock_management_plane_server.go
dhurley: oh good spot that was meant to be outside the for loop. Let me update that

PR #574, opened 2024-02-22
Draft: Management API Definitions for Review

oliveromahony:
### Proposed changes

This PR represents the proto definitions for version 3 of the NGINX Agent. The Management Plane Interface (MPI) is a specification defining the functionality exposed by a Management Server in the form of gRPC services and HTTPS endpoints. The NGINX Agent connects to a Management Server in order to administer an NGINX Instance (OSS, Plus).

### Checklist


nginx-nickc:
Some general notes:
- keep proto separate from generated bindings
- using full path , i.e.  f5.nginx.agent.api.grpc.mpi.v1.common.MessageRequest for imports? I think can be shorten to common.MessageRequest

api/grpc/mpi/v1/common.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+package f5.nginx.agent.api.grpc.mpi.v1.common;
+
+option go_package = "github.com/nginx/agent/v3/api/grpc/mpi/v1";
+
+// Meta-information associated with a message
+message MessageRequest {
```
nginx-nickc, line 8: nit, naming. Shouldn't this just be MessageMeta?

api/grpc/mpi/v1/file.proto
```diff
@@ -0,0 +1,85 @@
...
+
+import "common.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/nginx/agent/v3/api/grpc/mpi/v1";
+
+service FileService {
+    // Get the overview of files for a particular configuration version of an instance
+    rpc GetOverview(ConfigVersion) returns (FileOverview) {}
```
nginx-nickc, line 12: nit, i think would be nice to document the direction. While conventionally, we know the "Get" is from perspective of the Agent, it'll be good to be explicit.

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    // Reports on instance health
+    rpc Health(DataPlaneHealth) returns (google.protobuf.Empty) {}
+    // A decoupled communication mechanism between the data plane and management plane.
+    rpc Subscribe(stream DataPlaneMessage) returns (stream ManagementPlaneMessage) {}
+}
+
+// The connection request is an intial handshake to establish a connection, sending NGINX Agent instance information
+message ConnectionRequest {
+    // Meta-information associated with a message
+    f5.nginx.agent.api.grpc.mpi.v1.common.MessageRequest message_metadata = 1;
```
nginx-nickc, line 29: nit, `message_metadata`, we have "meta" in other places. Since metadata and meta are kind of the same thing, please pick one to be consistent.

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+}
+
+// This contains a series of NGINX Agent configurations
+message AgentConfig { 
+    // Command server settings
+    Command command = 1;
+    // Metrics server settings
+    Metrics metrics = 2;
+    // A series of key/value pairs to add more data to the NGINX Agent instance
+    repeated google.protobuf.Struct labels = 3;
```
nginx-nickc, line 41: `google.protobuf.Struct` is just `map<string, [Value]>` as per https://protobuf.dev/reference/protobuf/google.protobuf/#struct, where `Value` can be different things. https://protobuf.dev/reference/protobuf/google.protobuf/#value. Should we just go with `map<string, string>` ?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    // A list of features that the NGINX Agent has
+    repeated string features = 4;
```
nginx-nickc, line 43: should this be an enum, free form string makes it hard to synchronize on common set of "features" that's understood by the agent

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    // Message buffer size, maximum not acknowledged messages from the subscribe perspective
+    string message_buffer_size = 5;
+    // 
+}
+
+// The command settings, associated with messaging from an external source 
+message Command {
```
nginx-nickc, line 50: this feels like a one off, not used else where. Consider embed inside `AgentConfig`

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    // The connection and security settingss for the command server
+    ConnectionSettings connection_settings = 1;
+}
+
+// The metrics settings associated with orgins (sources) of the metrics and destinations (exporter)
+message Metrics {
```
nginx-nickc, line 56: Metrics feels like a one off, not used outside of AgentConfig, consider embed.

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    double multiplier = 3;
+    // Max interval in seconds between two retries
+    int64 max_interval = 4;
+    // Elapsed time in seconds after which backoff stops.
+    // It never stops if max_elapsed_time == 0.
+    int64 max_elapsed_time = 5;
+}
+
+// A set of runtime NGINX configuration that gets populated 
+message NGINXConfig {
```
nginx-nickc, line 149: I think we want installed modules as well

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+
+// A set of runtime NGINX configuration that gets populated 
+message NGINXPlusConfig {
+    // where the binary location is, if empty, this is a remote instance
+    string binary_path = 1;
+    // the hostname associated with NGINX Plus
+    string hostname = 2;
+    // the ip address associated with NGINX Plus
+    string ip_address = 3;
+    // the API inforation for NGINX Plus API
```
nginx-nickc, line 172: spelling, "information"

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    repeated string error_logs = 6;
+}
+
+// A set of runtime NGINX configuration that gets populated 
+message NGINXPlusConfig {
+    // where the binary location is, if empty, this is a remote instance
+    string binary_path = 1;
+    // the hostname associated with NGINX Plus
+    string hostname = 2;
+    // the ip address associated with NGINX Plus
```
nginx-nickc, line 170: nit, not sure what "associated" means, clarify?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,358 @@
...
+    // tls key
+    string key       = 3;
+    // certificate authoirty cert
+    string ca        = 4;
+    // enable verification of a server's certificate chain and host name
+    bool skip_verify = 5;
+}
+
+// backoff settings
+message Backoff {
```
nginx-nickc, line 134: this feels like Backoff setting for certain implementation of back-off algorithm/library, recommend simplify to "interval" and "max_elapsed"

oliveromahony:
> Some general notes:
> 
> * keep proto separate from generated bindings
Where is this recommendation from?

> * using full path , i.e.  f5.nginx.agent.api.grpc.mpi.v1.common.MessageRequest for imports? I think can be shorten to common.MessageRequest
We don't have a naming convention for the package. I prefer the longer one since it shows where the source originated from

api/grpc/mpi/v1/file.proto
```diff
@@ -0,0 +1,85 @@
...
+    // the instance identifier
+    string instance_id = 1;
+    // the version of the configuration
+    string version = 2;
+}
+
+// Represents a collection of files
+message FileOverview {
+    // A list of files
+    repeated File file = 1;
```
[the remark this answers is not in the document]
oliveromahony: This was specified in the definition

api/grpc/mpi/v1/common.proto
```diff
@@ -0,0 +1,35 @@
...
+    // monotonically increasing integer
+    string message_id = 1;
+    // if 2 or more messages associated with the same workflow, use this field as an association
+    string correlation_id = 2;
+}
```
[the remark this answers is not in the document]
oliveromahony: the message_id is a monotonically increasing integer. Is this sufficient?

api/grpc/mpi/v1/command.proto
oliveromahony: this payload is likely to change during implementation. I added it for reference but don't understand it's purpose. It was mentioned in the spec so I put it here for illustration. I can remove it and we can build the proper payload as we go

api/grpc/mpi/v1/file.proto
```diff
@@ -0,0 +1,85 @@
...
+    // byte representation of a file without encoding
+    bytes contents = 1;
+}
+
+// Meta information about the file, the name (including path) and hash
+message FileMeta {
+    // the name of the file
+    string name = 1;
+    // the hash of the file contents
+    string hash = 2;
```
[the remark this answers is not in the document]
oliveromahony: there is an outstanding comment on how this should be calculated on the definition

api/grpc/mpi/v1/common.proto
```diff
@@ -0,0 +1,35 @@
...
+
+// Represents a the status response of an command
+message CommandResponse {
+    // Command status enum
+    enum CommandStatus {
+        // Unknown status of command
+        CMD_UNKNOWN = 0;
```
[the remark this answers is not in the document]
oliveromahony: https://protobuf.dev/programming-guides/style/#enums

api/grpc/mpi/v1/file.proto
oliveromahony: followed the reference here: https://protobuf.dev/programming-guides/style/#services

api/grpc/mpi/v1/command.proto
oliveromahony: The values can be lists so I think this type is best

api/grpc/mpi/v1/command.proto
oliveromahony: potentially, this can evolve in the implementation

api/grpc/mpi/v1/command.proto
oliveromahony: this will diverge when we add more fields so I followed this recommendation https://protobuf.dev/programming-guides/api/#unique-protos

api/grpc/mpi/v1/command.proto
oliveromahony: changed to keepalive

api/grpc/mpi/v1/common.proto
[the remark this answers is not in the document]
oliveromahony: I updated all enums to follow the style guide and changed the defaults to PREFIX_UNSPECIFIED

api/grpc/mpi/v1/file.proto
nginx-nickc: not sure how your link apply here?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,379 @@
...
+// Messages provided by the Management Plane must be a FIFO ordered queue. Messages in the queue must have a monotonically-increasing integer index. 
+// The indexes do not need to be sequential. The index must be a 64-bit signed integer.
+// The index must not reset for the entire lifetime of a unique Agent (i.e. the index does not reset to 0 only because of a temporary disconnection or new session). 
+// Messages must not be removed from the Mangement Plane queue until Ack’d by the Agent. 
+// Messages sent but not yet Ack’d must be kept in an “in-flight” buffer as they may need to be retried.
+service CommandService {
+    // Connects NGINX Agent to the Management Plane agnostic of instance data
+    rpc CreateConnection(ConnectionRequest) returns (ConnectionResponse) {}
+    // Reports on instances and their configurations
+    rpc UpdateStatus(DataPlaneStatus) returns (google.protobuf.Empty) {}
```
nginx-nickc, line 32: https://protobuf.dev/programming-guides/api/#unique-protos recommends define Empty message

api/grpc/mpi/v1/command.proto
nginx-nickc: probably require discussion

api/grpc/mpi/v1/common.proto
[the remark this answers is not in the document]
nginx-nickc: timestamp can be use to calculate how long since the message was sent, etc

api/grpc/mpi/v1/command.proto
nginx-nickc: not sure how your link apply here

api/grpc/mpi/v1/common.proto
```diff
@@ -0,0 +1,40 @@
...
+    // if 2 or more messages associated with the same workflow, use this field as an association
+    string correlation_id = 2;
+}
+
+// Represents a the status response of an command
+message CommandResponse {
+    // Command status enum
+    enum CommandStatus {
+        // Unspecified status of command
+        COMMAND_STATUS_UNSPECIFIED = 0;
```
kate-osborn, line 25: I'm curious why all enums have an unspecified value. Is that a common/best practice? Is it just to create a zero value?

api/grpc/mpi/v1/file.proto
```diff
@@ -0,0 +1,97 @@
...
+// A SHA256 hash string is 64 bytes, therefore the configured max message size should be maxFileSize + 64.
+service FileService {
+    // Get the overview of files for a particular configuration version of an instance
+    rpc GetOverview(ConfigVersion) returns (FileOverview) {}
+    // Update the overview of files for a particular set of file changes on the data plane
+    rpc UpdateOverview(FileOverview) returns (google.protobuf.Empty) {}
+    // Get the file contents for a particular file
+    rpc GetFile(FileRequest) returns (FileContents) {}
+    // Update a file from the Agent to the Server
+    rpc UpdateFile(File) returns (FileMeta) {}
```
kate-osborn, line 30: Should this be `SendFile`?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,379 @@
...
+
+// Additional information associated with a ConfigApplyRequest
+message ConfigApplyRequest {
+    // the config version
+    f5.nginx.agent.api.grpc.mpi.v1.file.ConfigVersion config_version = 1;
+    // an optional set of files related to the request
+    f5.nginx.agent.api.grpc.mpi.v1.file.FileOverview overview = 2; // optional
+}
+
+message ConfigSyncRequest {
```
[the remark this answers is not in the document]
oliveromahony: It's a request from the management server for the Agent to perform an upload. The payload will evolve in the implementation. No config apply is done as part of this

api/grpc/mpi/v1/command.proto
oliveromahony: made consistent. Resolving

api/grpc/mpi/v1/command.proto
oliveromahony: done

api/grpc/mpi/v1/command.proto
oliveromahony: I'd rather keep the structure the way it is. If it is desired not to consume the full flexibility of the structures offered, that's fine, but other consumers of this API might want to

api/grpc/mpi/v1/command.proto
oliveromahony: Metrics and Command are wrappers for evolving proto specs and will diverge in the implementation. Burying them in the top level doesn't give us the flexibility to evolve

api/grpc/mpi/v1/common.proto
oliveromahony: yes https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum

api/grpc/mpi/v1/file.proto
oliveromahony: it can be used for both Add and Updating a file from the server perspective. We don't have a naming convention for this

mtbChef:
Wanted to check in on what was decided in regards to the message index. I had proposed that we use millis since epoch or timestamp vs (or in addition to?) the index. Doing so would bake in the ability to troubleshoot messages by time and, if a message would eventually be consumed by a user, would provide a timestamp for audit purposes.

nginx-nickc:
> Wanted to check in on what was decided in regards to the message index. I had proposed that we use millis since epoch or timestamp vs (or in addition to?) the index. Doing so would bake in the ability to troubleshoot messages by time and, if a message would eventually be consumed by a user, would provide a timestamp for audit purposes.

don't think we should mix purposes here. Since we are intending to be an open standard, the assumption would be different for message_index vs message timestamp for different implementation. The current requirement for the message index is that it SHOULD be monotonically increasing, where as the message timestamp should be the timestamp with timezone of the sending system. Even if you want to send the timestamp as epoch, should still attach timezone, since at some point we would want to correlate the event with logs from the sending system, which would most likely be logged in local timezone. Mixing the two concerns just makes it so much harder to troubleshoot down the road.

api/grpc/mpi/v1/file.proto
kate-osborn: Ok, I saw `SendFile` in the comments, so I was curious which name we are using.

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,387 @@
...
+    // Command server settings
+    Command command = 1;
+    // Metrics server settings
+    Metrics metrics = 2;
+    // A series of key/value pairs to add more data to the NGINX Agent instance
+    repeated google.protobuf.Struct labels = 3;
+    // A list of features that the NGINX Agent has
+    repeated string features = 4;
+    // Message buffer size, maximum not acknowledged messages from the subscribe perspective
+    string message_buffer_size = 5;
```
cohix, line 59: shouldn't this be an int?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,387 @@
...
+        StatusRequest status_request = 2;              
+        // triggers a DataPlaneHealth rpc
+        HealthRequest health_request = 3;               
+        // triggers a rpc GetFile(FileRequest) for overview list, 
+        // if overview is missing, triggers a rpc Overview(ConfigVersion) first
+        ConfigApplyRequest config_apply_request = 5;
+        // triggers a series of rpc SendFile(File) for that instances
+        ConfigUploadRequest config_upload_request = 6;
+        // triggers a reconciliation of  with a command_response for a particular action
+        ConfigSyncRequest config_sync_request = 7;
```
cohix, line 330: Why do we need sync if we have apply and upload?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,387 @@
...
+        HealthRequest health_request = 3;               
+        // triggers a rpc GetFile(FileRequest) for overview list, 
+        // if overview is missing, triggers a rpc Overview(ConfigVersion) first
+        ConfigApplyRequest config_apply_request = 5;
+        // triggers a series of rpc SendFile(File) for that instances
+        ConfigUploadRequest config_upload_request = 6;
+        // triggers a reconciliation of  with a command_response for a particular action
+        ConfigSyncRequest config_sync_request = 7; 
+        // triggers a DataPlaneResponse with a command_response for a particular action
+        ActionRequest action_request = 8;
```
cohix, line 332: Can we make this explicit and name it `APIActionRequest` ?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,387 @@
...
+
+message ConfigSyncRequest {
+    // the config version
+    mpi.v1.ConfigVersion config_version = 1;
+    // an optional set of files related to the request
+    optional mpi.v1.FileOverview overview = 2;
+}
+
+// Additional information associated with a ConfigUploadRequest
+message ConfigUploadRequest {
```
cohix, line 360: Need to add a FileOverview here to indicate which files the MP wants uploaded

api/grpc/mpi/v1/common.proto
```diff
@@ -0,0 +1,38 @@
...
+// Meta-information associated with a message
+message MessageMeta {
+    // monotonically increasing integer
+    string message_id = 1;
+    // if 2 or more messages associated with the same workflow, use this field as an association
+    string correlation_id = 2;
+}
+
+// Represents a the status response of an command
+message CommandResponse {
```
cohix, line 19: Should this be MessageResponse?

cohix (commented):
Overall looking good. Just a few things from me.

api/grpc/mpi/v1/command.proto
oliveromahony: if the Agent has been offline for awhile or has local changes not shared with the Management Plane a sync is required no?

api/grpc/mpi/v1/command.proto
oliveromahony: are the actions specific to APIs only?

api/grpc/mpi/v1/file.proto
oliveromahony: I've updated the comment, thanks

api/grpc/mpi/v1/command.proto
cohix: The workflow currently specifies that the Agent should upload a FileOverview after being offline, and then MP can send a ConfigUploadRequest if there's anything missing. I'd say we don't need Sync.

api/grpc/mpi/v1/command.proto
cohix: Yes

api/grpc/mpi/v1/command.proto
oliveromahony: added

api/grpc/mpi/v1/common.proto
oliveromahony: the CommandService uses this?

api/grpc/mpi/v1/common.proto
```diff
@@ -0,0 +1,38 @@
...
+// LICENSE file in the root directory of this source tree.
+syntax = "proto3";
+package mpi.v1;
+
+option go_package = "github.com/nginx/agent/v3/api/grpc/mpi/v1";
+
+// Meta-information associated with a message
+message MessageMeta {
+    // monotonically increasing integer
+    string message_id = 1;
```
cohix, line 13: This should be an integer

api/grpc/mpi/v1/file.proto
```diff
@@ -0,0 +1,135 @@
...
+    // the configuration version of the current set of files
+    ConfigVersion version = 2;
+}
+
+// Represents meta data about a file
+message File {
+    // Meta information about the file, the name (including path) and hash
+    FileMeta file_meta = 1;
+    // last modified time of the file (created time if never modified)
+    google.protobuf.Timestamp modified_time = 2;
```
cohix, line 73: Maybe move modified time, permissions, size into FileMeta?

api/grpc/mpi/v1/file.proto
```diff
@@ -0,0 +1,133 @@
...
+// Represents meta data about a file
+message File {
+    // Meta information about the file, the name (including path) and hash
+    FileMeta file_meta = 1;
+    // Action enumeration
+    enum FileAction {
+        // Default value, no action
+        FILE_ACTION_UNSPECIFIED = 0;
+        // No changes to the file
+        FILE_ACTION_UNCHANGED = 1;
```
nginx-nickc, line 77: can probably remove UNCHANGED, and use UNSPECIFIED as the default, indicating no change.

oliveromahony: UNSPECIFIED is for an unspecified file FileAction. UNCHANGED is an explicit no file has changed. They have different purposes

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,387 @@
...
+    // the instance identifier
+    string instance_id = 1;
+    // an optional set of files related to the request
+    optional mpi.v1.FileOverview overview = 2;
+}
+
+// Perform an associated API action on an instance
+message APIActionRequest {
```
ryepup, line 366: thinking about NGINXaaS' azure-specific features, I think this "API" feature might be one option; NGINXaaS could run a local web server on the dataplane that the agent hits via these `APIActionRequest`s to do our weird azure stuff without needing to pollute various `enum`s here.

Are there intentions to expand this to allow specifying URLs and request bodies for the API action, and for the API response to make it back up to the management plane as part of a `CommandResponse`?

Or is the scope of `APIActionRequest` intended to be limited to an allowlist of N+ / Unit requests?

api/grpc/mpi/v1/command.proto
ryepup: an enum makes it harder to extend, right? If I want to add an agent feature that only makes sense for NGINXaaS, the flexibility of `string` is useful.

nginx-nickc: why is it harder to extend? You'll need to add the string value somewhere anyways, since there would need to be agreement on what a particular value "means". So you don't end up with one side thinking it's "fooooo" for one thing, and another side thinking it's "fooo", contrive example, but you get the point.

ryepup: I understand and agree with enums to avoid magic string shenanigans, my concern is about extensibility. 
Maybe I'm missing something about gRPC protos, I don't understand how I can extend the enum in my custom NGINXaaS agent. Say we do use and enum here:

```proto
enum AgentFeature { CONFIG=0, METRICS=1 }
```

Then in my NGINXaaS agent, I have a azure-specific feature that is _not_ part of the shared protocol, say `AZURE_MONITOR`. I don't want to pollute all agents everywhere with my Azure-specific feature, and want freedom to iterate the NGINXaaS agent/server independently of other NGINX products. My agent is in golang, using this protocol directly via `import "github.com/nginx/agent/v3/api/grpc/mpi/v1"`.

Is there a way to add to `enum AgentFeature` after the fact for NGINXaaS?

api/grpc/mpi/v1/command.proto
```diff
@@ -0,0 +1,387 @@
...
+    // instance information associated with the NGINX Agent
+    Instance agent = 2;
+}
+
+// This contains a series of NGINX Agent configurations
+message AgentConfig { 
+    // Command server settings
+    Command command = 1;
+    // Metrics server settings
+    Metrics metrics = 2;
```
cohix, line 53: Missing a `File` settings field for configuring the file upload/download endpoint. `self` is a valid value indicating the use of the current gRPC connection, `grpc://` or `https://` endpoints are also valid.

api/grpc/mpi/v1/command.proto
```diff
@@ -41,8 +41,55 @@ service CommandService {
 message CreateConnectionRequest {
     // Meta-information associated with a message
     mpi.v1.MessageMeta message_meta = 1;
-    // instance information associated with the NGINX Agent
-    Instance agent = 2;
+    // instance and infrastructure information associated with the NGINX Agent
+    Resource resource = 2;
```
cohix, line 45: Even though the type is `Resource`, maybe we still name the field `agent` ?

oliveromahony: it represents where the Agent is deployed and the associated information around that. Happy to adjust the name but just want to say that it includes Agent, any NGINX product(s), host or container information

PR #578, opened 2024-02-28
Update NGINX config apply function

dhurley:
### Proposed changes

Update NGINX config apply function to use `syscall.Kill(pid, syscall.SIGHUP)` instead of `nginx -s reload`

### Checklist


Dean-Coakley (approved):

internal/service/config/nginx_config_service.go
```diff
@@ -147,12 +149,21 @@ func (n *Nginx) Validate() error {
...
+
+	processID := n.instance.GetMeta().GetNginxMeta().GetProcessId()
+
+	intProcessID, err := strconv.Atoi(processID)
+	if err != nil {
+		slog.Error("Invalid NGINX master process ID", "process_id", processID, "error", err)
+		return err
+	}
+
+	err = n.executor.KillProcess(intProcessID, syscall.SIGHUP)
```
oliveromahony, line 161: will we ever do a kill process with a different syscall? I think it's fine to encapsulate it in the KillProcess function for now and we can extend it later

dhurley: yeah probably not. Let me make that update there now

api/grpc/instances/instances.proto
```diff
@@ -23,6 +23,7 @@ message Meta {
 message NginxMeta {
     string config_path = 1;
     string exe_path = 2;
+    string process_id = 3;
```
olli-holmala, line 26: Why not use `int` directly instead of `string`?

olli-holmala (approved):

aphralG (approved):

PR #582, opened 2024-02-29
Add rollback & change configuration status

aphralG:
### Proposed changes

Add: 
- Rollback config if Write, Validate, Reload or Complete fails
- `GET /instances/{instanceId}/configurations/status` now returns a list of all the events for the most recent config apply (example response below)
- `InstanceConfigUpdateCompleteTopic` changed to `InstanceConfigUpdateTopic`

GET /instances/{instanceId}/configurations/status Example: 

``` json
{
    "correlationId": "",
    "instanceID": "",
    "status": [
        {
            "message": "Instance configuration update in progress",
            "status": "IN_PROGRESS",
            "timeStamp": "2024-02-28T16:40:04.823341Z"
        },
        {
            "message": "NGINX config test failed exit status 1: nginx: [emerg] directive \"worker_processes\" is not terminated by \";\" in /etc/nginx/nginx.conf:7\nnginx: configuration file /etc/nginx/nginx.conf test failed\n",
            "status": "FAILED",
            "timeStamp": "2024-02-28T16:40:08.847568Z"
        },
        {
            "message": "Rollback in progress",
            "status": "ROLLBACK_IN_PROGRESS",
            "timeStamp": "2024-02-28T16:40:08.847568Z"
        },
        {
            "message": "Rollback successful",
            "status": "ROLLBACK_SUCCESS",
            "timeStamp": "2024-02-28T16:40:08.847578Z"
        }
    ]
}
```

### Checklist


api/http/dataplane/dataplane-api.yaml
```diff
@@ -145,26 +145,38 @@ components:
     ConfigurationStatus:
       type: object
       properties:
+        instanceID:
```
dhurley, line 148: Think we are using `Id` in the REST API field names instead of `ID`

internal/datasource/config/config_writer.go
```diff
@@ -60,15 +61,32 @@ func (cw *ConfigWriter) SetConfigClient(configClient client.ConfigClientInterfac
...
+	}
+
+	return nil
+}
+
 func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
 	tenantID, instanceID string,
-) (skippedFiles map[string]struct{}, err error) {
+) (skippedFiles CacheContent, err error) {
+	slog.Info("Writing file")
```
dhurley, line 85: Should this be Debug?

dhurley: Also maybe add filesURL to the message so that we know what file is being written

internal/datasource/config/config_writer.go
```diff
@@ -60,15 +61,32 @@ func (cw *ConfigWriter) SetConfigClient(configClient client.ConfigClientInterfac
...
 	currentFileCache := make(CacheContent)
-	skippedFiles = make(map[string]struct{})
-	cacheContent, err := cw.fileCache.ReadFileCache()
-	if err != nil {
-		slog.Warn("Failed to read file cache")
-	}
+	skippedFiles = CacheContent{}
+
+	cacheContent, _ := cw.fileCache.ReadFileCache()
```
dhurley, line 89: Need to handle error here

internal/plugin/data_plane_server.go
```diff
@@ -68,17 +67,27 @@ func (dps *DataPlaneServer) Process(msg *bus.Message) {
...
-			dps.configurationStatues[configStatus.GetInstanceId()] = configStatus
+			dps.updateEvents(configStatus)
 		}
 	}
 }
 
+func (dps *DataPlaneServer) updateEvents(configStatus *instances.ConfigurationStatus) {
+	instanceID := configStatus.GetInstanceId()
+	if configStatus.GetStatus() == instances.Status_IN_PROGRESS {
+		dps.configEvents[instanceID] = nil
```
dhurley, line 80:
```suggestion
		dps.configEvents[instanceID] = make(map[string][]*instances.ConfigurationStatus)
```

internal/service/config_service.go
```diff
@@ -9,6 +9,10 @@ import (
 	"context"
 	"log/slog"
 
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	config2 "github.com/nginx/agent/v3/internal/datasource/config"
```
dhurley, line 14:
```suggestion
	datasourceConfig "github.com/nginx/agent/v3/internal/datasource/config"
```

internal/service/config_service.go
```diff
@@ -59,56 +64,69 @@ func (cs *ConfigService) SetConfigContext(instanceConfigContext any) {
...
+) (skippedFiles config2.CacheContent, configStatus *instances.ConfigurationStatus) {
 	// remove when tenantID is being set
 	tenantID := "7332d596-d2e6-4d1e-9e75-70f91ef9bd0e"
 
-	_, err := cs.configService.Write(ctx, location, tenantID)
+	skippedFiles, err := cs.configService.Write(ctx, location, tenantID)
 	if err != nil {
-		// Rollback
-		return &instances.ConfigurationStatus{
+		slog.Debug("Error writing config", "err", err)
```
dhurley, line 81: Should these be Error instead of Debug?

api/http/dataplane/dataplane-api.yaml
aphralG: Will fix, but is there a reason we are using InstanceId here but InstanceID everywhere else ?

internal/service/config_service.go
aphralG: Renamed to datasource as linter doesn't allow datasourceConfig or datasource_config. We should maybe do soemthing with all the package names being config

api/http/dataplane/data-plane-api.yaml
```diff
@@ -145,26 +145,38 @@ components:
...
+          example: "aecea348-62c1-4e3d-b848-6d6cdeb1cb9c"
         correlationId:
           type: string
           example: "9e2d49c9-ada2-4ed1-b6f2-391d5cf634d9"
-        lastUpdated:
+        events:
+          type: array
+          items:
+            $ref: "#/components/schemas/Events"
+    Events:
```
olli-holmala, line 158: This should be in the singular form, right?
```suggestion
    Event:
```

api/http/dataplane/data-plane-api.yaml
```diff
@@ -145,26 +145,38 @@ components:
...
           type: string
           example: "9e2d49c9-ada2-4ed1-b6f2-391d5cf634d9"
-        lastUpdated:
+        events:
+          type: array
+          items:
+            $ref: "#/components/schemas/Event"
+    Event:
+      type: object
+      properties:
```
olli-holmala, line 160: Should we make some of these fields required (so they aren't pointers in our generated code)?
```suggestion
    Event:
      type: object
      required:
        - timestamp
        - status
        - message
      properties:
```

api/http/dataplane/data-plane-api.yaml
```diff
@@ -145,26 +145,38 @@ components:
     ConfigurationStatus:
       type: object
       properties:
```
olli-holmala, line 147: Should we make some of these fields required (so they aren't pointers in our generated code)?
```suggestion
    ConfigurationStatus:
      type: object
      required:
        - instanceId
        - correlationId
        - events
      properties:
```

api/http/dataplane/data-plane-api.yaml
aphralG: Ya probably should make some things required. Not much in api is required at the moment so should we also discuss what should be required in the rest of the api ?

internal/datasource/config/config_writer_test.go
```diff
@@ -27,99 +27,188 @@ import (
...
+				err = fileCache.UpdateFileCache(cacheContent)
+				require.NoError(t, err)
 			}
 
 			configWriter, err := NewConfigWriter(&agentconfig, fileCache)
 			require.NoError(t, err)
 
 			configWriter.SetConfigClient(fakeConfigClient)
-			err = writeFile(fileContent, testConfPath)
+			testConent := []byte("location /test {\n    return 200 \"Before Write \\n\";\n}")
```
olli-holmala, line 105:
```suggestion
			testContent := []byte("location /test {\n    return 200 \"Before Write \\n\";\n}")
```

internal/plugin/config_plugin.go
```diff
@@ -120,16 +122,61 @@ func (c *Config) parseInstanceConfiguration(correlationID string, instance *inst
 }
 
 func (c *Config) updateInstanceConfig(request *model.InstanceConfigUpdateRequest) {
+	// remove when tenantID is being set
+	tenantID := "7332d596-d2e6-4d1e-9e75-70f91ef9bd0e"
```
olli-holmala, line 126: nit: move to `const`

internal/plugin/config_plugin.go
```diff
@@ -120,16 +122,61 @@ func (c *Config) parseInstanceConfiguration(correlationID string, instance *inst
...
+		rollbackInProgress := &instances.ConfigurationStatus{
+			InstanceId:    instanceID,
+			CorrelationId: request.CorrelationID,
+			Status:        instances.Status_ROLLBACK_IN_PROGRESS,
+			Timestamp:     timestamppb.Now(),
+			Message:       "Rollback in progress",
+		}
+		c.messagePipe.Process(&bus.Message{Topic: bus.InstanceConfigUpdateTopic, Data: rollbackInProgress})
+
+		err := c.configServices[instanceID].Rollback(c.messagePipe.Context(), skippedFiles,
```
olli-holmala, line 160: Adding a general comment here about our practice for passing contexts, feel free to close this thread before merging.

So I think we may potentially need to refactor our usage of `context.Context` inside our plugins. What we currently do is that we save the `ctx` to the `messagePipe` struct and we pass that same context to various functions that require a `context.Context`. Contexts should be request-bounded and should not be saved to structs: https://pkg.go.dev/context

A few solutions that come to mind: either always create a context with `context.WithCancel()` and append that cancel function to the `messagePipe` OR we keep the context in `messagePipe` and use that as a parent context for each request.

Thoughts @dhurley, @oliveromahony, @aphralG, @Dean-Coakley?

oliveromahony: I would have thought that on the intialization of the plugins we would create a new context from that, where the message pipe context is the parent. But based on that link this appears to be the incorrect way to do it. I guess we should widen the plugin interface to accept contexts then?

olli-holmala: > I guess we should widen the plugin interface to accept contexts then?

I'd say so if we want to conform to Go's conventions. It'll be important to pass contexts around correctly, as stuff like the correlation ID and tenant ID will likely be passed via the context.

olli-holmala (approved):

api/http/dataplane/data-plane-api.yaml
aphralG: I don't mind making the changes here to make the fields required, but ya might be best to discuss what needs to be required if all of them don't need to be required and do a follow up PR

internal/plugin/config_plugin.go
aphralG: Do I need to make the changes in this PR or will there be a follow up PR ?

olli-holmala: Let's do a follow-up PR, this is outside the scope of what you were doing.

oliveromahony (approved):

PR #588, opened 2024-03-07
Imporve unit test coverage

aphralG:
### Proposed changes

- Total Test Coverage at 86% now
- `internal/service` test coverage is now at 80%
- `internal/plugin` test coverage is now at 80%
- Switched all tests to use the same instance and correlation ID 
- Added helpers to create configuration status's
- Fixed issue with Rollback ConfigStatus message found while writing tests 

### Checklist


internal/plugin/config_plugin_test.go
```diff
@@ -152,10 +151,124 @@ func TestConfig_Process(t *testing.T) {
...
+			configService.RollbackReturns(test.rollbackReturns)
+
+			instanceService := []*instances.Instance{&instance}
+			configPlugin.configServices[instanceID] = configService
 			configPlugin.instances = instanceService
 
-			configPlugin.Process(test.input)
+			configPlugin.updateInstanceConfig(&request)
 
 			messages := messagePipe.GetMessages()
```
dhurley, line 274: Can you validate each message as well instead of just validating that we get the correct number of them?

internal/plugin/metrics_plugin_test.go
```diff
@@ -68,14 +70,59 @@ func TestMetrics_Process(t *testing.T) {
...
+		{
+			name:  "MetricsTopic can cast data",
+			topic: bus.MetricsTopic,
+			data:  dataEntry,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			metrics.Process(&bus.Message{Topic: test.topic, Data: test.data})
```
dhurley, line 123: Do you need an assert here?

internal/plugin/config_plugin_test.go
```diff
@@ -135,10 +124,20 @@ func TestConfig_Process(t *testing.T) {
 			},
 			expected: nil,
 		},
+		{
+			name: "Instance topic request ",
```
olli-holmala, line 128:
```suggestion
			name: "Instance topic request",
```

internal/service/config_service_test.go
```diff
@@ -52,56 +53,48 @@ func TestUpdateInstanceConfiguration(t *testing.T) {
 		name        string
 		writeErr    error
 		validateErr error
-		reloadErr   error
+		applyErr    error
+		completeErr error
 		expected    *instances.ConfigurationStatus
 	}{
 		{
 			name:        "write fails",
```
olli-holmala, line 61: I have a proposal: Should we adopt a naming convention for tests where we join words with either underscores or dashes?

Reasoning: When you have spaces in your test name, Go joins them with underscores when you run `go test ./...`. When a test fails, Go reports the failed test with the underscore concatenated name. You then have to manually text search for that specific test without the underscores.

e.g. if this test were to fail:
```bash
➜ service (imporve-unit-test-coverage) ✗ go test ./...
?   	github.com/nginx/agent/v3/internal/service/config/configfakes	[no test files]
?   	github.com/nginx/agent/v3/internal/service/instance/instancefakes	[no test files]
?   	github.com/nginx/agent/v3/internal/service/servicefakes	[no test files]
2024/03/07 11:10:44 ERROR Error writing config err="error writing config"
2024/03/07 11:10:44 ERROR Error validating config err="error validating config"
2024/03/07 11:10:44 ERROR Error applying config and reloading nginx err="error reloading config"
2024/03/07 11:10:44 ERROR Error completing config apply err="error completing config apply"
--- FAIL: TestUpdateInstanceConfiguration (0.00s)
    --- FAIL: TestUpdateInstanceConfiguration/write_fails (0.00s) # <--- Notice the new test name
        config_service_test.go:119:
            	Error Trace:	/Users/o.holmala/projects/agent/internal/service/config_service_test.go:119
            	Error:      	Not equal:
            	            	expected: "modified so this will fail"
            	            	actual  : "error writing config"

            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-modified so this will fail
            	            	+error writing config
            	Test:       	TestUpdateInstanceConfiguration/write_fails
FAIL
FAIL	github.com/nginx/agent/v3/internal/service	0.830s
ok  	github.com/nginx/agent/v3/internal/service/config	2.479s
ok  	github.com/nginx/agent/v3/internal/service/instance	1.021s
FAIL
```

So now when I want to search in my IDE for the failed test, I have to remove the underscores by hand instead of copypasting the result directly from terminal to my search box.

Thoughts?

P.S. This can be done in a follow-up PR so @aphralG doesn't have to implement my random ideas 😄

olli-holmala (approved):

internal/service/config_service_test.go
olli-holmala: We had a discussion at the office and decided that we want to enforce a convention where our test names are more descriptive, therefore having underscores or dashes will decrease readability too much. Opted to prepend test names with the test number, of the form "Test 1: <insert test name here>".

This will be done in a follow-up PR.

dhurley (approved):

PR #593, opened 2024-03-11
Centralize proto data for testing purposes

oliveromahony:
### Proposed changes

Centralize test data for configuration logic so it's easy to reuse the same test data across the project.

### Checklist


Dean-Coakley (approved):

internal/metrics/export/otel.go
```diff
@@ -29,7 +29,7 @@ const (
 )
 
 type OTelExporter struct {
-	conf           config.Config
+	conf           *config.Config
```
olli-holmala, line 32: Are we sure that we wanna pass a pointer to the config around instead of copying it around? I'd think a config struct should be immutable after it has been read from file.

olli-holmala (approved):

test/helpers/test_unit_utils.go
```diff
@@ -3,7 +3,7 @@
 // This source code is licensed under the Apache License, Version 2.0 license found in the
```
dhurley, line 3: can we rename this file to maybe `test_data.go`? And maybe move the functions that are creating & deleting files & directories to another file called `os_utils.go`.

test/types/config.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) F5, Inc.
```
dhurley, line 1: Should this be in `test/helpers`?

internal/config/config_test.go
```diff
@@ -387,3 +331,59 @@ func TestMissingServerTLS(t *testing.T) {
 	assert.Equal(t, expected.Auth, result.Auth)
 	assert.Nil(t, result.TLS)
 }
+
+func getAgentConfig() *Config {
```
dhurley, line 335: Should we be using the GetAgentConfig function in `test/type/config.go` here?

dhurley (commented):
Just a general comment. Should package names not be plural? Like for example should `test/helpers` be `test/helper`? We seem to be inconsistent in our naming convention for package names.

internal/config/config_test.go
oliveromahony: There is a circular dependency when using it here so I left this here

internal/metrics/export/otel.go
oliveromahony: It was inconsistent through the project. Being immutable is great, but I don't think we should have additional resources through copies of structs and it's more difficult if things change to update all the copies when in reality the config should be read once and only modified in some places

dhurley (approved):

PR #594, opened 2024-03-12
fixed the graceful shutdown of the app and the DataPlane API server

oliveromahony:
### Proposed changes

Fixed the graceful shutdown of the app.go and the DataPlane server API. Tidied much test code.

### Checklist

### DIFF
--- a/internal/app.go
+++ b/internal/app.go
@@ -7,7 +7,12 @@ package internal
 
 import (
 	"context"
+	"log"
 	"log/slog"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/nginx/agent/v3/internal/bus"
 	"github.com/nginx/agent/v3/internal/config"
@@ -57,6 +62,7 @@ func (a *App) Run() error {
 			return
 		}
 
+		handleSignals(ctx, cancel)
 		messagePipe.Run()
 	})
 	err := config.Execute()
@@ -65,4 +71,25 @@ func (a *App) Run() error {
 	}
 
 	return nil
+}
+
+// handleSignals handles signals to attempt graceful shutdown
+func handleSignals(
+	ctx context.Context,
+	cancel context.CancelFunc,
+) {
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		select {
+		case <-sigChan:
+			slog.Warn("NGINX Agent exiting")
+			cancel()
+
+			time.Sleep(config.DefGracefulShutdownPeriod)
+			// nolint: revive
+			log.Fatalf("Failed to gracefully shutdown within timeout of %v. Exiting", config.DefGracefulShutdownPeriod)
+		case <-ctx.Done():
+		}
+	}()
 }
--- a/internal/plugin/data_plane_server.go
+++ b/internal/plugin/data_plane_server.go
@@ -11,9 +11,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
-	"os/signal"
 	"strconv"
-	"syscall"
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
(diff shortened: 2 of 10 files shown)


internal/plugin/data_plane_server.go
```diff
@@ -34,7 +36,9 @@ type (
 		instances    []*instances.Instance
 		configEvents map[string][]*instances.ConfigurationStatus
 		messagePipe  bus.MessagePipeInterface
-		server       net.Listener
+		server       *http.Server
+		ctx          context.Context
```
olli-holmala, line 40: Is there another way to do this where we don't store the `context.Context` inside the struct?
https://pkg.go.dev/context
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

internal/plugin/data_plane_server.go
```diff
@@ -50,16 +54,25 @@ func NewDataPlaneServer(agentConfig *config.Config, logger *slog.Logger) *DataPl
...
+	go dps.run()
 
 	return nil
 }
 
 func (dps *DataPlaneServer) Close() error {
-	if dps.server != nil {
-		dps.server.Close()
+	dps.cncl()
+	dps.logger.Debug("shutting down gracefully, press Ctrl+C again to force")
```
olli-holmala, line 64: Should this be Info level?

internal/plugin/data_plane_server.go
```diff
@@ -99,26 +112,30 @@ func (*DataPlaneServer) Subscriptions() []string {
 	}
 }
 
-func (dps *DataPlaneServer) run(_ context.Context) {
+func (dps *DataPlaneServer) run() {
```
olli-holmala, line 115: The context that was being passed here (the `messagePipe`'s context) was already listening to the `SIGINT` and `SIGTERM` signals, as that is the "root" context that is initialized in `app.go`.

internal/app.go
```diff
@@ -66,3 +72,24 @@ func (a *App) Run() error {
 
 	return nil
 }
+
+// handleSignals handles signals to attempt graceful shutdown
+func handleSignals(
```
olli-holmala, line 77: Why not just manage the lifecycle of `App` with the parent context? We could just use `signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)` on line 42 and the context would automatically get cancelled when we receive an interrupt signal.

We could then pass that context around to child goroutines, who would then get cancelled if the parent context is cancelled.

internal/plugin/metrics_plugin.go
```diff
@@ -26,60 +26,60 @@ const (
...
-	shutdownFunc func()
-
 	// The Metrics plugin. Discovers and owns the data source that produce metrics for the Agent.
 	Metrics struct {
 		producers       map[model.MetricsSourceType]model.MetricsProducer
 		exporters       map[model.ExporterType]model.Exporter
-		conf            config.Config
+		config          *config.Config
 		pipe            bus.MessagePipeInterface
+		shutdownFuncs   []*context.CancelFunc
```
olli-holmala, line 35: nit: `context.CancelFunc` is a function reference, so I don't understand why it would need to be a pointer?
```suggestion
		shutdownFuncs   []context.CancelFunc
```

internal/plugin/metrics_plugin.go
```diff
@@ -186,12 +190,14 @@ func (m *Metrics) startExporters(ctx context.Context) {
 
 // Produces metrics from the MetricsProducer at the configured interval and pushes to the given channel.
 func (m *Metrics) runProducer(ctx context.Context, producer model.MetricsProducer) {
-	t := time.NewTicker(m.collectInterval)
+	m.metricsTimer = time.NewTicker(m.collectInterval)
+
+	defer m.metricsTimer.Stop()
```
olli-holmala, line 195: I'd like to better understand what the intention here was. So now instead of each producer having its own ticker, we are saving the ticker of the last started producer to `m.metricsTimer` and stopping our global ticker if **one** producer goes out of scope.

If the above behavior is what we want (i.e. all producers are tied to one another) then let's at least init the ticker one level higher, and potentially pass it as a parameter to `runProducer`, instead of in `runProducer`, which will be run multiple times.

I would think it would be better that producers are independent from one another but maybe there's something I haven't thought of.

oliveromahony: I didn't quite get that each producer had it's own timer here. If that's the case, I think the producer should own the timer? The metrics timer is specific to the metrics plugin in the current code, I was trying to tidy it up but I can revert for now until we decide on one way or another

olli-holmala (approved):

PR #595, opened 2024-03-13
Config Apply Delete File 

aphralG:
### Proposed changes

- Compare the `fileCache` with the cache from updating files and if file is no longer in the file meta data delete it. 
- Changed acceptable code complexity levels in the linter
- Add check that cache is not nil before `doesFileRequireUpdate` 

### Checklist


internal/datasource/config/config_writer.go
```diff
@@ -100,25 +99,44 @@ func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
...
 	}
 
 	cw.currentFileCache = currentFileCache
 
 	return skippedFiles, err
 }
 
+func (cw *ConfigWriter) removeFiles(currentFileCache, fileCache CacheContent) {
+	for _, file := range fileCache {
+		_, ok := currentFileCache[file.GetPath()]
```
oliveromahony, line 127:
```suggestion
		if _, ok := currentFileCache[file.GetPath()]; !ok {
```

internal/datasource/config/config_writer.go
```diff
@@ -100,25 +99,44 @@ func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
...
+		if !ok {
+			slog.Debug("removing file", "file_path", file.GetPath())
+			err := os.Remove(file.GetPath())
```
oliveromahony, line 130:
```suggestion
			if err := os.Remove(file.GetPath())l err != nil {
```

internal/datasource/config/config_writer.go
```diff
@@ -100,25 +99,44 @@ func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
...
+			if err != nil {
+				slog.Error("error removing file: ", "err", err)
+			}
+
+			continue
```
oliveromahony, line 135: is continue required?

internal/datasource/config/config_writer.go
```diff
@@ -100,25 +99,44 @@ func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
...
 }
 
+func (cw *ConfigWriter) removeFiles(currentFileCache, fileCache CacheContent) {
+	for _, file := range fileCache {
+		_, ok := currentFileCache[file.GetPath()]
+		if !ok {
+			slog.Debug("removing file", "file_path", file.GetPath())
+			err := os.Remove(file.GetPath())
+			if err != nil {
+				slog.Error("error removing file: ", "err", err)
```
dhurley, line 132: If we fail to delete a file then we should return an error from this function. A config apply should fail if we can't update or delete files.

internal/datasource/config/config_writer_test.go
```diff
@@ -131,6 +131,73 @@ func TestWriteConfig(t *testing.T) {
...
+			if test.fileDeleted {
+				assert.NoFileExists(t, testConf.Name())
+			} else {
+				assert.FileExists(t, testConf.Name())
+			}
+			helpers.RemoveFileWithErrorCheck(t, cacheFile.Name())
+		})
+	}
+
+	helpers.RemoveFileWithErrorCheck(t, metricsConf.Name())
```
dhurley, line 197: This functions might not get called if the unit test fails. Its best to call `defer helpers.RemoveFileWithErrorCheck` straight away after calling `helpers.CreateFileWithErrorCheck` to insure that the file gets deleted at the end of the test.

internal/datasource/config/config_writer.go
aphralG: No left it in after making changes, removing

internal/datasource/config/config_writer.go
aphralG: Ok, I wasn't sure if we should be failing the config apply if we can't remove a file. I'll change it back to returning an error from that function.

internal/datasource/config/config_writer.go
```diff
@@ -100,25 +99,42 @@ func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
...
 	return skippedFiles, err
 }
 
+func (cw *ConfigWriter) removeFiles(currentFileCache, fileCache CacheContent) error {
+	for _, file := range fileCache {
+		if _, ok := currentFileCache[file.GetPath()]; !ok {
+			slog.Debug("removing file", "file_path", file.GetPath())
+			if err := os.Remove(file.GetPath()); err != nil {
+				return fmt.Errorf("error removing file: %s error: %w", file.GetPath(), err)
+			}
```
olli-holmala, line 131: Should we add a check here for `os.IsNotExist(err))`, in case the file was already deleted before we attempt to delete it ourselves?

olli-holmala: Something like:
```suggestion
			if err := os.Remove(file.GetPath()); err != nil && !os.IsNotExist(err) {
				return fmt.Errorf("error removing file: %s error: %w", file.GetPath(), err)
			}
```

internal/datasource/config/config_writer.go
```diff
@@ -100,25 +99,42 @@ func (cw *ConfigWriter) Write(ctx context.Context, filesURL,
...
+		err = cw.removeFiles(currentFileCache, cacheContent)
+	}
+
 	return skippedFiles, err
 }
 
+func (cw *ConfigWriter) removeFiles(currentFileCache, fileCache CacheContent) error {
+	for _, file := range fileCache {
+		if _, ok := currentFileCache[file.GetPath()]; !ok {
+			slog.Debug("removing file", "file_path", file.GetPath())
```
dhurley, line 128:
```suggestion
			slog.Debug("Removing file", "file_path", file.GetPath())
```

dhurley (approved):

olli-holmala (approved):

Dean-Coakley (approved):

oliveromahony (approved):

PR #608, opened 2024-03-20
Refactor how context is used

dhurley:
### Proposed changes

- Refactor how context is used
- Refactor how we gracefully shutdown the agent

### Checklist

### DIFF
--- a/internal/bus/message_pipe.go
+++ b/internal/bus/message_pipe.go
@@ -27,19 +27,18 @@ type (
 
 	MessagePipeInterface interface {
 		Register(size int, plugins []Plugin) error
-		DeRegister(plugins []string) error
+		DeRegister(ctx context.Context, plugins []string) error
 		Process(messages ...*Message)
-		Run()
-		Context() context.Context
+		Run(ctx context.Context)
 		GetPlugins() []Plugin
 		IsPluginAlreadyRegistered(pluginName string) bool
 	}
 
 	Plugin interface {
-		Init(messagePipe MessagePipeInterface) error
-		Close() error
+		Init(ctx context.Context, messagePipe MessagePipeInterface) error
+		Close(ctx context.Context) error
 		Info() *Info
-		Process(msg *Message)
+		Process(ctx context.Context, msg *Message)
 		Subscriptions() []string
 	}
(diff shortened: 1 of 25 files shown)


internal/bus/message_pipe.go
```diff
@@ -83,36 +78,76 @@ func (p *MessagePipe) Register(size int, plugins []Plugin) error {
...
+				r.Close(ctx)
+			}
+
+			close(p.messageChannel)
+
+			return
+		case m := <-p.messageChannel:
+			p.mu.Lock()
+			p.bus.Publish(m.Topic, ctx, m)
+			p.mu.Unlock()
```
olli-holmala, line 125: I know that this isn't new code but I'm curious why we are locking for the duration of the publish?

The `Publish()` implementation already uses a lock:
```go
func (b *messageBus) Publish(topic string, args ...interface{}) {
	rArgs := buildHandlerArgs(args)

	b.mtx.RLock()
	defer b.mtx.RUnlock()

	if hs, ok := b.handlers[topic]; ok {
		for _, h := range hs {
			h.queue <- rArgs
		}
	}
}
```

olli-holmala (approved):

internal/bus/message_pipe.go
dhurley: Good question. I will have another PR going up after this one where I am going to review our use of `sync.Mutex`. I'll address it then

aphralG (approved):

Dean-Coakley (approved):

internal/bus/message_pipe.go
```diff
@@ -83,36 +78,76 @@ func (p *MessagePipe) Register(size int, plugins []Plugin) error {
...
+	for _, plugin := range p.GetPlugins() {
+		if plugin.Info().Name == pluginName {
+			pluginAlreadyRegistered = true
+		}
+	}
+
+	return pluginAlreadyRegistered
+}
+
+func (p *MessagePipe) unsubscribePlugin(ctx context.Context, index int, plugin Plugin) error {
```
oliveromahony, line 146: unsubscribe seems sufficient here

PR #611, opened 2024-03-21
Race condition detection

dhurley:
### Proposed changes

- Add race condition detection to pipeline
- Fix race condition errors in code

### Checklist


Makefile
```diff
@@ -102,6 +102,9 @@ unit-test: $(TEST_BUILD_DIR) ## Run unit tests
 	@$(GOTOOL) cover -html=$(TEST_BUILD_DIR)/coverage.out -o $(TEST_BUILD_DIR)/coverage.html
 	@printf "\nTotal code coverage: " && $(GOTOOL) cover -func=$(TEST_BUILD_DIR)/coverage.out | grep 'total:' | awk '{print $$3}'
 
+unit-test-with-race-condition-detection: $(TEST_BUILD_DIR) ## Run unit tests with race condition detection
```
olli-holmala, line 105: Can't we use something like `race-unit-test` or `unit-test-with-race`? This feels a little too verbose for my taste.

internal/bus/message_pipe.go
```diff
@@ -46,20 +46,20 @@ type (
...
-		mu             sync.RWMutex
+		pluginsMutex   sync.Mutex
 	}
 )
 
 func NewMessagePipe(size int) *MessagePipe {
 	return &MessagePipe{
 		messageChannel: make(chan *Message, size),
-		mu:             sync.RWMutex{},
+		pluginsMutex:   sync.Mutex{},
```
olli-holmala, line 56: Why the switch to a `Mutex` instead of a `RWMutex`?

internal/bus/message_pipe.go
```diff
@@ -116,13 +116,9 @@ func (p *MessagePipe) Run(ctx context.Context) {
 				r.Close(ctx)
 			}
 
-			close(p.messageChannel)
-
```
olli-holmala, line 120: Why are we no longer closing the channel?

internal/plugin/data_plane_server.go
```diff
@@ -29,23 +30,29 @@ type (
...
 
 	return &DataPlaneServer{
-		address:      address,
-		logger:       logger,
-		configEvents: make(map[string][]*instances.ConfigurationStatus),
+		address:           address,
+		logger:            logger,
+		configEvents:      make(map[string][]*instances.ConfigurationStatus),
+		serverMutex:       &sync.Mutex{},
+		configEventsMutex: &sync.Mutex{},
```
olli-holmala, line 54: What's the distribution of reads and writes of config events? Would we expect more reads than writes, or is it roughly equal?

internal/plugin/data_plane_server.go
```diff
@@ -29,23 +30,29 @@ type (
...
+		instancesMutex:    &sync.Mutex{},
```
olli-holmala, line 55: Ditto here, do we expect the list of instances to change very often? I would think that the list of instances changes fairly rarely (writes are rare) but reads of the instances list to occur more regularly => `RWMutex` would be better.

internal/plugin/instance_plugin.go
```diff
@@ -57,7 +60,9 @@ func (i *Instance) Process(_ context.Context, msg *bus.Message) {
 
 		instanceList := i.instanceService.GetInstances(newProcesses)
 		if len(instanceList) > 0 {
+			i.instancesMutex.Lock()
 			i.instances = instanceList
+			i.instancesMutex.Unlock()
```
olli-holmala, line 65: I'm curious why we are storing the instance list in the plugin, since it's not used anywhere?

internal/bus/message_pipe.go
```diff
@@ -116,13 +116,9 @@ func (p *MessagePipe) Run(ctx context.Context) {
...
 			return
 		case m := <-p.messageChannel:
-			p.mu.Lock()
```
oliveromahony, line 123: was this removed because bus has it's own mutex?

dhurley: This was removed as a mutex here isn't required. There is no shared resource that needs to be locked.

internal/plugin/instance_plugin.go
dhurley: Its used in the unit tests

internal/plugin/data_plane_server.go
```diff
@@ -121,10 +137,12 @@ func (dps *DataPlaneServer) run(ctx context.Context) {
 	router.Use(sloggin.NewWithConfig(dps.logger, sloggin.Config{DefaultLevel: slog.LevelDebug}))
 	dataplane.RegisterHandlersWithOptions(router, dps, dataplane.GinServerOptions{BaseURL: "/api/v1"})
 
+	dps.serverMutex.Lock()
 	dps.server = &http.Server{
 		Addr:    dps.address,
 		Handler: router,
 	}
+	dps.serverMutex.Unlock()
```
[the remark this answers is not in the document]
dhurley: Its need by the `getServerAddress` function as well used in the unit test.

internal/bus/message_pipe.go
dhurley: It caused a panic when the agent is exiting. 
The process function was trying to send a message to a channel that was closed.
There is no need to close the channel anyway since the only time it should be closed when when the agent is being stopped.

Makefile
dhurley: Maybe `race-condition-test` might be better?

internal/plugin/instance_plugin.go
olli-holmala: Are we sure that we wanna add extra overhead (and locking) to production code just to make our code more testable?

dhurley: thats a far point. Looking at the code again I think I might be able to test the code a different way which doesn't require locks

Makefile
olli-holmala: Yeah `race-condition-test` is definitely the clearest option 👌

internal/plugin/instance_plugin.go
dhurley: updated code to remove mutex

internal/bus/message_pipe.go
dhurley: switched to be consistent with other plugins

internal/plugin/data_plane_server.go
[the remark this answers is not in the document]
dhurley: The Init and Close functions are both running in their own go routines and are both accessing the server variable in the struct so a lock is required. I know that it is very unlikely that we will call the two functions at the same time, but there is no harm in adding the extra check.

internal/plugin/data_plane_server.go
olli-holmala: As discussed on Slack, we will default to `sync.Mutex` unless we notice a need for a `sync.RWMutex` (when it is obvious that there are far more reads than writes).

olli-holmala (approved):

aphralG (approved):

internal/plugin/data_plane_server.go
dhurley: Sticking with Mutex unless we see a need for RWMutex

PR #617, opened 2024-03-25
Keep looking for master process

spencerugbo:
### Proposed changes

Continuously searches for master process

### Checklist


src/plugins/registration.go
```diff
@@ -29,6 +29,7 @@ const (
 	dataplaneSoftwareDetailsMaxWaitTime = time.Duration(5 * time.Second)
 	// Time between attempts to gather DataplaneSoftwareDetails
 	softwareDetailsOperationInterval = time.Duration(1 * time.Second)
+	nginxStartMaxWaitTime            = 0
```
olli-holmala, line 32: Would this not essentially turn off backoff altogether, if the `MaxElapsedTime` is zero? If that is intentional, then I would maybe adding a comment explaining why it is turned off.

src/plugins/registration.go
```diff
@@ -153,23 +156,43 @@ func (r *OneTimeRegistration) areDataplaneSoftwareDetailsReady() error {
...
+					_, err := r.binary.ReadConfig(nginxDetails.GetConfPath(), nginxDetails.NginxId, r.env.GetSystemUUID())
+					if err != nil {
+						log.Warnf("Unable to read config for NGINX instance %s, %v", nginxDetails.NginxId, err)
+					}
+					log.Info("Master process has been found")
+					return nil
+				}
+			} else {
+				log.Tracef("NGINX non-master process: %d", proc.Pid)
+				return nil
```
olli-holmala, line 184: This `return` causes the whole function to exit, meaning that the function will only ever find the master process if it is the first process in `r.processes`.

src/plugins/registration.go
spencerugbo: When the max wait time is set to 0 backoff will continue waiting indefinitely until the master process is found. I'll add in a comment now

src/plugins/registration.go
```diff
@@ -153,23 +156,43 @@ func (r *OneTimeRegistration) areDataplaneSoftwareDetailsReady() error {
...
+		for _, proc := range r.processes {
+			// only need master process for registration
+			if proc.IsMaster {
+				nginxDetails := r.binary.GetNginxDetailsFromProcess(proc)
+				details = append(details, nginxDetails)
+				if len(details) > 0 {
+					// Reading nginx config during registration to populate nginx fields like access/error logs, etc.
+					_, err := r.binary.ReadConfig(nginxDetails.GetConfPath(), nginxDetails.NginxId, r.env.GetSystemUUID())
+					if err != nil {
+						log.Warnf("Unable to read config for NGINX instance %s, %v", nginxDetails.NginxId, err)
```
olli-holmala, line 177: I know that this line isn't originally your code, but I'm wondering if we should actually be returning the error here instead of merely logging it 🤔 

@oliveromahony, @dhurley, @Dean-Coakley, @aphralG thoughts?

src/plugins/registration.go
```diff
@@ -153,23 +156,43 @@ func (r *OneTimeRegistration) areDataplaneSoftwareDetailsReady() error {
...
 		}
+		return fmt.Errorf("No master process found, waiting...")
 	}
-	if len(details) == 0 {
-		log.Info("No master process found")
+	err := backoff.WaitUntil(
+		context.Background(), backoffSetting, findMaster,
+	)
+	if err != nil {
+		log.Warn(err.Error())
```
olli-holmala, line 193: This was previously:
```
    log.Info("No master process found")
```
But my instinct is that this should be an error log, as I'd think we can't complete registration if we haven't found a master process.

Others could also weigh in on this, as I wouldn't consider myself an expert on this.

src/plugins/registration.go
dhurley: I believe the reason we dont return an error here is because we dont want to stop registration just because we cant read the config. If the config is invalid it should not stop the agent from performing other tasks like reporting the status of NGINX is that makes sense.

src/plugins/registration.go
dhurley: Yes I believe previously we would continue with registration even if we didn't find master processes but now that we are waiting until we find one I think it should be an error log message.
```
log.Errorf("Unable to find NGINX master processes, %v", err)
```

src/plugins/registration.go
```diff
@@ -153,23 +159,42 @@ func (r *OneTimeRegistration) areDataplaneSoftwareDetailsReady() error {
...
-				log.Warnf("Unable to read config for NGINX instance %s, %v", nginxDetails.NginxId, err)
+	backoffSetting := backoff.BackoffSettings{
+		InitialInterval: softwareDetailsOperationInterval,
+		MaxInterval:     softwareDetailsOperationInterval,
+		MaxElapsedTime:  nginxStartMaxWaitTime,
+		Jitter:          backoff.BACKOFF_JITTER,
+		Multiplier:      backoff.BACKOFF_MULTIPLIER,
+	}
+
+	findMaster := func() error {
```
dhurley, line 170:
```suggestion
	findNginxMasterProcesses := func() error {
```

dhurley (approved):

PR #620, opened 2024-03-26
Build packages for integration test

aphralG:
### Proposed changes

Add make target that checks the OS_RELEASE and checks if the package is already there before running the correct make local package target for that OS Release. make integration tests now uses this target

### Checklist


Makefile
```diff
@@ -194,3 +194,6 @@ generate-pgo-profile: build-mock-management-plane-http build-mock-management-pla
 	@CGO_ENABLED=0 $(GOTEST) -count 10 -timeout 2m -bench=. -benchmem -run=^# ./internal/service/instance -cpuprofile perf_instance_cpu.pprof
 	@$(GOTOOL) pprof -proto perf_config_cpu.pprof perf_instance_cpu.pprof integration_cpu.pprof > default.pgo
 	rm perf_config_cpu.pprof perf_instance_cpu.pprof config.test instance.test integration_cpu.pprof integration.test profile.pprof
+
+build-test-package: 
+	if ! [ -e ./build/*.deb ] && [ $(OS_RELEASE) == 'ubuntu' ]; then make local-deb-package; elif ! [ -e ./build/*.rpm ] && [ $(OS_RELEASE) == 'redhatenterprise'  ]; then make local-rpm-package; elif ! [ -e ./build/*.apk ] && [ $(OS_RELEASE) == 'alpine' ]; then make local-apk-package; fi
```
olli-holmala, line 199: This isn't a very "Makefiley" way to do things. The "normal" way to use Makefiles is to name the target as the output file (e.g. if the output of the target is `build/agent.deb` then the name of the target should also be `build/agent.deb`).

To achieve that, we would need a few tweaks to the Makefile:
```makefile
VERSION_NUMBER  = $(shell echo $(VERSION) | tr -d 'v')
APK_PACKAGE := $(BUILD_DIR)/$(PACKAGE_PREFIX)-$(VERSION_NUMBER)-SNAPSHOT-$(COMMIT).apk
DEB_PACKAGE := $(BUILD_DIR)/$(PACKAGE_PREFIX)-$(VERSION_NUMBER)-SNAPSHOT-$(COMMIT).deb
RPM_PACKAGE := $(BUILD_DIR)/$(PACKAGE_PREFIX)-$(VERSION_NUMBER)-SNAPSHOT-$(COMMIT).rpm

$(APK_PACKAGE):
	@echo "Building .apk package..."
	@$(MAKE) local-apk-package

$(DEB_PACKAGE):
	@echo "Building .deb package..."
	@$(MAKE) local-deb-package

$(RPM_PACKAGE):
	@echo "Building .rpm package..."
	@$(MAKE) local-rpm-package
```

You would then have to have a conditional variable to determine the correct package for the selected `OS_RELEASE`:

```makefile
// This is our default
SELECTED_PACKAGE = $(DEB_PACKAGE)
ifeq ($(OS_RELEASE),redhatenterprise)
	SELECTED_PACKAGE = $(RPM_PACKAGE)
endif

ifeq ($(OS_RELEASE),alpine)
	SELECTED_PACKAGE = $(APK_PACKAGE)
endif
```

And finally you'd add the `SELECTED_PACKAGE` as a dependency to your test target:
```makefile
integration-test: $(SELECTED_PACKAGE) build-mock-management-plane-http build-mock-management-plane-grpc
```

This way `make` **should** automatically run the correct build target based on whether the package exists. Haven't tested the above, so might need some adjustment to get it working.

We don't necessarily need to do it the idiomatic way though, and we can just continue the convention of using Make targets for running scripts. If we keep it as is, then please add `build-test-package` to `.PHONY` to tell `make` that the target will not produce a file called `build-test-package`.

Makefile
```diff
@@ -174,15 +197,15 @@ generate-mocks: ## Regenerate all needed mocks, in order to add new mocks genera
 
 local-apk-package: ## Create local apk package
 	@CGO_ENABLED=0 GOARCH=$(OSARCH) GOOS=linux $(GOBUILD) -o $(BUILD_DIR)/$(BINARY_NAME) -pgo=default.pgo -ldflags=$(LDFLAGS) $(PROJECT_DIR)/$(PROJECT_FILE)
-	ARCH=$(OSARCH) VERSION=$(shell echo $(VERSION) | tr -d 'v') $(GORUN) $(NFPM) pkg --config ./scripts/packages/.local-nfpm.yaml --packager apk --target ./build/$(PACKAGE_PREFIX)-$(shell echo $(VERSION) | tr -d 'v')-SNAPSHOT-$(COMMIT).apk;
+	ARCH=$(OSARCH) VERSION=$(shell echo $(VERSION) | tr -d 'v') $(GORUN) $(NFPM) pkg --config ./scripts/packages/.local-nfpm.yaml --packager apk --target ./build/$(PACKAGE_NAME).apk;
```
dhurley, line 200:
```suggestion
	ARCH=$(OSARCH) VERSION=$(shell echo $(VERSION) | tr -d 'v') $(GORUN) $(NFPM) pkg --config ./scripts/packages/.local-nfpm.yaml --packager apk --target ${APK_PACKAGE);
```

dhurley (approved):

Dean-Coakley (approved):

PR #623, opened 2024-03-27
Compare performance test results

aphralG:
### Proposed changes

Added compare bench mark results to pipeline.

- performance tests are ran 
- then the results are compared to the results from v3
- if the performance has gotten worse (the default is 200%) then the pipeline will fail 
- when the pipeline is run on main the results from the performance tests will be committed to the [benchmark-results](https://github.com/nginx/agent/tree/benchmark-results) branch.
- this branch contains `data.js` which has the previous results over time and `index.htm`l which puts the results in a graph 

### Checklist


docs/proto/protos.md
```diff
@@ -1,689 +0,0 @@
-# Protocol Documentation
```
dhurley, line 1: Guessing this got deleted by mistake?

aphralG: Ya my bad will fix now

.github/workflows/ci.yml
```diff
@@ -128,10 +127,28 @@ jobs:
...
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version-file: 'go.mod'
       - name: Run Performance Tests
-        run: make performance-test
+        run: |
+          make performance-test
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
```
Dean-Coakley, line 141: As far as I understand it's best practice to pin actions to a full length SHA

```suggestion
        uses: benchmark-action/github-action-benchmark@cc9ac13ce81036c9b67fcfe2cb95ca366684b9ea # v1.19.3
```

Ref: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

olli-holmala (approved):

oliveromahony (approved):

PR #624, opened 2024-03-29
Add correlation id to context

dhurley:
### Proposed changes

- Add correlation id to context
- Log correlation id when present in context

### Checklist

### DIFF
--- a/internal/bus/message_pipe.go
+++ b/internal/bus/message_pipe.go
@@ -21,6 +21,11 @@ type (
 		Data  Payload
 	}
 
+	MessageWithContext struct {
+		ctx     context.Context
+		message *Message
+	}
+
 	Info struct {
 		Name string
 	}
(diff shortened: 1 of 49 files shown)


internal/bus/message_pipe.go
```diff
@@ -21,14 +21,19 @@ type (
 		Data  Payload
 	}
 
+	MessageWithContext struct {
+		ctx     context.Context
+		message *Message
+	}
```
olli-holmala, line 27: Have we considered adding `context.Context` inside the `Message` struct?

dhurley: The reason for the `MessageWithContext` is that we dont want to be passing around the context inside a struct in the plugins. Its only used for sending the context and message over the message pipe channel.

I don't think it would be good to change the plugin Process function from `Process(ctx context.Context, msg *Message)` to `Process(msg *Message)` as its not clear that the message has the context. Always best to have the context as the first parameter in a function.

oliveromahony (approved):

PR #628, opened 2024-04-03
Send data plane status update request

dhurley:
### Proposed changes

- Send data plane status update request
- Update process monitor to only send updates if the process have changed 

### Checklist


internal/plugin/grpc_client_plugin.go
```diff
@@ -124,10 +126,24 @@ func (gc *GrpcClient) Info() *bus.Info {
...
 }
 
-func (gc *GrpcClient) Process(_ context.Context, _ *bus.Message) {}
+func (gc *GrpcClient) Process(ctx context.Context, msg *bus.Message) {
+	switch {
+	case msg.Topic == bus.InstancesTopic:
+		if newInstances, ok := msg.Data.([]*v1.Instance); ok {
+			err := gc.sendDataPlaneStatusUpdate(ctx, newInstances)
+			if err != nil {
+				slog.Error("Unable to send data plane status update", "error", err)
```
olli-holmala, line 135: Is there no `slog.ErrorContext`?
```suggestion
				slog.ErrorContext(ctx, "Unable to send data plane status update", "error", err)
```

internal/plugin/process_monitor_plugin.go
```diff
@@ -105,22 +105,45 @@ func (pm *ProcessMonitor) run(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-pm.processTicker.C:
-			newCtx := context.WithValue(ctx, logger.CorrelationIDContextKey, logger.GenerateCorrelationID())
-			slog.DebugContext(newCtx, "Checking for process changes")
+			slog.Debug("Checking for process updates")
```
olli-holmala, line 108: Why the change to just `slog.Debug()`? Maybe we should clarify when we should log with context and when not to 🤔

dhurley: my bad. It should be DebugContext. We should always pass context to the logger if it is available in the function.

olli-holmala (approved):

aphralG (approved):

internal/plugin/process_monitor_plugin.go
```diff
@@ -105,22 +105,45 @@ func (pm *ProcessMonitor) run(ctx context.Context) {
...
+				haveChanged = true
+				break
+			}
+		}
+	} else {
+		haveChanged = true
 	}
+
+	return haveChanged
 }
```
Dean-Coakley, line 149: Totally optional but I think something more like the following would be easier to read:
```go
func haveProcessesChanged(oldProcesses, newProcesses []*model.Process) bool {
	if len(oldProcesses) != len(newProcesses) {
		return true
	}

	oldPIDMap := make(map[int32]struct{})
	for _, oldProcess := range oldProcesses {
		oldPIDMap[oldProcess.Pid] = struct{}{}
	}

	for _, newProcess := range newProcesses {
		if _, ok := oldPIDMap[newProcess.Pid]; !ok {
			return true
		}
	}

	return false
}
```

Dean-Coakley (approved):

internal/plugin/process_monitor_plugin.go
dhurley: Updated and added some comments to make it easier to read.

craigell (approved):

PR #631, opened 2024-04-09
Fix send data plane status update nil pointer exception

dhurley:
### Proposed changes

Need to check that the gRPC client is connected before sending a data plane status update

### Checklist

### DIFF
--- a/internal/plugin/grpc_client_plugin.go
+++ b/internal/plugin/grpc_client_plugin.go
@@ -186,9 +186,9 @@ func (gc *GrpcClient) Process(ctx context.Context, msg *bus.Message) {
 		gc.isConnected.Store(true)
 
 		gc.instancesMutex.Lock()
-		defer gc.instancesMutex.Unlock()
-
 		err := gc.sendDataPlaneStatusUpdate(ctx, gc.instances)
+		gc.instancesMutex.Unlock()
+
 		if err != nil {
 			slog.ErrorContext(ctx, "Unable to send data plane status update", "error", err)
 		}
--- a/internal/plugin/grpc_client_plugin.go
+++ b/internal/plugin/grpc_client_plugin.go
@@ -11,6 +11,7 @@ import (
 	"log/slog"
 	"net"
 	"sync"
+	"sync/atomic"
 
 	"github.com/google/uuid"
 	"github.com/nginx/agent/v3/api/grpc/mpi/v1"
@@ -31,7 +32,7 @@ type (
 		messagePipe          bus.MessagePipeInterface
 		config               *config.Config
 		conn                 *grpc.ClientConn
-		isConnected          bool
+		isConnected          *atomic.Bool
 		commandServiceClient v1.CommandServiceClient
 		cancel               context.CancelFunc
 		instances            []*v1.Instance
@@ -47,9 +48,12 @@ func NewGrpcClient(agentConfig *config.Config) *GrpcClient {
 			return nil
 		}
 
+		isConnected := &atomic.Bool{}
+		isConnected.Store(false)
+
 		return &GrpcClient{
 			config:          agentConfig,
-			isConnected:     false,
+			isConnected:     isConnected,
 			instances:       []*v1.Instance{},
 			connectionMutex: sync.Mutex{},
 			instancesMutex:  sync.Mutex{},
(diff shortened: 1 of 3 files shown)


internal/plugin/grpc_client_plugin.go
```diff
@@ -161,13 +167,26 @@ func (gc *GrpcClient) Process(ctx context.Context, msg *bus.Message) {
...
+			gc.instancesMutex.Unlock()
+
 			err := gc.sendDataPlaneStatusUpdate(ctx, newInstances)
 			if err != nil {
 				slog.ErrorContext(ctx, "Unable to send data plane status update", "error", err)
 			}
 		}
 	case bus.GrpcConnectedTopic:
 		slog.DebugContext(ctx, "Agent connected")
+		gc.isConnected = true
```
oliveromahony, line 181: this should have a lock around it. Probably better to use an atomic bool

internal/plugin/grpc_client_plugin.go
```diff
@@ -46,7 +49,10 @@ func NewGrpcClient(agentConfig *config.Config) *GrpcClient {
 
 		return &GrpcClient{
 			config:          agentConfig,
+			isConnected:     false,
+			instances:       []*v1.Instance{},
```
olli-holmala, line 53: It seems a little bit weird to me that a gRPC client would have a list of instances. Would it be better if the status update was handled by the instance plugin instead of our gRPC client?

internal/plugin/grpc_client_plugin.go
dhurley: sure I'll update that

internal/plugin/grpc_client_plugin.go
dhurley: If the gRPC client plugin is disabled then the instances plugin wont know that and try to make a request.
The instance plugin should only be concerned with finding instances and nothing else.

The only reason I need to store them is because the Process function is getting called before the Init function which might need further investigation to see if there is a way to prevent this.

internal/plugin/grpc_client_plugin.go
```diff
@@ -120,7 +130,8 @@ func (gc *GrpcClient) createConnection() error {
 		},
 	}
 
-	reqCtx, reqCancel := context.WithTimeout(context.Background(), gc.config.Common.MaxElapsedTime)
+	ctx := context.Background()
```
oliveromahony, line 133: not sure this change is necessary?

internal/plugin/grpc_client_plugin.go
olli-holmala: Yeah not calling `Init()` before `Process()` sounds like a problem alright. Do you wanna look at it in another PR or do it here?

internal/plugin/grpc_client_plugin.go
dhurley: its needed because I cant pass `reqCtx` to the Process function below. The `reqCtx` times out when I am sending the data plane status update

internal/plugin/grpc_client_plugin.go
```diff
@@ -161,13 +172,26 @@ func (gc *GrpcClient) Process(ctx context.Context, msg *bus.Message) {
...
 			if err != nil {
 				slog.ErrorContext(ctx, "Unable to send data plane status update", "error", err)
 			}
 		}
 	case bus.GrpcConnectedTopic:
 		slog.DebugContext(ctx, "Agent connected")
+		gc.isConnected.Store(true)
+
+		gc.instancesMutex.Lock()
+		defer gc.instancesMutex.Unlock()
```
olli-holmala, line 189: Small side note, while this is not a bug right now, this has the potential to introduce bugs if we add code after the `switch`, as the `defer` is only executed at the end of **function** block, not the current code block (i.e. the `case` block of this `switch`). So it would technically be safer to manually unlock it after line 191.

This could be something to mention in our style guide, assuming it is not already in the Uber one.

dhurley: thats a fair point. Let me update that

dhurley: updated

oliveromahony (approved):

PR #633, opened 2024-04-10
Add uuid and rename token to authorization header for gRPC connections

oliveromahony:
### Proposed changes

Updated interceptor headers. Added uuid to agent config

### Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -54,7 +54,13 @@ func RegisterConfigFile() error {
 
 	slog.Debug("Configuration file loaded", "config_path", configPath)
 	viperInstance.Set(ConfigPathKey, configPath)
-	viperInstance.Set(UUIDKey, uuidLibrary.Generate(configPath))
+
+	exePath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	viperInstance.Set(UUIDKey, uuidLibrary.Generate(exePath, configPath))
 
 	return nil
 }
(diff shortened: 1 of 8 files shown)


internal/config/config.go
```diff
@@ -53,12 +54,14 @@ func RegisterConfigFile() error {
 
 	slog.Debug("Configuration file loaded", "config_path", configPath)
 	viperInstance.Set(ConfigPathKey, configPath)
+	viperInstance.Set(UUIDKey, uuidLibrary.Generate(configPath))
```
olli-holmala, line 57: Since we are hashing the config path, does that not mean that we will have the same UUID for each instance in a cluster of instances, assuming all instances use the default Agent conf path?

internal/plugin/grpc_client_plugin.go
```diff
@@ -122,7 +121,7 @@ func (gc *GrpcClient) createConnection() error {
 		},
 		Agent: &v1.Instance{
 			InstanceMeta: &v1.InstanceMeta{
-				InstanceId:   uuidLibrary.Generate("/etc/nginx-agent/nginx-agent"),
+				InstanceId:   agentGrpc.UUID,
```
olli-holmala, line 124: So `InstanceId` is hardcoded to "uuid"?

internal/config/config.go
olli-holmala: Looks like we don't send this out, so I suppose it doesn't matter ıf there's a collision between instances

oliveromahony: I think it does matter, since we set this to be the instance id eventually. Will update

nginx-nickc: can we make sure this is unique since it looks like it's being use in the auth header

oliveromahony:
UUID generation uniqueness is coming in a follow-up PR. This is to ensure connectivity is successful

olli-holmala (approved):

PR #641, opened 2024-04-18
Retry on RPC errors

oliveromahony:
### Proposed changes

Add retry logic into RPC requests

### Checklist

### DIFF
--- a/internal/plugin/grpc_client_plugin.go
+++ b/internal/plugin/grpc_client_plugin.go
@@ -7,6 +7,7 @@ package plugin
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net"
@@ -82,11 +83,8 @@ func (gc *GrpcClient) Init(ctx context.Context, messagePipe bus.MessagePipeInter
 		return err
 	}
 
-	gc.connectionMutex.Lock()
-	defer gc.connectionMutex.Unlock()
-
 	if gc.conn == nil || gc.conn.GetState() == connectivity.Shutdown {
-		return fmt.Errorf("can't connect to server")
+		return errors.New("can't connect to server")
 	}
 
 	gc.commandServiceClient = v1.NewCommandServiceClient(gc.conn)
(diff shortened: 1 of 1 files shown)


internal/backoff/backoff.go
```diff
@@ -60,7 +58,32 @@ func WaitUntil(
...
-	backoffWithContext := backoff.WithContext(eb, ctx)
+	backoffWithContext := Context(ctx, backoffSettings)
 
 	return backoff.Retry(operation, backoffWithContext)
 }
+
+// WaitUntilWithData Implementation of backoff operations that increases the back off period for each retry
+// attempt using a randomization function that grows exponentially. This does not allow for parameters.
+// nolint: ireturn
+func WaitUntilWithData[T any](
```
dhurley, line 69: Is this function is being used anywhere?

oliveromahony: not yet but I added it in because every time I go to look at WaitUntil I think I have to write a function like this

internal/plugin/grpc_client_plugin.go
```diff
@@ -221,12 +221,30 @@ func (gc *GrpcClient) sendDataPlaneStatusUpdate(
...
-	slog.DebugContext(ctx, "Sending data plane status update request", "request", request)
-	if gc.commandServiceClient == nil {
-		return fmt.Errorf("command service client is not initialized")
+	backOffCtx, backoffCancel := context.WithTimeout(ctx, gc.config.Client.Timeout)
+	defer backoffCancel()
+
+	sendDataPlaneStatus := func() (*v1.UpdateDataPlaneStatusResponse, error) {
+		slog.DebugContext(ctx, "Sending data plane status update request", "request", request)
+		if gc.commandServiceClient == nil {
+			return nil, fmt.Errorf("command service client is not initialized")
```
olli-holmala, line 230: nit: https://github.com/uber-go/guide/blob/master/style.md#error-types
```suggestion
			return nil, errors.New("command service client is not initialized")
```
Technically we should also be exporting the error per the guide.

oliveromahony: We probably need a series of tasks to adhere to the style guide

oliveromahony: I made the change from fmt.Errorf to errors.New but didn't externalize the errors

dhurley (approved):

PR #648, opened 2024-04-24
Add LoadableModules to NGINX Info

aphralG:
### Proposed changes

- Add LoadableModules to NGINX Info
- Added Unit Tests 
- Made Changes to test protos `GetNginxPlusInstance` so Loadable Modules can be set for unit tests

### Checklist


internal/service/instance/nginx_instance_service.go
```diff
@@ -271,3 +280,34 @@ func isFlag(vals []string) bool {
 func isKeyValueFlag(vals []string) bool {
 	return len(vals) == 2
 }
+
+func getLoadableModules(nginxInfo *Info) (modules []string, err error) {
```
dhurley, line 284: probably could just log the error instead of returning it since you aren't using the error for anything in the caller function.
Would be consistent then with the other functions in this file like https://github.com/nginx/agent/blob/0078c98ec0c18d29533e5dde1d9d89d3ffecd97b/internal/service/instance/nginx_instance_service.go#L243
and 
https://github.com/nginx/agent/blob/0078c98ec0c18d29533e5dde1d9d89d3ffecd97b/internal/service/instance/nginx_instance_service.go#L259

test/protos/instances.go
```diff
@@ -36,15 +36,15 @@ func GetNginxOssInstance() *v1.Instance {
...
+					LoadableModules: nil,
 					DynamicModules:  []string{},
 				},
 			},
 		},
 	}
 }
 
-func GetNginxPlusInstance() *v1.Instance {
+func GetNginxPlusInstance(loadableModules string) *v1.Instance {
```
dhurley, line 47: Should the `GetNginxOssInstance` function also be updated to take a loadableModules param as well to be consistent?

dhurley, line 47: Should the `loadableModules` param be an array of strings?

internal/service/instance/nginx_instance_service_test.go
```diff
@@ -42,7 +46,7 @@ const (
 		"--with-http_v2_module --with-ipv6 --with-mail --with-mail_ssl_module --with-pcre " +
 		"--with-pcre-jit --with-stream --with-stream_realip_module --with-stream_ssl_module " +
 		"--with-stream_ssl_preread_module"
-	plusConfigArgs = "--prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules " +
+	plusConfigArgs = "--prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=%s " +
```
dhurley, line 49: Should the `ossConfigArgs` be updated as well to make sure the change works for both plus and oss?

test/protos/instances.go
aphralG: I was thinking more about this and as we add more information I'll need it to take more params so I'm not sure if this is the best way to do this or should I be passing in the entire runtime info

internal/service/instance/nginx_instance_service_test.go
aphralG: We didn't have a module path in the `ossConfigArgs` and thought there might be a reason it was missing from the test so didn't change it but I can add it

test/protos/instances.go
aphralG: Changed GetNginxOssInstance to take loadableModules param can see what to do about more params later

craigell (approved):

internal/service/instance/nginx_instance_service.go
```diff
@@ -111,6 +113,11 @@ func (n *Nginx) getInfo(ctx context.Context, nginxProcess *model.Process) (*Info
 	nginxInfo.ExePath = exePath
 	nginxInfo.ProcessID = nginxProcess.Pid
 
+	loadableModules := getLoadableModules(nginxInfo)
+	nginxInfo.LoadableModules = loadableModules
+	slog.Info("", "", loadableModules)
```
olli-holmala, line 118: Meant to be deleted?

internal/service/instance/nginx_instance_service.go
```diff
@@ -271,3 +278,36 @@ func isFlag(vals []string) bool {
...
+	var err error
+	if nginxInfo.ConfigureArgs["modules-path"] != nil {
+		modulePath, ok := nginxInfo.ConfigureArgs["modules-path"].(string)
+		if !ok {
+			slog.Warn("error parsing modules-path")
+			return modules
+		}
+		modules, err = readDirectory(modulePath, ".so")
+		if err != nil {
+			slog.Warn("error reading module dir", "dir", modulePath, "error", err)
```
olli-holmala, line 292:
```suggestion
			slog.Warn("Error reading module dir", "dir", modulePath, "error", err)
```

internal/service/instance/nginx_instance_service.go
```diff
@@ -271,3 +278,36 @@ func isFlag(vals []string) bool {
...
 	return len(vals) == 2
 }
+
+func getLoadableModules(nginxInfo *Info) (modules []string) {
+	var err error
+	if nginxInfo.ConfigureArgs["modules-path"] != nil {
+		modulePath, ok := nginxInfo.ConfigureArgs["modules-path"].(string)
+		if !ok {
+			slog.Warn("error parsing modules-path")
+			return modules
```
olli-holmala, line 288:
```suggestion
	if mp, ok := nginxInfo.ConfigureArgs["modules-path"]; ok {
		modulePath, ok := mp.(string)
		if !ok {
			slog.Warn("Error parsing modules-path")
			return modules
```

internal/service/instance/nginx_instance_service.go
```diff
@@ -271,3 +278,36 @@ func isFlag(vals []string) bool {
...
+		return modules
+	}
+
+	return modules
+}
+
+func readDirectory(dir, ext string) (files []string, err error) {
+	dirInfo, err := os.ReadDir(dir)
+	if err != nil {
+		return files, fmt.Errorf("unable to read directory %s, %w", dir, err)
```
olli-holmala, line 305: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping
> When adding context to returned errors, keep the context succinct by avoiding phrases like "failed to", which state the obvious and pile up as the error percolates up through the stack:
```suggestion
		return files, fmt.Errorf("read directory %s, %w", dir, err)
```

internal/service/instance/nginx_instance_service.go
```diff
@@ -271,3 +278,36 @@ func isFlag(vals []string) bool {
...
+			return modules
+		}
+
+		return modules
+	}
+
+	return modules
+}
+
+func readDirectory(dir, ext string) (files []string, err error) {
```
olli-holmala, line 302: Perhaps add a comment explaining what the `ext` is supposed to do, as it's not obvious based on the function's name.

internal/service/instance/nginx_instance_service_test.go
```diff
@@ -70,6 +75,20 @@ const (
...
+
+	helpers.CreateDirWithErrorCheck(t, modulePath)
+	defer helpers.RemoveFileWithErrorCheck(t, modulePath)
+
+	testModule := helpers.CreateFileWithErrorCheck(t, modulePath, "test.so")
+	defer helpers.RemoveFileWithErrorCheck(t, testModule.Name())
+
+	plusArgs := fmt.Sprintf(plusConfigArgs, modulePath)
+	ossArgs := fmt.Sprintf(ossConfigArgs, modulePath)
+	noModuleArgs := fmt.Sprintf(ossConfigArgs, t.TempDir()+"\"/usr/lib/nginx/modules\"")
```
Dean-Coakley, line 89: nit/optional:

This is a good place to use `path.Join` so escaping slashes is no longer required. (Extra import required too)
```suggestion
	noModuleArgs := path.Join(ossConfigArgs, t.TempDir(), "/usr/lib/nginx/modules")
```

Dean-Coakley (approved):

olli-holmala (approved):

PR #660, opened 2024-05-01
Change GetProcesses to only return nginx processes & fix get Instances

aphralG:
### Proposed changes

Previously anytime the any process changed it would trigger a getInstances which would cause a Data plane Status Update to be sent, this was every minute or so. This was a bug and has now changed so that getInstances only happens when the processID has changed or there is a new nginx process. 

- Change GetProcesses to only return nginx processes and not all processes (because of this `isNginxProcess` & `getNginxProcesses` have moved to process.go) 
- The resource plugin keeps a map of nginxConfig Contexts with the accessLogs, errorLogs and Stub Status API and Plus API for each instance so that when the resource instances are updated we don't need to parse the config again

### Checklist


internal/datasource/host/process.go
```diff
@@ -8,12 +8,13 @@ package host
...
 	"context"
 	"log/slog"
+	"strings"
 
 	"github.com/nginx/agent/v3/internal/model"
 	"github.com/shirou/gopsutil/v3/process"
 )
 
-func GetProcesses(ctx context.Context) ([]*model.Process, error) {
+func GetProcesses(ctx context.Context) (map[int32]*model.Process, error) {
```
olli-holmala, line 17: I would propose we either add a comment here explaining what the key of our map is (i.e. the PID) or then we add a clarifying type alias for the maps key:
```go
type PID int32 // alternatively could call it 'ProcessID'

func GetProcesses(ctx context.Context) (map[PID]*model.Process, error) 
```

olli-holmala: Also, do we want to rename this to `GetNginxProcesses` now?

internal/plugin/config_plugin.go
```diff
@@ -75,6 +77,18 @@ func (c *Config) Process(ctx context.Context, msg *bus.Message) {
 			c.resource = resource
 			c.resourceMutex.Unlock()
 		}
+
+		once.Do(func() {
+			// This will be replaced when we implement config upload
+			instanceList := c.resource.GetInstances()
+			if len(instanceList) > 1 {
```
olli-holmala, line 84: I assume this is because we treat the Agent itself as an instance, so the second element is an NGINX instance?

internal/datasource/host/process.go
aphralG: To the first comment, yes I'll add that I meant to and it slipped my mind.
Second comment, I can ya I left it as is so we could in future add checks for other processes but can always be changed again later when we add those in.

internal/plugin/config_plugin.go
aphralG: Yes, need to check there is more than one as the Agent is an instance in this case

internal/datasource/host/process.go
olli-holmala: TBH I think we should use the type alias, as we are passing this map thoughout our code base and I think it significantly improves readability.

aphralG: Agreed, I'll add an alias

internal/plugin/resource_plugin_test.go
```diff
@@ -49,37 +50,79 @@ func TestResourceMonitor_Subscriptions(t *testing.T) {
...
+		{
+			name:             "Test 2: Instance Config Context Topic",
+			processesMessage: &bus.Message{Topic: bus.InstanceConfigContextTopic, Data: nginxConfigContext},
+			resource:         protos.GetHostResource(),
+			topic:            bus.ResourceTopic,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
olli-holmala, line 100: Technically a shadow declaration of `t`.
```suggestion
		t.Run(test.name, func(tt *testing.T) {
```
Everything in the function should then use `tt` instead of `t`.

olli-holmala: Ditto this in other tests.

internal/datasource/host/process.go
olli-holmala: > Second comment, I can ya I left it as is so we could in future add checks for other processes but can always be changed again later when we add those in.

Makes sense 👍

internal/datasource/host/process_benchmark_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+package host
+
+import (
+	"testing"
+
+	"github.com/nginx/agent/v3/internal/model"
+)
+
+func BenchmarkNginxService_getNginxProcesses(b *testing.B) {
```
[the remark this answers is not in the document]
olli-holmala: Yeah unit tests for that function makes more sense 👍

olli-holmala (approved):

craigell (approved):

dhurley (approved):

Dean-Coakley (approved):

PR #661, opened 2024-05-02
Fix NGINX Plus Instance, NGINX Version

aphralG:
### Proposed changes

The version sent for a NGINX Plus Instance currently in the instance meta looks like `version: nginx-plus-r30-p2` when it should also contain the NGINX version `1.25.3 (nginx-plus-r31-p1)`

### Checklist

### DIFF
--- a/internal/service/instance/nginx_instance_service.go
+++ b/internal/service/instance/nginx_instance_service.go
@@ -31,7 +31,6 @@ var (
 type Info struct {
 	ProcessID       int32
 	Version         string
-	PlusVersion     string
 	Prefix          string
 	ConfPath        string
 	ConfigureArgs   map[string]interface{}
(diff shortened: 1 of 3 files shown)


internal/service/instance/nginx_instance_service.go
```diff
@@ -202,29 +201,15 @@ func parseNginxVersionCommandOutput(ctx context.Context, output *bytes.Buffer) *
 	return nginxInfo
 }
 
-func parseNginxVersion(line string) (version, plusVersion string) {
-	ossVersionMatches := ossVersionRegex.FindStringSubmatch(line)
-	plusVersionMatches := plusVersionRegex.FindStringSubmatch(line)
+func parseNginxVersion(line string) string {
+	ossVersionMatches := ossVersionRegex.FindString(line)
```
dhurley, line 205: I think we can actually simplify this further by using just one regex for both oss and plus. Something like this will work for both I think:
```
(?P<name>\S+)\/(?P<version>.*)
```

olli-holmala (approved):

Dean-Coakley (approved):

dhurley (approved):

craigell (approved):

PR #671, opened 2024-05-09
Add additional checks in chunking functionality

dhurley:
### Proposed changes

Add additional checks in chunking functionality

### Checklist

### DIFF
--- a/sdk/checksum/checksum.go
+++ b/sdk/checksum/checksum.go
@@ -10,6 +10,9 @@ package checksum
 import (
 	"crypto/sha256"
 	"fmt"
+	"math"
+
+	log "github.com/sirupsen/logrus"
 )
 
 // Checksum - calculate checksum from []byte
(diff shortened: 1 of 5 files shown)


sdk/checksum/checksum.go
```diff
@@ -26,13 +26,26 @@ func HexChecksum(b []byte) string {
...
+
+	if bufSize == 0 {
+		return [][]byte{}
+	}
+
+	if bufSize <= lim {
+		return [][]byte{buf}
+	}
+
+	chunks := make([][]byte, 0)
```
olli-holmala, line 39: Why do we no longer set the capacity for `chunks`?

craigell (approved):

sdk/checksum/checksum.go
```diff
@@ -36,8 +39,19 @@ func Chunk(buf []byte, lim int) [][]byte {
...
+		log.Error("Unable to chuck payload, chunk size is too small")
+		return [][]byte{}
+	}
+
+	chuckSize := bufSize / lim
+
+	if chuckSize > math.MaxInt64-1 {
+		log.Error("Unable to chuck payload, data too large")
+		return [][]byte{}
+	}
```
Dean-Coakley, line 52: nit/opt:
`chuck` -> `chunk`
`chuckSize` -> `chunkSize`
?

Dean-Coakley (approved):

sdk/checksum/checksum.go
olli-holmala: Donal answered here at the office, so the reason was that the tool was complaining about our capacity potentially overflowing.

FWIW, I did the math on this one. The capacity was `len(buf)/lim+1` meaning the only way we exceed the maximum length of an integer (i.e. overflow) is if the buffer (=`buf`) had 2^63 elements in it (so 9,223,372,036,854,775,807 elements) on a 64-bit system. Since `buf` is a byte slice, that means you would have 2^63 bytes, which translates to ~9,223 **peta**bytes, meaning you definitely would have run into problem far earlier than when entering this.

PR #675, opened 2024-05-16
Add watcher plugin and instance watcher service

dhurley:
### Proposed changes

Add watcher plugin and instance watcher service

### Checklist


internal/bus/fake_message_pipe.go
```diff
@@ -90,6 +103,9 @@ func (p *FakeMessagePipe) Run(ctx context.Context) {
 func (p *FakeMessagePipe) RunWithoutInit(ctx context.Context) {
 	var message *Message
 
+	// p.messagesLock.Lock()
```
aphralG, line 106: assume this was left in or commented out by mistake

dhurley: updated

internal/watcher/watcher_plugin.go
```diff
@@ -0,0 +1,93 @@
...
+func (w *Watcher) Init(ctx context.Context, messagePipe bus.MessagePipeInterface) error {
+	slog.DebugContext(ctx, "Starting watcher plugin")
+	w.messagePipe = messagePipe
+
+	watcherContext, cancel := context.WithCancel(ctx)
+	w.cancel = cancel
+
+	go w.instanceWatcherService.Watch(watcherContext, w.instanceUpdatesChannel)
+	go w.monitorInstanceUpdates(watcherContext)
+
```
olli-holmala, line 45: Hmm, this is explicitly called out as a thing **not** to do in the Uber Style guide:
https://github.com/uber-go/guide/blob/master/style.md#no-goroutines-in-init
I suppose our hands are a little tied when it comes to the interface we have to implement.

Which reminds me, we should also verify interface compliance here too (verify that this is a Plugin):
https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

internal/watcher/watcher_plugin.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+func (*Watcher) Process(_ context.Context, _ *bus.Message) {}
+
+func (*Watcher) Subscriptions() []string {
+	return []string{}
+}
+
+func (w *Watcher) monitorInstanceUpdates(ctx context.Context) {
```
olli-holmala, line 71: We may want this to be more generic so that we listen to updates from all watchers? Or do we want a unique consumer goroutine for each watcher?

internal/watcher/process_watcher.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"context"
+
+	"github.com/nginx/agent/v3/internal/model"
+	"github.com/shirou/gopsutil/v3/process"
+)
+
+type (
+	ProcessWatcher struct{}
```
olli-holmala, line 16: Could we maybe call this an operator (it's stateless and is called by the instance watcher)? Don't want to muddy the water on the difference between an operator and a watcher this soon in the refactor 😄

internal/watcher/watcher_plugin.go
dhurley: The channels will be of different types depending on the watcher. The instance watcher will return instance updates, file watcher will return file updates, etc. So I think its better to have a separate channel for each watcher unless we want to have a generic object for storing updates from watchers.

internal/watcher/watcher_plugin.go
dhurley: This `Init` function here is different from the built in `init` function in golang so it is ok.
I'll add the interface compliance.

internal/watcher/process_watcher.go
dhurley: yeah good point. Let me update that.

internal/watcher/watcher_plugin.go
olli-holmala: But could we not listen to events from multiple different channels in the `select`? Regardless, maybe this can be added in later if we really want to.

internal/watcher/watcher_plugin.go
olli-holmala: Agree about the Init thing, think it's fine for us here 👌

internal/watcher/watcher_plugin.go
dhurley: Oh yeah that makes sense. I think thats a good idea actually let me update that.

internal/watcher/instance_watcher_service.go
```diff
@@ -0,0 +1,133 @@
...
+func (iw *InstanceWatcherService) updates(ctx context.Context) (
+	instanceUpdates InstanceUpdates,
+	err error,
+) {
+	processes, err := iw.processOperator.Processes(ctx)
+	if err != nil {
+		return instanceUpdates, err
+	}
+
+	instancesFound := []*v1.Instance{}
```
olli-holmala, line 91: Is there a way where we could save the PID and/or the PPID of NGINX processes in a map and then check against those before we parse into instances? In other words, we wouldn't perform the Process -> Instance conversion unless there are new instances and we could potentially remove instances without even needing to parse into instances.

You could also only parse the processes that are new instead of the whole processes list.

All of this may just be premature optimization though.

internal/watcher/instance_watcher_service.go
```diff
@@ -0,0 +1,133 @@
...
+		case <-ctx.Done():
+			close(ch)
+			return
+		case <-instanceWatcherTicker.C:
+			newCtx := context.WithValue(ctx, logger.CorrelationIDContextKey, logger.GenerateCorrelationID())
+			instanceUpdates, err := iw.updates(newCtx)
+			if err != nil {
+				slog.ErrorContext(newCtx, "Instance watcher updates", "error", err)
+			} else {
+				ch <- instanceUpdates
```
olli-holmala, line 76: Maybe we shouldn't send out updates to the channel if there is nothing to update?

internal/watcher/nginx_process_parser.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func (npp *NginxProcessParser) Parse(ctx context.Context, processes []*model.Process) []*v1.Instance {
+	processList := []*v1.Instance{}
+
+	nginxProcesses := filterNginxProcesses(processes)
+
+	for _, nginxProcess := range nginxProcesses {
+		_, ok := nginxProcesses[nginxProcess.Ppid]
+		if !ok {
```
olli-holmala, line 66: nit: maybe add a comment to explain what's happening here (finding the master NGINX process I believe?).

internal/watcher/nginx_process_parser.go
```diff
@@ -0,0 +1,340 @@
...
+)
+
+func NewNginxProcessParser() *NginxProcessParser {
+	return &NginxProcessParser{
+		executer: &exec.Exec{},
+	}
+}
+
+func (npp *NginxProcessParser) Parse(ctx context.Context, processes []*model.Process) []*v1.Instance {
+	processList := []*v1.Instance{}
```
olli-holmala, line 60: nit:
```suggestion
	instanceList := []*v1.Instance{}
```

internal/watcher/nginx_process_parser.go
```diff
@@ -0,0 +1,340 @@
...
+func sanitizeExeDeletedPath(exe string) string {
+	firstSpace := strings.Index(exe, "(deleted)")
+	if firstSpace != -1 {
+		return strings.TrimSpace(exe[0:firstSpace])
+	}
+
+	return strings.TrimSpace(exe)
+}
+
+func convertInfoToProcess(nginxInfo Info) *v1.Instance {
```
olli-holmala, line 148: nit: it's confusing that we call the outcome a "process" when the type is "instance"
```suggestion
func convertInfoToInstance(nginxInfo Info) *v1.Instance {
```

internal/watcher/instance_watcher_service.go
dhurley: Its not going to help much I'd say. At this stage we dont know which processes are NGINX or not. So we need to call `Parse` to filter out the NGINX processes and get the process info. From what I have seen before the parsing of the NGINX process is fast. The slow piece is getting the full process list.

internal/watcher/instance_watcher_service.go
dhurley: yes thats a good point let me update that.

internal/watcher/instance_watcher_service.go
olli-holmala: Fair enough, let's leave it as is 👍

internal/watcher/watcher_plugin.go
```diff
@@ -0,0 +1,96 @@
...
+	Watcher struct {
+		messagePipe            bus.MessagePipeInterface
+		agentConfig            *config.Config
+		instanceWatcherService *InstanceWatcherService
+		instanceUpdatesChannel chan InstanceUpdatesMessage
+		cancel                 context.CancelFunc
+	}
+)
+
+func NewWatcher(agentConfig *config.Config) *Watcher {
```
olli-holmala, line 27:
```suggestion
)

var _ bus.Plugin = (*Watcher)(nil)

func NewWatcher(agentConfig *config.Config) *Watcher {
```

olli-holmala (approved):

aphralG (approved):

oliveromahony (approved):

PR #685, opened 2024-05-23
Parse NGINX config and update NGINX instances with config context

dhurley:
### Proposed changes

- Parse NGINX config and update NGINX instances with config context
- Add a NginxConfigContext topic for sending NGINX config updates to other plugins

### Checklist


internal/watcher/instance_watcher_service.go
```diff
@@ -187,3 +231,61 @@ func compareInstances(oldInstances, instances []*v1.Instance) (newInstances, del
 
 	return newInstances, deletedInstances
 }
+
+func (iw *InstanceWatcherService) updateNginxInstanceRuntime(
```
aphralG, line 235: Should unit tests be added for these ? I think there was one in the resource plugin

internal/config/types.go
```diff
@@ -125,3 +128,16 @@ type InstanceWatcher struct {
...
+
+	for _, allowedDirectory := range c.AllowedDirectories {
+		if strings.HasPrefix(directory, allowedDirectory) {
+			isDirectoryAllowed = true
+			break
+		}
+	}
+
+	return isDirectoryAllowed
+}
```
olli-holmala, line 143: style: the variable and `break` aren't necessary imo.
```suggestion
func (c *Config) IsDirectoryAllowed(directory string) bool {
	for _, allowedDirectory := range c.AllowedDirectories {
		if strings.HasPrefix(directory, allowedDirectory) {
			return true
		}
	}

	return false
}
```

olli-holmala: This implementation assumes that `directory` and `allowedDirectory` are both absolute paths. Will that always be the case?

internal/watcher/instance_watcher_service.go
```diff
@@ -35,12 +38,18 @@ type (
...
-		processOperator processOperator
-		processParsers  []processParser
-		cache           []*v1.Instance
-		executer        exec.ExecInterface
+		agentConfig       *config.Config
+		processOperator   processOperator
+		processParsers    []processParser
+		nginxConfigParser nginxConfigParser
+		instanceCache     []*v1.Instance
+		nginxConfigCache  map[string]*model.NginxConfigContext
```
olli-holmala, line 51: nit:
```suggestion
		nginxConfigCache  map[string]*model.NginxConfigContext // key is instance ID
```

olli-holmala (approved):
Only had minor comments, looks good 👍

internal/config/types.go
dhurley: yes both have to be absolute paths. If they were relative paths I dont think we would have a way of finding out what directory they are in.

PR #686, opened 2024-05-24
Add Instance Children & Worker Processes

aphralG:
### Proposed changes

- Added Instance Children to Instance Runtime 
- Get Worker Processes in `nginx_process_parser.go `
     - If process is worker process and has a parent in the list of NGINX processes it is added as a worker of that instance
      
     - If the process is a worker process but has no parent in the list of NGINX processes a instance is created but with a pid of 0 indicating there is no master process 
      

### Checklist


internal/watcher/nginx_process_parser.go
```diff
@@ -56,28 +56,70 @@ func NewNginxProcessParser() *NginxProcessParser {
...
 
 	nginxProcesses := filterNginxProcesses(processes)
 
 	for _, nginxProcess := range nginxProcesses {
-		// Here we are determining if the nginxProcess is a master process.
-		// NGINX worker processes are ignored.
-		_, ok := nginxProcesses[nginxProcess.Ppid]
-		if !ok {
+		// Here we are determining if the nginxProcess is a worker process
+		if masterProcess, ok := nginxProcesses[nginxProcess.Ppid]; ok {
```
olli-holmala, line 70: Dunno if this is a big hassle, but could we perhaps rename the fields in `Process` so that they are capitalized? Tripped up several times here confusing `Pid` and `Ppid`, so think this might help with that slightly.
```go
type Process struct {
	Pid  int32
	Ppid int32
	Name string
	Cmd  string
	Exe  string
}

```
becomes

```go
type Process struct {
	PID  int32
	PPID int32
	Name string
	Cmd  string
	Exe  string
}
```
Initialisms and acronyms should be capitalized per Go conventions: https://google.github.io/styleguide/go/decisions.html#initialisms

I know this isn't part of your PR, so can do it in a follow-up PR.

olli-holmala (approved):

oliveromahony (approved):

dhurley (approved):

PR #694, opened 2024-05-29
Add file plugin

dhurley:
### Proposed changes

- Add file plugin
- Add common grpc connection struct

### Checklist


internal/grpc/grpc.go
```diff
@@ -34,11 +60,75 @@ var (
...
-type wrappedStream struct {
-	grpc.ClientStream
-	*protovalidate.Validator
+func NewGrpcConnection(ctx context.Context, agentConfig *config.Config) (*GrpcConnection, error) {
+	if agentConfig == nil || agentConfig.Command.Server.Type != "grpc" {
+		return nil, fmt.Errorf("invalid command server settings")
+	}
+
+	if agentConfig.Common == nil {
+		return nil, fmt.Errorf("invalid common configuration settings")
```
olli-holmala, line 73: https://github.com/uber-go/guide/blob/master/style.md#errors
```suggestion
		return nil, errors.New("invalid common configuration settings")
```

internal/grpc/grpc.go
```diff
@@ -34,11 +60,75 @@ var (
...
+
+	_ GrpcConnectionInterface = (*GrpcConnection)(nil)
 )
 
-type wrappedStream struct {
-	grpc.ClientStream
-	*protovalidate.Validator
+func NewGrpcConnection(ctx context.Context, agentConfig *config.Config) (*GrpcConnection, error) {
+	if agentConfig == nil || agentConfig.Command.Server.Type != "grpc" {
+		return nil, fmt.Errorf("invalid command server settings")
```
olli-holmala, line 69: https://github.com/uber-go/guide/blob/master/style.md#errors
```suggestion
		return nil, errors.New("invalid command server settings")
```

internal/grpc/grpc.go
```diff
@@ -34,11 +60,75 @@ var (
...
 	defaultCredentials = insecure.NewCredentials()
+
+	_ GrpcConnectionInterface = (*GrpcConnection)(nil)
 )
 
-type wrappedStream struct {
-	grpc.ClientStream
-	*protovalidate.Validator
+func NewGrpcConnection(ctx context.Context, agentConfig *config.Config) (*GrpcConnection, error) {
+	if agentConfig == nil || agentConfig.Command.Server.Type != "grpc" {
```
olli-holmala, line 68: We should probably have an enum for `agentConfig.Command.Server.Type` at some point.

internal/grpc/grpc.go
```diff
@@ -9,23 +9,49 @@ import (
...
+	GrpcConnectionInterface interface {
+		CommandServiceClient() mpi.CommandServiceClient
+		FileServiceClient() mpi.FileServiceClient
+		Close(ctx context.Context) error
+	}
+
+	GrpcConnection struct {
+		config          *config.Config
+		conn            *grpc.ClientConn
+		connectionMutex sync.Mutex
```
olli-holmala, line 46: style: `connectionMutex` is a little verbose if we aren't planning on adding more mutexes to this struct.
https://google.github.io/styleguide/go/decisions.html#external-context-vs-local-names

dhurley: yeah I'll update it there. I probably copy pasted that from somewhere else

internal/grpc/grpc.go
dhurley: good point let me see what I can do

oliveromahony (approved):

internal/file/file_plugin.go
```diff
@@ -0,0 +1,154 @@
...
+func (fp *FilePlugin) Process(ctx context.Context, msg *bus.Message) {
+	switch msg.Topic {
+	case bus.NginxConfigContextTopic:
+		fp.handleNginxConfigUpdate(ctx, msg)
+	case bus.ConfigUploadRequestTopic:
+		fp.handleConfigUploadRequest(ctx, msg)
+	default:
+		slog.DebugContext(ctx, "File plugin unknown topic", "topic", msg.Topic)
+	}
+}
```
olli-holmala, line 67: Don't know if it would be a good idea to add comments here for each `case` on what the event represents. So for example:
```go
  // File updates originating from file watchers.
  case bus.NginxConfigContextTopic:
```
Or then potentially rename the topic, as a NGINX config context topic doesn't really describe what events are being sent or the purpose of the topic?

olli-holmala (approved):

internal/file/file_plugin.go
dhurley: Let me rename the topic to make it clearer

PR #696, opened 2024-05-30
Detect unhealthy instance

aphralG:
### Proposed changes

Added degraded, unhealthy instance health
- If instance has no children, instance health is DEGRADED 
- If process is not running or status is "zombie", instance health is UNHEALTHY
- If instance has been deleted, instance health of UNHEALTHY is sent once 
- If we are unable to determine the health, due to error etc, instance health is UNSPECIFIED
- Other wise instance is healthy

Added Running and Status to Process struct 
Changed Imports of `v1` to `mpi` in files I noticed 
Add health, instance and process folders to watchers 
```
- watcher
     - health 
           health_watcher_service.go
           ...
     - instance
          instance_watcher_service.go 
          ...
     - process 
          process.go
 watcher_plugin.go
 ```

### Checklist

### CODE REVIEW

File: internal/watcher/health/health_watcher_service.go [hunk 1]
```diff
@@ -131,7 +151,25 @@ func (hw *HealthWatcherService) updateCache(currentHealth map[string]*v1.Instanc
...
 
-func (hw *HealthWatcherService) compareHealth(currentHealth map[string]*v1.InstanceHealth) bool {
+func (hw *HealthWatcherService) compareCache(healthStatuses []*mpi.InstanceHealth) []*mpi.InstanceHealth {
+	if len(hw.cache) != len(hw.instances) {
+		for instanceID := range hw.cache {
+			if _, ok := hw.instances[instanceID]; !ok {
+				health := &mpi.InstanceHealth{
+					InstanceId:           instanceID,
+					InstanceHealthStatus: mpi.InstanceHealth_INSTANCE_HEALTH_STATUS_UNHEALTHY,
+					Description:          fmt.Sprintf("instance has been deleted %s", instanceID),
```
dhurley, line 161:
```suggestion
					Description:          fmt.Sprintf("instance %s not found", instanceID),
```

I think not found might be a better way to word it than saying it was deleted

internal/watcher/process/process_operator.go
```diff
@@ -0,0 +1,89 @@
...
+
+	internalProcesses := []*model.Process{}
+
+	for _, proc := range processes {
+		ppid, _ := proc.PpidWithContext(ctx)
+		name, _ := proc.NameWithContext(ctx)
+		cmd, _ := proc.CmdlineWithContext(ctx)
+		exe, _ := proc.ExeWithContext(ctx)
+		status, _ := proc.Status()
+		running, _ := proc.IsRunning()
```
dhurley, line 48:
```suggestion
		status, _ := proc.StatusWithContext(ctx)
		running, _ := proc.IsRunningWithContext(ctx)
```

internal/watcher/process/process_operator.go
```diff
@@ -0,0 +1,89 @@
...
+		})
+	}
+
+	return internalProcesses, nil
+}
+
+func (pw *ProcessOperator) Process(ctx context.Context, pid int32) (*model.Process, error) {
+	newProc, err := process.NewProcessWithContext(ctx, pid)
+	status, _ := newProc.Status()
+	running, _ := newProc.IsRunning()
```
dhurley, line 67:
```suggestion
	status, _ := proc.StatusWithContext(ctx)
	running, _ := proc.IsRunningWithContext(ctx)
```

internal/watcher/health/nginx_health_watcher_operator.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	health := &mpi.InstanceHealth{
+		InstanceId:           instance.GetInstanceMeta().GetInstanceId(),
+		InstanceHealthStatus: mpi.InstanceHealth_INSTANCE_HEALTH_STATUS_HEALTHY,
+	}
+	if strings.Contains(proc.Status, "zombie") || !proc.Running {
+		health.Description = fmt.Sprintf("pid: %d is degraded, status: %s", proc.PID, proc.Status)
```
olli-holmala, line 43:
```suggestion
		health.Description = fmt.Sprintf("PID: %d is degraded, status: %s", proc.PID, proc.Status)
```

internal/watcher/health/health_watcher_service.go
```diff
@@ -98,12 +109,19 @@ func (hw *HealthWatcherService) Watch(ctx context.Context, ch chan<- InstanceHea
...
+	currentHealth := make(map[string]*mpi.InstanceHealth, len(hw.watchers))
 
 	for instanceID, watcher := range hw.watchers {
-		instanceHealth := watcher.Health(ctx, instanceID)
+		instanceHealth, err := watcher.Health(ctx, hw.instances[instanceID])
+		if instanceHealth == nil {
+			instanceHealth = &mpi.InstanceHealth{
+				InstanceId:           instanceID,
+				InstanceHealthStatus: mpi.InstanceHealth_INSTANCE_HEALTH_STATUS_UNSPECIFIED,
+				Description:          fmt.Sprintf("failed to get health for instance %s, error: %v", instanceID, err),
```
olli-holmala, line 122: Tiny nit:
```suggestion
				Description:          fmt.Sprintf("failed to get health for instance %s, error: %s", instanceID, err.Error()),
```

internal/watcher/health/health_watcher_service.go
```diff
@@ -98,12 +109,19 @@ func (hw *HealthWatcherService) Watch(ctx context.Context, ch chan<- InstanceHea
...
+			}
+		}
 		healthStatuses = append(healthStatuses, instanceHealth)
```
olli-holmala, line 126: Could we perhaps not use the `healthStatuses` slice here and init like a new slice for readability purposes.

So this could be something like a `allStatuses` slice and then on line 135 we could have a `updatedStatuses` slice? Just cuz imo it's harder to follow with only the one slice in the current implementation.

internal/watcher/health/health_watcher_service.go
```diff
@@ -15,57 +16,67 @@ import (
...
+		hw.instances[instance.GetInstanceMeta().GetInstanceId()] = instance
+	}
+}
+
+func (hw *HealthWatcherService) DeleteHealthWatcher(instances []*mpi.Instance) {
 	for _, instance := range instances {
 		delete(hw.watchers, instance.GetInstanceMeta().GetInstanceId())
+		delete(hw.instances, instance.GetInstanceMeta().GetInstanceId())
 	}
 }
```
olli-holmala, line 81: Shouldn't we always delete the `cache` entry here for the Instance too? That way we wouldn't need to delete `cache`entries in `updateCache` 🤔

aphralG: Discussed in office, the instance isn't deleted in `DeleteHealthWatcher()` as we need a way to tell if an instance has been deleted in `health()` so we can send a health status for that instance that is unhealthy. This is currently being done by comparing the cache with the instanceList. 
Add comments to updateCache to explain what the function is doing

internal/watcher/health/health_watcher_service.go
aphralG: I think I have made those changes if you could have another look. Not 100% sure I understood the comment correctly. Thank you

olli-holmala (approved):

oliveromahony (approved):

internal/watcher/health/nginx_health_watcher_operator.go
```diff
@@ -7,9 +7,12 @@ package health
...
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
+	process2 "github.com/shirou/gopsutil/v3/process"
+
 	mpi "github.com/nginx/agent/v3/api/grpc/mpi/v1"
 	"github.com/nginx/agent/v3/internal/datasource/host/exec"
 	"github.com/nginx/agent/v3/internal/watcher/process"
```
oliveromahony, line 18: personally, would prefer to rename this import to process_watcher over renaming the gopsutil one

olli-holmala: This is also a clear downside to having each watcher in its own package instead of just having a flat `watcher` package. Was there some specific other reason for splitting each into its own directory, other than for cleaner organization purposes?

internal/watcher/instance/instance_watcher_service.go
```diff
@@ -248,14 +250,26 @@ func checkForProcessChanges(
...
-		if oldInstance.GetInstanceRuntime().GetProcessId() != instance.GetInstanceRuntime().GetProcessId() {
+		if compareInstance(oldInstance.GetInstanceRuntime(), instance.GetInstanceRuntime()) {
 			updatedInstances = append(updatedInstances, instance)
 		}
 	}
 
 	return updatedInstances
 }
 
+func compareInstance(oldRuntime, currentRuntime *mpi.InstanceRuntime) (updated bool) {
```
olli-holmala, line 261: nit: `compareInstance` sounds like it's used to compare order of instances in a list. I'd consider renaming this to `areInstancesEqual` or `instancesEqual` or even the Java-style `isEqual`.

You'd have to invert the booleans in the function and add a `!` to the if clause, but I think it would slightly improve readability. Not a big deal if you wanna leave it as is though.

internal/watcher/instance/instance_watcher_service.go
```diff
@@ -248,26 +245,38 @@ func compareInstances(oldInstances, instances []*v1.Instance) (
...
+	if oldRuntime.GetProcessId() != currentRuntime.GetProcessId() {
+		return true
+	}
+
+	if !reflect.DeepEqual(oldRuntime.GetInstanceChildren(), currentRuntime.GetInstanceChildren()) {
+		return true
+	}
+
+	return false
```
olli-holmala, line 270:
```suggestion
	return !reflect.DeepEqual(oldRuntime.GetInstanceChildren(), currentRuntime.GetInstanceChildren())
```

internal/watcher/instance/instance_watcher_service.go
aphralG: Done

dhurley (approved):

PR #698, opened 2024-06-07
feat: create OTel collector config from template

olli-holmala:
### Proposed changes

* Introduces `config.Collector`, which replaces the old `config.Metrics`:
  * `config_path`: Config file path to the OTel Collector config that the embedded collector will generate based on a template.
  * `receivers`:  A list of receivers, along with server and TLS information, that will be generated in the OTel Collector config.
  * `processors`:  A list of processors that will be generated in the OTel Collector config.
  * `exporters`:  A list of exporters, along with server and TLS information, that will be generated in the OTel Collector config.
  * `health`: Server configuration (i.e. host:port) for the health check endpoint of the OTel collector.
* The OTel Collector config YAML is generated by default in `/var/run/nginx-agent-otelcol.yaml` based on the `otelcol.tmpl` Go template file, which consumes the `config.Collector` content and injects it into the template.
* The collector config truncates during write, meaning the file's content is **replaced** upon bootup of the NGINX Agent.

### Upcoming followup work

* Further `extension` and `receiver` support, depending on what we deem appropriate.

### Checklist


internal/config/config_test.go
```diff
@@ -49,6 +49,12 @@ func TestGetConfig(t *testing.T) {
 	assert.Equal(t, 30*time.Second, result.DataPlaneConfig.Nginx.ReloadMonitoringPeriod)
 	assert.False(t, result.DataPlaneConfig.Nginx.TreatWarningsAsError)
 
+	require.NotNil(t, result.Metrics)
+	assert.True(t, result.Metrics.Collector)
+	assert.Equal(t, "localhost:3000", result.Metrics.OTLPExportURL)
+	assert.Equal(t, "localhost:4318", result.Metrics.OTLPReceiverURL)
+	assert.Equal(t, "/var/etc/nginx-agent/nginx-agent-otelcol.yaml", result.Metrics.CollectorConfigPath)
```
oliveromahony, line 56: this path looks unusual, even for a test

internal/config/testdata/nginx-agent.conf
```diff
@@ -15,4 +15,12 @@ data_plane_config:
...
 
+metrics:
+  collector: true
+  otlp_export_url: "localhost:3000"
+  otlp_receiver_url: "localhost:4318"
+  collector_receivers:
+    - hostmetrics
+  collector_config_path: "/var/etc/nginx-agent/nginx-agent-otelcol.yaml"
+
 config_dirs: "/etc/nginx:/usr/local/etc/nginx:/usr/share/nginx/modules:invalid/path"
```
oliveromahony, line 26: maybe we should add ```/var/run/``` to config_dirs and validate that the OTel config is within one of those directories as part of the config load?

internal/config/config_test.go
olli-holmala: Heh, I can change it! Didn't really put a lot of thought into the path considering it's a unit test 😅

internal/config/testdata/nginx-agent.conf
olli-holmala: Great idea, I'll do that!

internal/config/testdata/nginx-agent.conf
```diff
@@ -15,4 +15,12 @@ data_plane_config:
 client: 
   timeout: 10s
 
+metrics:
```
dhurley, line 18: The structure of the metrics config could do with being refactored a bit I think.
I dont think the boolean field `collector` is needed.

Would a structure like this be cleaner:
```
metrics:
  collector:
    otlp_export_url: "localhost:3000"
    otlp_receiver_url: "localhost:4318"
    receivers:
      - hostmetrics
    config_path: "/var/etc/nginx-agent/nginx-agent-otelcol.yaml"
```

Also for the urls do we want to specify if they are `http` or `https`?

internal/config/types.go
```diff
@@ -25,103 +25,116 @@ func parseServerType(str string) (ServerType, bool) {
...
+		PermitWithoutStream bool          `yaml:"-" mapstructure:"permit_without_stream"`
+	}
 
-type GRPC struct {
-	Target         string        `yaml:"-" mapstructure:"target"`
-	ConnTimeout    time.Duration `yaml:"-" mapstructure:"connection_timeout"`
-	MinConnTimeout time.Duration `yaml:"-" mapstructure:"minimum_connection_timeout"`
-	BackoffDelay   time.Duration `yaml:"-" mapstructure:"backoff_delay"`
-}
+	Metrics struct {
```
oliveromahony, line 71: I was thinking metrics would follow a similar structure to Commad?
with an exporter list sub entry in this case. Then Server, Auth  and TLS under it? 

Perhaps something like:
```
metrics:
    collector: true,
    exporters:
        - otlp (default)
          server:
            ...
          auth: 
           ...
          tls: 
          ....
``` 
or alternatively?

internal/metrics/collector/settings.go
```diff
@@ -62,15 +68,47 @@ func createConverterFactories() []confmap.ConverterFactory {
...
+	if err != nil {
+		return err
+	}
+
+	err = otelcolTemplate.Execute(file, conf)
+	if err != nil {
+		return err
+	}
+
+	err = file.Close()
```
dhurley, line 108: should we use `defer` when closing the file? In the case where `otelcolTemplate.Execute(file, conf)` fails we won't close the file

internal/metrics/collector/otelcol.tmpl
```diff
@@ -0,0 +1,55 @@
...
+    timeout: 10s
+    retry_on_failure:
+      enabled: true
+      initial_interval: 10s
+      max_interval: 60s
+      max_elapsed_time: 10m
+
+extensions:
+  health_check:
+    endpoint: "localhost:13133"
```
dhurley, line 40: Does this health_check endpoint need to be parameterized?

internal/metrics/collector/settings.go
```diff
@@ -62,15 +68,47 @@ func createConverterFactories() []confmap.ConverterFactory {
...
-	if !ex {
-		return "/tmp/otel-collector-config.yaml"
+// Generates a OTel Collector config to a file by injecting the Metrics Config to a Go template.
+func writeCollectorConfig(conf *config.Metrics) error {
+	otelcolTemplate, err := template.New(otelTemplatePath).ParseFiles(otelTemplatePath)
+	if err != nil {
+		return err
+	}
+
+	// Should check that we are within an allowed directory in the future.
```
dhurley, line 81: Can you make sure to add a ticket to capture this future work in case it is forgotten about

internal/config/types.go
olli-holmala: We can do that, sure. Let me see what I can do.

internal/config/testdata/nginx-agent.conf
olli-holmala: Hmm yeah, I'm unsure whether the `collector` namespace even necessary, as we will only support metrics via the collector going forward (I believe). Regardless, will try to clean up the config structure a bit (also working off @oliveromahony's [comment](https://github.com/nginx/agent/pull/698#discussion_r1631251782)).

internal/metrics/collector/settings.go
olli-holmala: Good point! The reason why I didn't use `defer` here is that the static security code scan tool we use will complain about not checking `file.Close()` errors. I guess we can check it in the `defer` but it then cannot be returned.

Perhaps if I log the error in the event that it happens in the `defer`, then that will work as a middle ground?

olli-holmala: Added the following `defer`:
```go
defer func() {
  err = file.Close()
  if err != nil {
    slog.Warn("Failed to close file", "file_path", confPath)
  }
}()
```

internal/metrics/collector/otelcol.tmpl
olli-holmala: Probably would be a good idea, will add a config for it 👍

internal/config/types.go
olli-holmala: Okay the structure of `Collector` now follows the suggested format 👍 took a minute to get the `viper.Unmarshal` magic to work for unmarshalling full `mapstructure` structs.

internal/metrics/collector/settings.go
olli-holmala: This is now checked when validating the `Collector` config. Moving forward we should validate all paths in our `Config` in a similar fashion.

I will add a ticket for this.

internal/metrics/collector/otelcol.tmpl
olli-holmala: This is now configurable in the `Collector.HealthzEndpoint` config.

internal/config/testdata/nginx-agent.conf
olli-holmala: As discussed yesterday, `Metrics` has now been renamed to `Collector` and we are parsing a set of `Exporter`s and `Receiver`s.

internal/collector/settings.go
```diff
@@ -19,6 +21,11 @@ import (
 	"go.opentelemetry.io/collector/otelcol"
 )
 
+const (
+	otelTemplatePath     = "otelcol.tmpl"
```
oliveromahony, line 25: just wondering if we should use go:embed  for this instead? 
If it is done this way, I think it would need to get added to the .nfpm config

internal/config/config.go
```diff
@@ -65,32 +66,45 @@ func RegisterConfigFile() error {
...
-		Log:                getLog(),
-		DataPlaneConfig:    getDataPlaneConfig(),
-		Client:             getClient(),
-		ConfigDir:          getConfigDir(),
-		AllowedDirectories: []string{},
-		Metrics:            getMetrics(),
-		Command:            getCommand(),
-		Common:             getCommon(),
-		Watchers:           getWatchers(),
+		Log:                resolveLog(),
```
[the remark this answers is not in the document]
oliveromahony: I see what's going on now. I think having the collector have it's own function resolveCollector is better for consistency. It also simplifies the logic in this function. Splitting the resolves into their own functions seems like a cleaner and more testable approach

internal/config/config.go
```diff
@@ -248,47 +267,71 @@ func normalizeFunc(f *flag.FlagSet, name string) flag.NormalizedName {
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal collector config: %w", err)
+	}
+
+	col := &Collector{
+		ConfigPath:      viperInstance.GetString(CollectorConfigPathKey),
+		Exporters:       exporters,
+		Receivers:       receivers,
+		HealthzEndpoint: &healthCheck,
```
oliveromahony, line 323: I'd prefer if we had a list of processors here... But I don't mind leaving this as a top level item if we always have it. I do mind the naming :) Perhaps it should be called Health?

olli-holmala: Health works, it's a quick rename

internal/collector/settings.go
olli-holmala: Yeah definitely need the embed, gonna add that after done with demoing.

internal/config/config.go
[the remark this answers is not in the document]
olli-holmala: Yeah I switched from Getters to use `Resolve*` naming just to get rid of the dreaded `Get` keyword 😄 The reason why `resolveCollector()` in particular is separate from the others here is that it now returns an `error` because it is now validated, which I think should be standard practice going forward.

Configs should be validated already at this stage so that we can return a failure early and prompt the user to fix a misconfigured Agent.

internal/config/config.go
olli-holmala: Renamed to `Health`  and added Processor support (only "batch" currently supported).

internal/collector/settings.go
olli-holmala: Done now ✅

internal/metrics/collector/settings.go
olli-holmala: Ticket added ✅

internal/collector/otel_collector_plugin_test.go
```diff
@@ -0,0 +1,80 @@
...
+	require.NoError(t, collector.Close(ctx), "Close should not return an error")
+	select {
+	case <-collector.appDone:
+		t.Log("Success")
+	case <-time.After(10 * time.Second):
+		t.Error("Timed out waiting for app to shut down")
+	}
+}
+
+func setupOTelConfig(t *testing.T) *config.Config {
```
oliveromahony, line 48: could move this to ```test/types/config.go``` for reuse outside of this? Might be useful for integration tests for example

oliveromahony: alternatively if you want to keep it as a helper it could go under ```test/helpers/otel_config_utils.go``` or something similar

internal/collector/settings_test.go
```diff
@@ -0,0 +1,103 @@
...
+	assert.NotEmpty(t, settings.ResolverSettings.URIs, "URIs should not be empty")
+	assert.Equal(t, "/etc/nginx-agent/nginx-agent-otelcol.yaml", settings.ResolverSettings.URIs[0],
+		"Default URI should match")
+}
+
+func TestTemplateWrite(t *testing.T) {
+	cfg := types.GetAgentConfig()
+	actualConfPath := filepath.Join(t.TempDir(), "nginx-agent-otelcol-test.yaml")
+	cfg.Collector.ConfigPath = actualConfPath
+	// cfg.Collector.ConfigPath = "testdata/test-otelcol.yaml"
```
oliveromahony, line 47: can you remove this line?

internal/collector/settings_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	cfg.Collector.Exporters = append(cfg.Collector.Exporters, config.Exporter{
```
oliveromahony, line 49: if you do https://github.com/nginx/agent/pull/698/files#r1636799918 or the next comment you can call GetConfiig().Exporters then and reuse the centralised config here and in the below lines

aphralG (approved):

internal/collector/testdata/test-otelcol.yaml
```diff
@@ -0,0 +1,75 @@
+receivers:
```
oliveromahony, line 1: we store configs like these under `test/config/*` historically

internal/config/config_test.go
```diff
@@ -122,30 +139,107 @@ func TestNormalizeFunc(t *testing.T) {
...
+			shouldErr: true,
+			errMsg:    "collector path /path/to/secret not allowed",
+		},
+		{
+			name: "Test 3: Unsupported Exporter",
+			expected: &Collector{
+				ConfigPath: testDefault.Collector.ConfigPath,
+				Exporters: []Exporter{
+					{
+						Type: "envoy",
```
oliveromahony, line 186: maybe change this to nginx or a different exporter :)

internal/config/config_test.go
```diff
@@ -122,30 +139,107 @@ func TestNormalizeFunc(t *testing.T) {
...
+			errMsg:    "unsupported processor type: custom-processor",
+		},
+		{
+			name: "Test 5: Unsupported Receiver",
+			expected: &Collector{
+				ConfigPath: testDefault.Collector.ConfigPath,
+				Exporters:  testDefault.Collector.Exporters,
+				Receivers: []Receiver{
+					{
+						Type: "haproxy",
```
oliveromahony, line 214: maybe change this to nginx or a different receiver :)

oliveromahony (approved):

internal/config/types.go
```diff
@@ -25,107 +46,177 @@ func parseServerType(str string) (ServerType, bool) {
...
+	return nil
 }
 
 func (c *Config) IsDirectoryAllowed(directory string) bool {
-	for _, allowedDirectory := range c.AllowedDirectories {
-		if strings.HasPrefix(directory, allowedDirectory) {
+	return isAllowedDir(directory, c.AllowedDirectories)
+}
+
+func isAllowedDir(dir string, allowedDirs []string) bool {
```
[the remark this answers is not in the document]
oliveromahony: I think it can get reverted? Seems like unnecessary nesting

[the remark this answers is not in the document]
oliveromahony: I see it's used differently on line 175, so maybe not

internal/collector/settings_test.go
olli-holmala: Thought it'd be useful for quickly generating a new test expectation in `testdata/test-otelcol.yaml`. But I'll get rid of it.

internal/config/config_test.go
olli-holmala: Haha the joke is that they aren't allowed 😂

internal/config/types.go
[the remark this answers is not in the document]
olli-holmala: Yup, it's reused in `Collector.Validate()` where the `config.Config` is not fully available yet. We can refactor this later (if we want) in the config validation task that I created today.

internal/collector/otel_collector_plugin_test.go
olli-holmala: Did the first one, second one caused a cyclical dependency.

internal/config/config_test.go
olli-holmala: Renamed to the more serious "not-allowed"

internal/config/config_test.go
olli-holmala: Renamed to the more serious "not-allowed"

internal/collector/settings_test.go
olli-holmala: I'd like to keep this as is, so that the version in `test/types/config.go` has a default configuration, which will be more sensible for integration tests. Here we are going for the "all out" configuration with all possible supported exporters, producers and receivers.

internal/collector/testdata/test-otelcol.yaml
olli-holmala: Moved it there 👍

PR #699, opened 2024-06-07
Add command plugin

dhurley:
### Proposed changes

- Add new command plugin
- Remove grpc client plugin and config plugin
- Fix file permissions regex
- Clean up log statements

### Checklist


internal/command/command_service.go
```diff
@@ -0,0 +1,264 @@
...
+	if !cs.isConnected.Load() {
+		err := cs.createConnection(ctx, resource)
+		if err != nil {
+			return err
+		}
+
+		var subscribeCtx context.Context
+
+		cs.subscribeMutex.Lock()
+		subscribeCtx, cs.subscribeCancel = context.WithCancel(ctx)
```
olli-holmala, line 66: Are we sure we want to pass the parent context to `context.WithCancel` here? I would think that we wouldn't want to close the gRPC connection if the context of a Data Plane Status request cancelled 🤔

I'd have thought we only cancel the `subscribe` goroutine when `CancelSubscription` is called?

internal/command/command_service.go
```diff
@@ -0,0 +1,264 @@
...
+		validatedError := grpc.ValidateGrpcError(updateError)
+
+		if validatedError != nil {
+			slog.ErrorContext(ctx, "Failed to send update data plane health", "error", validatedError)
+
+			return nil, validatedError
+		}
+
+		return response, nil
+	}
```
olli-holmala, line 152: I wonder if we could parameterize these `func`s via generics or something, as we seem to be repeating this same pattern.

internal/command/command_service.go
dhurley: Moved the subscribe function to the creation of the command service instead.
Also added a subscribe channel for the management plane requests.

internal/command/command_service.go
dhurley: I did try to take a look at how we could reduce the similar code but I couldn't see an easy way of doing it.

olli-holmala (approved):

internal/command/command_service.go
```diff
@@ -0,0 +1,286 @@
...
+				InitialInterval:     cs.agentConfig.Common.InitialInterval,
+				MaxInterval:         cs.agentConfig.Common.MaxInterval,
+				MaxElapsedTime:      createConnectionMaxElapsedTime,
+				RandomizationFactor: cs.agentConfig.Common.RandomizationFactor,
+				Multiplier:          cs.agentConfig.Common.Multiplier,
+			}
+
+			retryError := backoff.Retry(subscribeFn, backoffHelpers.Context(ctx, commonSettings))
+			if retryError != nil {
+				slog.ErrorContext(ctx, "Failed to receive messages from subscribe stream", "error", retryError)
```
oliveromahony, line 236: should this be a warning?

internal/command/command_service.go
```diff
@@ -0,0 +1,286 @@
...
+	request := &mpi.CreateConnectionRequest{
+		MessageMeta: &mpi.MessageMeta{
+			MessageId:     uuid.NewString(),
+			CorrelationId: correlationID,
+			Timestamp:     timestamppb.Now(),
+		},
+		Resource: resource,
+	}
+
+	connectFn := func() (*mpi.CreateConnectionResponse, error) {
```
oliveromahony, line 254: I'd say for readability, move this to it's own function and not inline. Can potentially have a unit test then

internal/file/file_manager_service.go
```diff
@@ -73,7 +111,34 @@ func (fms *FileManagerService) UpdateFile(
 		},
 	}
 
-	_, updateFileErr := fms.fileServiceClient.UpdateFile(ctx, request)
+	backOffCtx, backoffCancel := context.WithTimeout(ctx, fms.agentConfig.Common.MaxElapsedTime)
+	defer backoffCancel()
 
-	return updateFileErr
+	sendUpdateFile := func() (*mpi.UpdateFileResponse, error) {
```
oliveromahony, line 117: same comment here, for readability, maybe have it as it's own function at the file level

internal/grpc/grpc.go
```diff
@@ -299,6 +288,20 @@ func ProtoValidatorStreamClientInterceptor() (grpc.StreamClientInterceptor, erro
 	}, nil
 }
 
+func ValidateGrpcError(err error) error {
```
oliveromahony, line 291: is there a unit test for this?

internal/plugin/plugin_manager.go
```diff
@@ -16,18 +20,21 @@ import (
...
-	configPlugin := NewConfig(agentConfig)
-
-	plugins = append(plugins, configPlugin)
-
 	if isGrpcClientConfigured(agentConfig) {
-		grpcClient := NewGrpcClient(agentConfig)
-		plugins = append(plugins, grpcClient)
+		grpcConnection, err := grpc.NewGrpcConnection(ctx, agentConfig)
+		if err != nil {
+			slog.ErrorContext(ctx, "Failed to create gRPC connection", "error", err)
```
oliveromahony, line 31: warning?

internal/service/config/nginx_config_service.go
```diff
@@ -347,6 +347,10 @@ func (n *Nginx) parseAddressesFromServerDirective(parent *crossplane.Directive)
 	foundHosts := []string{}
 	port := "80"
 
+	if parent == nil {
```
oliveromahony, line 350: is there a test scenario to capture this? where the parent is nil?

internal/command/command_service.go
dhurley: cant since it needs variables that are outside of the function like `ctx` & `request`

internal/file/file_manager_service.go
dhurley: same answer as above

internal/grpc/grpc.go
dhurley: I'll add one

internal/command/command_service.go
olli-holmala: You can have a function that returns a function:

```go
func (cs *CommandService) createConnectionFunc(ctx context.Context, request *mpi.CreateConnectionRequest) func() (*mpi.CreateConnectionResponse, error) {
    return func() (*mpi.CreateConnectionResponse, error) {
        response, connectErr := cs.commandServiceClient.CreateConnection(ctx, request)

        validatedError := grpc.ValidateGrpcError(connectErr)
        if validatedError != nil {
             slog.ErrorContext(ctx, "Failed to create connection", "error", validatedError)

            return nil, validatedError
        }

	return response, nil
    }
}
```

Dunno if it'll get messy though

PR #702, opened 2024-06-11
Upload nginx config

dhurley:
### Proposed changes

Upload nginx config

### Checklist

### DIFF
--- a/internal/command/command_plugin.go
+++ b/internal/command/command_plugin.go
@@ -68,32 +68,43 @@ func (cp *CommandPlugin) Info() *bus.Info {
 	}
 }
 
-// nolint: revive, cyclop
 func (cp *CommandPlugin) Process(ctx context.Context, msg *bus.Message) {
 	switch msg.Topic {
 	case bus.ResourceUpdateTopic:
-		if resource, ok := msg.Data.(*mpi.Resource); ok {
-			err := cp.commandService.UpdateDataPlaneStatus(ctx, resource)
-			if err != nil {
-				slog.ErrorContext(ctx, "Unable to update data plane status", "error", err)
-			}
-		}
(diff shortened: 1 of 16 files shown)


internal/command/command_plugin.go
```diff
@@ -67,7 +68,7 @@ func (cp *CommandPlugin) Info() *bus.Info {
 	}
 }
 
-// nolint: revive
+// nolint: revive, cyclop
```
oliveromahony, line 71: cyclop? Can you place case bus.DataPlaneResponseTopic: into it's own function that's testable?

dhurley: fixed

internal/command/command_service.go
```diff
@@ -193,27 +224,34 @@ func (cs *CommandService) subscribe(ctx context.Context) {
 			return
 		default:
 			subscribeFn := func() error {
-				if subscribeClient == nil {
+				cs.subscribeClientMutex.Lock()
```
olli-holmala, line 227: I'm slightly concerned that we are locking constantly in this subscribe loop. Have we considered a `RWLock` here?

olli-holmala: Okay didn't use a `RWLock` but moved the locking bit out of the for loop.

oliveromahony (approved):

aphralG (approved):

internal/command/command_service.go
dhurley: Need to keep the locking bit in the for loop otherwise the retry logic will fail.
I wouldn't be too concerned about the lock in the for loop as it won't be called that often. The `cs.subscribeClient.Recv()` is a blocking call, so unless we get a lot of messages sent to us in one go we won't be iterating in the for loop that often.

craigell (approved):

PR #706, opened 2024-06-13
Change environment prefix from nms to nginx_agent

spencerugbo:
### Proposed changes

Agent now uses NGINX_AGENT as its environment prefix as opposed to NMS. If the old prefix is used, Agent migrates the value to the correct key and a message is logged warning the user to switch to the new prefix

### Checklist

### DIFF
--- a/src/core/config/config.go
+++ b/src/core/config/config.go
@@ -47,7 +47,10 @@ const (
 `
 )
 
-var Viper = viper.NewWithOptions(viper.KeyDelimiter(agent_config.KeyDelimiter))
+var (
+	Viper       = viper.NewWithOptions(viper.KeyDelimiter(agent_config.KeyDelimiter))
+	MigratedEnv = false
+)
 
 func SetVersion(version, commit string) {
 	ROOT_COMMAND.Version = version + "-" + commit
@@ -133,7 +136,7 @@ func deprecateFlags() {
 }
 
 func RegisterFlags() {
-	Viper.SetEnvPrefix(EnvPrefix)
+	Viper.SetEnvPrefix(NewEnvPrefix)
 	Viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	Viper.AutomaticEnv()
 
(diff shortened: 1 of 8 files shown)


src/core/config/defaults.go
```diff
@@ -103,7 +103,8 @@ const (
 	DynamicConfigFileAbsFreeBsdPath = "/var/db/nginx-agent/agent-dynamic.conf"
 	ConfigFileName                  = "nginx-agent.conf"
 	ConfigFileType                  = "yaml"
-	EnvPrefix                       = "nms"
+	OldEnvPrefix                    = "nms"
+	NewEnvPrefix                    = "nginx_agent"
```
olli-holmala, line 107: nit: I think `OldEnvPrefix` should rather be called `LegacyEnvPrefix` and `NewEnvPrefix` just `EnvPrefix`. That way we could eventually get rid of `LegacyEnvPrefix` and we'd be left with `EnvPrefix`.

src/core/config/config.go
```diff
@@ -149,6 +152,17 @@ func RegisterFlags() {
 		if err := Viper.BindPFlag(strings.ReplaceAll(flag.Name, "-", "_"), fs.Lookup(flag.Name)); err != nil {
 			return
 		}
+
+		oldKey := strings.ToUpper(OldEnvPrefix + "_" + strings.ReplaceAll(flag.Name, "-", "_"))
+		newKey := strings.ToUpper(NewEnvPrefix + "_" + strings.ReplaceAll(flag.Name, "-", "_"))
+
+		if os.Getenv(oldKey) != "" && os.Getenv(newKey) == "" {
```
olli-holmala, line 159: What about the (admittedly unlikely) case where both `oldKey` **and** `newKey` are set?

olli-holmala:
Could you add some unit tests for this? I think Go's `testing.T` should support setting env vars nicely with `t.Setenv()`, so this should be testable.

src/core/config/config.go
```diff
@@ -149,6 +152,17 @@ func RegisterFlags() {
 		if err := Viper.BindPFlag(strings.ReplaceAll(flag.Name, "-", "_"), fs.Lookup(flag.Name)); err != nil {
 			return
 		}
+
+		oldKey := strings.ToUpper(OldEnvPrefix + "_" + strings.ReplaceAll(flag.Name, "-", "_"))
+		newKey := strings.ToUpper(NewEnvPrefix + "_" + strings.ReplaceAll(flag.Name, "-", "_"))
```
olli-holmala, line 157: Generally it's good practice to use `fmt.Sprintf()` for building strings instead of "string" + "another_string". However, I think your solution is more readable so it's fine here.

I would perhaps reuse the`agent_config.KeyDelimiter` variable we have here:
```suggestion
		oldKey := strings.ToUpper(
			OldEnvPrefix + agent_config.KeyDelimiter + strings.ReplaceAll(flag.Name, "-", agent_config.KeyDelimiter)
		)
		newKey := strings.ToUpper(
			NewEnvPrefix + agent_config.KeyDelimiter + strings.ReplaceAll(flag.Name, "-", agent_config.KeyDelimiter)
		)
```

src/core/config/config.go
spencerugbo: If both oldKey and newKey are set we assume the newKey is the one with the correct value and ignore the value from the old key

craigell (approved):

aphralG:
Will create task to check MigratedEnv in UnitTests in the future, otherwise LGTM

aphralG (approved):

PR #711, opened 2024-06-18
Config Apply - Write Config 

aphralG:
### Proposed changes

- Added Config apply
- Add CompareFileHash function to compare the files on disk to the what was sent in the fileOverview to determine if files still need the actions performed
- Changed File permission proto
- Added Topics for config apply
- Added File Operator to write files
- Changed generation of fileHash and config version 
- Follow up PR will add Rollback etc

### Checklist


internal/file/file_manager_service.go
```diff
@@ -23,15 +24,30 @@ import (
...
+	fileOperator interface {
+		Write(ctx context.Context, fileContent []byte, file *mpi.FileMeta) error
+	}
+)
+
 type FileManagerService struct {
 	fileServiceClient mpi.FileServiceClient
 	agentConfig       *config.Config
+	fileOperator      fileOperator
+	filesCache        map[string]*mpi.File // key is filePath
```
olli-holmala, line 40:
```suggestion
	filesCache        map[string]*mpi.File // key is file path
```

internal/file/file_manager_service.go
```diff
@@ -142,3 +158,108 @@ func (fms *FileManagerService) UpdateFile(
...
+	return nil
+}
+
+func (fms *FileManagerService) compareHash(filePath string) (bool, error) {
+	_, fileHash, err := files.GenerateHashWithReadFile(filePath)
+	if err != nil {
+		return false, err
+	}
+	slog.Info("hash", "", fileHash)
+	slog.Info("hash", "", fms.filesCache[filePath].GetFileMeta().GetHash())
```
olli-holmala, line 247: These were probably meant to be deleted?

internal/file/file_manager_service.go
```diff
@@ -142,3 +158,108 @@ func (fms *FileManagerService) UpdateFile(
...
+
+	ok, err := fms.compareHash(file.GetFileMeta().GetName())
+	if !ok || err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (fms *FileManagerService) compareHash(filePath string) (bool, error) {
```
olli-holmala, line 241: What does it mean when `compareHash` returns `true`? I would consider renaming this function, as I don't think it's immediately obvious.

olli-holmala: Also, do we need to return the `bool` from `compareHash` at all, as we are always returning an error when the bool is `false`?

Could we just simply return an `error`?

internal/file/file_manager_service_test.go
```diff
@@ -56,3 +60,191 @@ func TestFileManagerService_UpdateFile(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, 1, fakeFileServiceClient.UpdateFileCallCount())
 }
+
+func TestFileManagerService_ConfigApply_Add(t *testing.T) {
+	ctx := context.Background()
+	tempDir := t.TempDir()
+	addAction := mpi.File_FILE_ACTION_ADD
+
+	filePath := fmt.Sprintf("%s/nginx.conf", tempDir)
```
olli-holmala, line 69: Prefer `filepath.Join()`, as it automatically sets the correct path separator.
```suggestion
	filePath := filepath.Join(tempDir, "nginx.conf")
```

internal/file/file_operator_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/nginx/agent/v3/test/helpers"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileOperator_Write(t *testing.T) {
+	ctx := context.Background()
+
+	tempDir := t.TempDir()
+	filePath := fmt.Sprintf("%s/nginx.conf", tempDir)
```
olli-holmala, line 24: Ditto here with `filepath.Join()`

internal/command/command_service.go
```diff
@@ -172,6 +174,36 @@ func (cs *CommandService) UpdateDataPlaneHealth(ctx context.Context, instanceHea
...
+
+		err := cs.subscribeClient.Send(response)
+		if err != nil {
+			slog.ErrorContext(ctx, "Failed to send data plane response", "error", err)
+
+			return err
+		}
+
+		return nil
+	}
```
olli-holmala, line 199: This can be pulled out into its own function: https://github.com/nginx/agent/blob/5f289815d11e5e6ef9852e25982db13abc202763/internal/command/command_service.go#L250-L272

internal/command/command_service.go
```diff
@@ -184,7 +216,6 @@ func (cs *CommandService) CancelSubscription(ctx context.Context) {
 
 // nolint: revive,gocognit
 func (cs *CommandService) subscribe(ctx context.Context) {
```
olli-holmala, line 218: Looks like the adjustments I made to Donal's PR didn't make it in here?
https://github.com/nginx/agent/blob/5f289815d11e5e6ef9852e25982db13abc202763/internal/command/command_service.go#L181-L207

The issue was the repeated locking in the for loop.

pkg/files/file_helpers.go
```diff
@@ -45,34 +48,113 @@ func GetPermissions(fileMode os.FileMode) string {
...
-// GenerateFileHash returns the hash value of a file's contents.
-func GenerateFileHash(filePath string) (string, error) {
-	f, err := os.Open(filePath)
-	if err != nil {
-		return "", err
+func GenerateHash(b []byte) string {
+	return uuid.NewMD5(uuid.Nil, b).String()
+}
+
+// GenerateHashWithReadFile reads file generates hash and returns the content of the file, hash
```
olli-holmala, line 80: Maybe consider rewording this comment, as it's a public facing function.

pkg/files/file_helpers.go
```diff
@@ -45,34 +48,113 @@ func GetPermissions(fileMode os.FileMode) string {
...
+) {
+	fileDiff = make(map[string]*mpi.File)
+	fileContents = make(map[string][]byte)
+
+	for _, file := range fileOverview.GetFiles() {
+		fileName := file.GetFileMeta().GetName()
+		switch file.GetAction() {
+		case mpi.File_FILE_ACTION_DELETE:
+			if _, err = os.Stat(fileName); os.IsNotExist(err) {
+				// File is already deleted skip
```
olli-holmala, line 113:
```suggestion
				// File is already deleted, skip
```

pkg/files/file_helpers.go
```diff
@@ -45,34 +48,113 @@ func GetPermissions(fileMode os.FileMode) string {
...
+			if readErr != nil {
+				return nil, nil, fmt.Errorf("error reading file %s, error: %w", fileName, readErr)
+			}
+			fileContents[fileName] = fileContent
+			fileDiff[fileName] = file
+
+			continue
+		case mpi.File_FILE_ACTION_ADD:
+			if _, err = os.Stat(fileName); os.IsNotExist(err) {
+				// file is new nothing to compare
```
olli-holmala, line 126:
```suggestion
				// file is new, nothing to compare
```

pkg/files/file_helpers.go
```diff
@@ -45,34 +48,113 @@ func GetPermissions(fileMode os.FileMode) string {
...
+			fallthrough
+		case mpi.File_FILE_ACTION_UPDATE:
+			fileContent, fileHash, fileHashErr := GenerateHashWithReadFile(fileName)
+			if fileHashErr != nil {
+				return nil, nil, fmt.Errorf("error generating hash for file %s, error: %w", fileName, fileHashErr)
+			}
+
+			slog.Info("Action", "", file.GetAction())
+			slog.Info("fileHash", "", fileHash)
+			slog.Info("fileHash", "", file.GetFileMeta().GetHash())
```
olli-holmala, line 143: These probably are meant to be deleted?

pkg/files/file_helpers.go
```diff
@@ -45,34 +48,113 @@ func GetPermissions(fileMode os.FileMode) string {
...
+			if fileHash == file.GetFileMeta().GetHash() {
+				// file is same as on disk skip
```
olli-holmala, line 145:
```suggestion
				// file is same as on disk, skip
```

pkg/files/file_helpers.go
```diff
@@ -45,34 +48,113 @@ func GetPermissions(fileMode os.FileMode) string {
...
+				continue
+			}
+
+			fileContents[fileName] = fileContent
+			fileDiff[fileName] = file
+		case mpi.File_FILE_ACTION_UNSPECIFIED, mpi.File_FILE_ACTION_UNCHANGED:
+			// FileAction is UNSPECIFIED or UNCHANGED skipping, treat UNSPECIFIED as if it is UNCHANGED
```
olli-holmala, line 152:
```suggestion
			// FileAction is UNSPECIFIED or UNCHANGED, skipping. Treat UNSPECIFIED as if it is UNCHANGED.
```

internal/file/file_manager_service.go
aphralG: Changed it to return an error, it just checks what was written against the hash didn't really know what to name it, would `fileHashMatch` be better ?

internal/command/command_service.go
aphralG: Merged Donals PR should be changed now

olli-holmala (approved):

internal/file/file_manager_service.go
olli-holmala: Yeah I like `fileHashMatch` more. Alternatively something like `checkFileHash` or `validateFileHash`, which tells the caller that we comparing the hash against a known valid hash.

Up to you though, just so long as it's less ambiguous than `compareHash` from the point of view of whoever is calling the function.

internal/file/file_manager_service.go
```diff
@@ -23,15 +24,30 @@ import (
 	backoffHelpers "github.com/nginx/agent/v3/internal/backoff"
 )
 
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.8.1 -generate
+//counterfeiter:generate . fileOperator
+
+type (
+	fileOperator interface {
```
oliveromahony, line 31: do we need to check for interface compliance here?

aphralG: Interface compliance is checked in file_operator.go https://github.com/nginx/agent/blob/b87eeb1d3d14c3a2e4ad076e18a8a518a1b55b6d/internal/file/file_operator.go#L22

internal/bus/topics.go
```diff
@@ -6,18 +6,16 @@
...
+	UpdatedInstancesTopic             = "updated-instances"
+	DeletedInstancesTopic             = "deleted-instances"
+	ResourceUpdateTopic               = "resource-update"
+	NginxConfigUpdateTopic            = "nginx-config-update"
+	InstanceHealthTopic               = "instance-health"
+	ConfigUploadRequestTopic          = "config-upload-request"
+	DataPlaneResponseTopic            = "data-plane-response"
+	ConfigApplyRequestTopic           = "config-apply-topic"
+	ConfigApplySuccessfulRequestTopic = "config-apply-successful-topic"
+	ConfigApplyFailedRequestTopic     = "config-apply-failed-topic"
```
dhurley, line 19:
```suggestion
	ConfigApplyRequestTopic           = "config-apply-request"
	ConfigApplySuccessfulRequestTopic = "config-apply-successful-request"
	ConfigApplyFailedRequestTopic     = "config-apply-failed-request"
```

pkg/files/file_helpers.go
```diff
@@ -7,35 +7,37 @@
...
 		return nil, err
 	}
 
-	return &v1.FileMeta{
+	return &mpi.FileMeta{
 		Name:         filePath,
 		Hash:         hash,
 		ModifiedTime: timestamppb.New(fileInfo.ModTime()),
-		Permissions:  fileInfo.Mode().Perm().String(),
+		Permissions:  "0777",
```
dhurley, line 40: Is this meant to be hardcoded to `0777`?

pkg/files/file_helpers.go
```diff
@@ -45,34 +47,109 @@ func GetPermissions(fileMode os.FileMode) string {
...
-func GenerateFileHash(filePath string) (string, error) {
-	f, err := os.Open(filePath)
-	if err != nil {
-		return "", err
+func GenerateHash(b []byte) string {
+	return uuid.NewMD5(uuid.Nil, b).String()
+}
+
+// GenerateHashWithReadFile returns the content of a file and the hash of that file.
+func GenerateHashWithReadFile(filePath string) ([]byte, string, error) {
```
dhurley, line 80: Feels like this function is doing too much. Would it be better to just have a `ReadFile` function and call the `GenerateHash` function outside of it?

```
contents, err := ReadFile("...")
hash := GenerateHash(contents)
```

pkg/files/file_helpers.go
```diff
@@ -45,34 +47,109 @@ func GetPermissions(fileMode os.FileMode) string {
...
+				continue
+			}
+			fileContent, _, readErr := GenerateHashWithReadFile(fileName)
+			if readErr != nil {
+				return nil, nil, fmt.Errorf("error reading file %s, error: %w", fileName, readErr)
+			}
+			fileContents[fileName] = fileContent
+			fileDiff[fileName] = file
+
+			continue
```
dhurley, line 121: unnecessary `continue`

internal/file/file_plugin.go
```diff
@@ -73,9 +75,83 @@ func (fp *FilePlugin) Subscriptions() []string {
...
+		response := &mpi.DataPlaneResponse{
+			MessageMeta: &mpi.MessageMeta{
+				MessageId:     uuid.NewString(),
+				CorrelationId: correlationID,
+				Timestamp:     timestamppb.Now(),
+			},
+			CommandResponse: &mpi.CommandResponse{
+				Status:  mpi.CommandResponse_COMMAND_STATUS_ERROR,
+				Message: "Config apply failed",
+				Error:   "unable to cast to message payload",
```
dhurley, line 105: To the end user they won't know what `unable to cast to message payload` means. Would it be better here to say something like `Internal server error`?

internal/file/file_plugin.go
```diff
@@ -73,9 +75,83 @@ func (fp *FilePlugin) Subscriptions() []string {
...
+	}
+	configApplyRequest := request.ConfigApplyRequest
+	var response *mpi.DataPlaneResponse
+
+	err := fp.fileManagerService.ConfigApply(ctx, configApplyRequest)
+
+	if err != nil {
+		slog.ErrorContext(
+			ctx,
+			"Failed to update file overview",
```
dhurley, line 120:
```suggestion
			"Failed to apply config changes",
```

internal/file/file_plugin.go
```diff
@@ -90,6 +166,8 @@ func (fp *FilePlugin) handleNginxConfigUpdate(ctx context.Context, msg *bus.Mess
 			"instance_id", nginxConfigContext.InstanceID,
 			"error", err,
 		)
+
+		return
```
dhurley, line 170: unnecessary `return`

internal/file/file_operator_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	err := fileOp.Write(ctx, fileContent, fileMeta)
+	require.NoError(t, err)
+	assert.FileExists(t, filePath)
+
+	data, err := os.ReadFile(filePath)
+	require.NoError(t, err)
+	assert.Equal(t, fileContent, data)
+
+	helpers.RemoveFileWithErrorCheck(t, filePath)
```
dhurley, line 43: Might be better to use `defer` at the start of the test to ensure that the file gets removed even if the test fails

internal/file/file_manager_service.go
```diff
@@ -160,3 +176,106 @@ func (fms *FileManagerService) UpdateFile(
...
+
+	fileErr := fms.fileActions(ctx)
+	if fileErr != nil {
+		return fileErr
+	}
+
+	return nil
+}
+
+func (fms *FileManagerService) fileActions(ctx context.Context) error {
```
dhurley, line 208:
```suggestion
func (fms *FileManagerService) executeFileActions(ctx context.Context) error {
```

internal/file/file_manager_service.go
```diff
@@ -160,3 +176,106 @@ func (fms *FileManagerService) UpdateFile(
...
+			continue
+		case mpi.File_FILE_ACTION_ADD, mpi.File_FILE_ACTION_UPDATE:
+			updateErr := fms.fileUpdate(ctx, file)
+			if updateErr != nil {
+				return updateErr
+			}
+		case mpi.File_FILE_ACTION_UNSPECIFIED, mpi.File_FILE_ACTION_UNCHANGED:
+			fallthrough
+		default:
+			slog.DebugContext(ctx, "File Action not implemented")
```
dhurley, line 225:
```suggestion
			slog.DebugContext(ctx, "File Action not implemented", "action", file.GetAction())
```

internal/bus/topics.go
aphralG: I've changed these in Rollback for Failed and success no longer have request in the name, can I change ConfigApplyRequestTopic  there instead to avoid issues ?

pkg/files/file_helpers.go
aphralG: No will change it back

pkg/files/file_helpers.go
aphralG: Fixed

internal/file/file_manager_service.go
```diff
@@ -160,3 +176,107 @@ func (fms *FileManagerService) UpdateFile(
...
+			CorrelationId: logger.GetCorrelationID(ctx),
+			Timestamp:     timestamppb.Now(),
+		},
+		FileMeta: file.GetFileMeta(),
+	})
+	if getFileErr != nil {
+		return fmt.Errorf("error getting file data for %s: %w", file.GetFileMeta(), getFileErr)
+	}
+
+	writeErr := fms.fileOperator.Write(ctx, getFileResp.GetContents().GetContents(), file.GetFileMeta())
```
oliveromahony, line 245: Marginally prefer something like this: 

```if writeErr := fms.fileOperator.Write(ctx, getFileResp.GetContents().GetContents(), file.GetFileMeta()); writeErr != nil {
        return writeErr
    }```

for lines 245-249

Comment (oliveromahony) [reply internal/file/file_manager_service.go:245]: same comments for below. Seems less verbose in the 3 lines format

File: internal/file/file_manager_service_test.go [hunk 26]
```diff
@@ -58,3 +62,191 @@ func TestFileManagerService_UpdateFile(t *testing.T) {
...
+	ctx := context.Background()
+	tempDir := t.TempDir()
+	addAction := mpi.File_FILE_ACTION_ADD
+
+	filePath := filepath.Join(tempDir, "nginx.conf")
+	fileContent := []byte("location /test {\n    return 200 \"Test location\\n\";\n}")
+	fileHash := files.GenerateHash(fileContent)
+	defer helpers.RemoveFileWithErrorCheck(t, filePath)
+
+	overview := mpi.FileOverview{
```
oliveromahony, line 76: can you add FileOverview to the test helpers? you might need to pass filePath and fileHash

internal/file/file_manager_service_test.go
```diff
@@ -58,3 +62,191 @@ func TestFileManagerService_UpdateFile(t *testing.T) {
...
+		Files: []*mpi.File{
+			{
+				FileMeta: &mpi.FileMeta{
+					Name:         filePath,
+					Hash:         fileHash,
+					ModifiedTime: timestamppb.Now(),
+					Permissions:  "0777",
```
oliveromahony, line 83: default these to 0640, even for tests

internal/file/file_manager_service_test.go
```diff
@@ -58,3 +62,191 @@ func TestFileManagerService_UpdateFile(t *testing.T) {
...
+	fakeFileServiceClient.GetFileReturns(&mpi.GetFileResponse{
+		Contents: &mpi.FileContents{
+			Contents: fileContent,
+		},
+	}, nil)
+	agentConfig := types.AgentConfig()
+	agentConfig.AllowedDirectories = []string{tempDir}
+	fileManagerService := NewFileManagerService(fakeFileServiceClient, agentConfig)
+
+	request := &mpi.ConfigApplyRequest{
```
oliveromahony, line 160: can protos.CreateManagementPlaneRequestConfigApplyRequest() be used here?

internal/file/file_manager_service_test.go
```diff
@@ -58,3 +62,191 @@ func TestFileManagerService_UpdateFile(t *testing.T) {
...
+	assert.NoFileExists(t, tempFile.Name())
+	assert.Equal(t, fileManagerService.fileContentsCache[tempFile.Name()], fileContent)
+	assert.Equal(t, fileManagerService.filesCache[tempFile.Name()], overview.GetFiles()[0])
+}
+
+func TestFileManagerService_checkAllowedDirectory(t *testing.T) {
+	fakeFileServiceClient := &v1fakes.FakeFileServiceClient{}
+	fileManagerService := NewFileManagerService(fakeFileServiceClient, types.AgentConfig())
+
+	allowedFiles := []*mpi.File{
```
oliveromahony, line 222: this would be a useful test helper I think

internal/file/file_operator.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/nginx/agent/v3/pkg/files"
+
+	mpi "github.com/nginx/agent/v3/api/grpc/mpi/v1"
+)
+
+type FileOperator struct{}
+
+var _ fileOperator = (*FileOperator)(nil)
+
+func NewFileOperator() *FileOperator {
```
oliveromahony, line 24: Can you add comments to describe the purpose of the File Operator?

internal/file/file_operator_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileOperator_Write(t *testing.T) {
+	ctx := context.Background()
+
+	tempDir := t.TempDir()
+	filePath := filepath.Join(tempDir, "nginx.conf")
+	fileContent := []byte("location /test {\n    return 200 \"Test location\\n\";\n}")
```
oliveromahony, line 26: I think we should centralize the nginx configurations for testing purposes rather than have embedded strings everywhere

internal/file/file_operator_test.go
```diff
@@ -0,0 +1,44 @@
...
+	defer helpers.RemoveFileWithErrorCheck(t, filePath)
+	fileOp := NewFileOperator()
+	fileMeta := &mpi.FileMeta{
```
oliveromahony, line 29: looks like this could be centralized as well

internal/file/file_plugin.go
```diff
@@ -73,9 +75,83 @@ func (fp *FilePlugin) Subscriptions() []string {
...
+			CommandResponse: &mpi.CommandResponse{
+				Status: mpi.CommandResponse_COMMAND_STATUS_ERROR,
+				Message: fmt.Sprintf("Config apply failed for instanceId: %s", configApplyRequest.
+					GetConfigVersion().GetInstanceId()),
+				Error: err.Error(),
+			},
+		}
+	} else {
+		response = &mpi.DataPlaneResponse{
+			MessageMeta: &mpi.MessageMeta{
```
oliveromahony, line 140: looks like MessageMeta is the same for both blocks

pkg/files/file_helpers_test.go
```diff
@@ -20,10 +24,10 @@ func TestGetFileMeta(t *testing.T) {
 	defer helpers.RemoveFileWithErrorCheck(t, file.Name())
 	require.NoError(t, err)
 
-	expected := &v1.FileMeta{
+	expected := &mpi.FileMeta{
 		Name:        file.Name(),
-		Hash:        "47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=",
-		Permissions: "-rw-------",
+		Hash:        "4ae71336-e44b-39bf-b9d2-752e234818a5",
```
oliveromahony, line 29: I don't think this should be in UUID format?

internal/file/file_plugin_test.go
```diff
@@ -85,6 +90,103 @@ func TestFilePlugin_Process_NginxConfigUpdateTopic(t *testing.T) {
...
+	ctx := context.Background()
+	tempDir := t.TempDir()
+	addAction := mpi.File_FILE_ACTION_ADD
+
+	filePath := fmt.Sprintf("%s/nginx.conf", tempDir)
+	fileContent := []byte("location /test {\n    return 200 \"Test location\\n\";\n}")
+	fileHash := files.GenerateHash(fileContent)
+
+	message := &mpi.ManagementPlaneRequest{
+		Request: &mpi.ManagementPlaneRequest_ConfigApplyRequest{
```
oliveromahony, line 103: this seems useful to reuse in helpers

pkg/files/file_helpers_test.go
```diff
@@ -51,47 +55,279 @@ func TestGetPermissions(t *testing.T) {
...
+
+	tests := []struct {
+		name             string
+		fileOverview     *mpi.FileOverview
+		expectedDiff     map[string]*mpi.File
+		expectedContents map[string][]byte
+	}{
+		{
+			name: "Test 1: Delete, Add & Update Files",
+			fileOverview: &mpi.FileOverview{
```
oliveromahony, line 191: same comment here

PR #719, opened 2024-06-24
Fix check for comparing children PIDs

dhurley:
### Proposed changes

Fix check for comparing children PIDs

### Checklist

### DIFF
--- a/internal/watcher/instance/instance_watcher_service.go
+++ b/internal/watcher/instance/instance_watcher_service.go
@@ -273,6 +273,7 @@ func areInstancesEqual(oldRuntime, currentRuntime *mpi.InstanceRuntime) (equal b
 		for _, currentChild := range currentRuntimeChildren {
 			if oldChild.GetProcessId() == currentChild.GetProcessId() {
 				childFound = true
+				break
 			}
 		}
 
--- a/internal/watcher/instance/instance_watcher_service.go
+++ b/internal/watcher/instance/instance_watcher_service.go
@@ -8,16 +8,14 @@ package instance
 import (
 	"context"
 	"log/slog"
-	"reflect"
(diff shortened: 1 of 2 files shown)


internal/watcher/instance/instance_watcher_service.go
```diff
@@ -263,7 +261,27 @@ func areInstancesEqual(oldRuntime, currentRuntime *mpi.InstanceRuntime) (equal b
...
+
+	if len(oldRuntimeChildren) != len(currentRuntimeChildren) {
+		return false
+	}
+
+	for _, oldChild := range oldRuntimeChildren {
+		childFound := false
+		for _, currentChild := range currentRuntimeChildren {
+			if oldChild.GetProcessId() == currentChild.GetProcessId() {
+				childFound = true
```
oliveromahony, line 275: break here?

internal/watcher/instance/instance_watcher_service.go
```diff
@@ -263,7 +261,27 @@ func areInstancesEqual(oldRuntime, currentRuntime *mpi.InstanceRuntime) (equal b
...
+			}
+		}
+
+		if !childFound {
```
oliveromahony, line 279: return childFound is ok on it's own?

internal/watcher/instance/instance_watcher_service.go
dhurley: yes I'll update that

internal/watcher/instance/instance_watcher_service.go
dhurley: no if its true I dont want to return. I want to check if the next child is found

oliveromahony (approved):

PR #730, opened 2024-07-04
Update Data Plane Response Statuses & Clean Up

aphralG:
### Proposed changes

- Cleaned up data plane responses to match diagrams 
- When we fail to cast the payload return (we weren't in a few places)
- Added error to ConfigApply model so that we have access to the original cause of the config apply failure when sending the Failure response after rollback
- If there is no difference between the file overview sent in the config apply request and the files on disk send a Command Status OK response and skip validate, reload and monitor 
- General Clean Up

### Checklist


oliveromahony (approved):

internal/file/file_manager_service.go
```diff
@@ -77,6 +79,10 @@ func (r *RollbackRequiredError) Error() string {
 	return fmt.Sprintf("rollback required: %v", r.Err)
 }
 
+func (r *NoChangeError) Error() string {
+	return "no change required"
```
oliveromahony, line 83: maybe add a const for this string

oliveromahony:
can you remove the IN_PROGRESS from being sent over the wire?

internal/file/file_plugin.go
```diff
@@ -36,7 +35,8 @@ type FilePlugin struct {
 func NewFilePlugin(agentConfig *config.Config, grpcConnection grpc.GrpcConnectionInterface) *FilePlugin {
 	return &FilePlugin{
 		config: agentConfig,
-		conn:   grpcConnection,
+		// why?
```
oliveromahony, line 38: why? 👯

aphralG: Fixed

internal/model/config.go
```diff
@@ -34,3 +35,14 @@ type ErrorLog struct {
 	Permissions string
 	Readable    bool
 }
+
+type (
```
dhurley, line 39: If `WriteStatus` is only used in the `file` package, then I think it makes sense to move this enum there instead of the `model` package.

dhurley (approved):

internal/model/config.go
aphralG: Was talking with Oliver about it, had an issue with import cycle and we might end up using it for the status later when we implement it.

internal/resource/resource_plugin.go
```diff
@@ -114,24 +120,27 @@ func (*Resource) Subscriptions() []string {
...
+		slog.ErrorContext(ctx, "Unable to cast message payload to *model.ConfigApplyMessage", "payload", msg.Data)
+
+		return
 	}
 	err := r.resourceService.ApplyConfig(ctx, data.InstanceID)
 	if err != nil {
 		slog.Error("errors found during config apply, sending failure status", "err", err)
-
-		response := r.createDataPlaneResponseWithError(data.CorrelationID, mpi.CommandResponse_COMMAND_STATUS_ERROR,
+		// data.Error = err
```
olli-holmala, line 132: Delete this extra comment?

PR #739, opened 2024-07-09
Fix Post Install Script Issues

spencerugbo:
### Proposed changes

This PR aims to prevent silent fails when agent is installed in a Debian docker image. PostInstall.sh no longer uses the ps command to detect nginx users.

### Checklist

### DIFF
--- a/scripts/packages/postinstall.sh
+++ b/scripts/packages/postinstall.sh
@@ -21,7 +21,7 @@ WORKER_USER=""
 AGENT_GROUP="nginx-agent"
 
 detect_nginx_users() {
-    if command -V systemctl >/dev/null 2>&1; then
+    if command -V systemctl; then
         printf "PostInstall: Reading NGINX systemctl unit file for user information\n"
         nginx_unit_file=$(systemctl status nginx | grep -Po "\(\K\/.*service")
         pid_file=$(grep -Po "PIDFile=\K.*$" "${nginx_unit_file}")
@@ -122,7 +122,7 @@ ensure_agent_path() {
 create_agent_group() {
     printf "PostInstall: Adding nginx-agent group %s\n" "${AGENT_GROUP}"
 
-    if command -V groupadd >/dev/null 2>&1; then
+    if command -V groupadd; then
         if [ ! "$(getent group $AGENT_GROUP)" ]; then
             groupadd "${AGENT_GROUP}"
         fi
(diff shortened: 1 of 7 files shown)


scripts/packages/postinstall.sh
```diff
@@ -39,7 +40,18 @@ detect_nginx_users() {
 
     if [ -z "${nginx_user}" ]; then
         printf "PostInstall: Reading NGINX process information to determine NGINX user\n"
-        nginx_user=$(ps aux | grep "nginx: master process" | grep -v grep | head -1 | awk '{print $1}')
+        nginx_pid=""
+        for pid in $(ls /proc | grep -E '^[0-9]+$'); do
```
olli-holmala, line 44: Remember to install the Shellcheck extension for VS Code: https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck

My Shellcheck is complaining about this `ls` command, here's the error it refers to: https://www.shellcheck.net/wiki/SC2010

dhurley:
@spencerugbo Looks like you need to merge the main into your branch.
Also I noticed that some integration tests are failing. Think its related to the changes you made.

scripts/packages/postinstall.sh
```diff
@@ -141,7 +168,7 @@ create_run_dir() {
 
 update_unit_file() {
     # Fill in data to unit file that's acquired post install
-    if command -V systemctl >/dev/null 2>&1; then
+    if command -v systemctl && [ "$(cat /proc/1/comm)" == "systemd" ]; then
```
olli-holmala, line 171: Probably should keep the uppercase `-V`.

scripts/packages/postinstall.sh
```diff
@@ -20,7 +21,7 @@ WORKER_USER=""
 AGENT_GROUP="nginx-agent"
 
 detect_nginx_users() {
-    if command -V systemctl >/dev/null 2>&1; then
+    if command -v systemctl && [ "$(cat /proc/1/comm)" == "systemd" ]; then
```
olli-holmala, line 24: Probably should keep the uppercase -V.
```suggestion
    if command -V systemctl && [ "$(cat /proc/1/comm)" == "systemd" ]; then
```
Unless there was a reason for the change?

olli-holmala (approved):

dhurley (approved):

scripts/packages/postremove.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
```
oliveromahony, line 2: can we remove >/dev/null 2>&1 from this file as well?

scripts/packages/postupgrade.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
```
oliveromahony, line 2: can we remove >/dev/null 2>&1 from this file?

scripts/packages/preremove.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+set -e
```
oliveromahony, line 2: can we remove >/dev/null 2>&1 from this file?

aphralG (approved):

oliveromahony (approved):

PR #740, opened 2024-07-09
move grpc close logic to a single place

oliveromahony:
### Proposed changes

Move the gRPC close connection on the close signal of application instead of being called in multiple plugins

### Checklist


internal/app.go
```diff
@@ -22,51 +31,115 @@ const (
...
 			return
 		}
 
-		agentConfig, err := config.ResolveConfig()
-		if err != nil {
-			slog.ErrorContext(ctx, "Invalid config", "error", err)
+		agentConfig, configErr := config.ResolveConfig()
+		if configErr != nil {
+			slog.ErrorContext(ctx, "Invalid config", "error", configErr)
 			return
```
olli-holmala, line 58: Now that we've renamed the error to `configErr`, when we hit `return` here `err` is nil. Are we sure we want to return a nil `err` here?

internal/app.go
```diff
@@ -22,51 +31,115 @@ const (
...
 		)
 
 		messagePipe := bus.NewMessagePipe(defaultMessagePipeChannelSize)
-		err = messagePipe.Register(defaultQueueSize, plugin.LoadPlugins(ctx, agentConfig))
-		if err != nil {
-			slog.ErrorContext(ctx, "Failed to register plugins", "error", err)
+		msgPipeErr := messagePipe.Register(defaultQueueSize, a.loadPlugins(agentConfig))
+		if msgPipeErr != nil {
+			slog.ErrorContext(ctx, "Failed to register plugins", "error", msgPipeErr)
 			return
```
olli-holmala, line 73: Ditto here

internal/app.go
```diff
@@ -22,51 +31,115 @@ const (
...
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		select {
+		case <-sigChan:
+			slog.DebugContext(ctx, "NGINX Agent App exiting")
+		case <-ctx.Done():
+			if a.grpcConn != nil {
+				connCloseErr := a.grpcConn.Close(ctx)
+				slog.ErrorContext(ctx, "Issue closing gRPC connection", "error", connCloseErr)
```
olli-holmala, line 97: Should we have a nil check here for `connCloseErr` before logging that there was an error?

internal/app.go
```diff
@@ -22,51 +31,115 @@ const (
...
+func (a *App) loadPlugins(agentConfig *config.Config) []bus.Plugin {
+	plugins := make([]bus.Plugin, 0)
+
+	plugins = a.addResourcePlugin(plugins, agentConfig)
+
+	if a.isGrpcClientConfigured(agentConfig) {
+		commandPlugin := command.NewCommandPlugin(agentConfig, a.grpcConn)
+		plugins = append(plugins, commandPlugin)
+		filePlugin := file.NewFilePlugin(agentConfig, a.grpcConn)
+		plugins = append(plugins, filePlugin)
```
olli-holmala, line 112:
```suggestion
		commandPlugin := command.NewCommandPlugin(agentConfig, a.grpcConn)
		plugins = append(plugins,
		  command.NewCommandPlugin(agentConfig, a.grpcConn),
		  file.NewFilePlugin(agentConfig, a.grpcConn),
		)
```

internal/app.go
```diff
@@ -22,51 +31,115 @@ const (
...
+		agentConfig.Command.Server.Type == config.Grpc
+}
+
+func (a *App) addCollector(agentConfig *config.Config, plugins []bus.Plugin) []bus.Plugin {
+	if agentConfig.Collector != nil {
+		oTelCollector, err := collector.New(agentConfig)
+		if err == nil {
+			plugins = append(plugins, oTelCollector)
+		} else {
+			slog.Error("init collector plugin", "error", err)
```
olli-holmala, line 140:
```suggestion
			slog.Error("Failed to initialize the OTel collector plugin", "error", err)
```

internal/app.go
```diff
@@ -22,51 +31,115 @@ const (
...
+
+	return plugins
+}
+
+func (a *App) addResourcePlugin(plugins []bus.Plugin, agentConfig *config.Config) []bus.Plugin {
+	resourcePlugin := resource.NewResource(agentConfig)
+	plugins = append(plugins, resourcePlugin)
+
+	return plugins
+}
```
olli-holmala, line 126: nitpick on consistency: Does this need to be in its own function, since we are doing this same "init plugin and append to slice" combo in `loadPlugins` for other plugins?

olli-holmala (approved):
Good stuff, just had minor comments 👌

internal/app.go
oliveromahony: it is inside config.RegisterRunner and the return breaks out of the function. I'm not sure what the best way to do this is since the Run function can't get that error because previously, the ```err := config.Execute(ctx)``` overrode the err value

olli-holmala: Ah right, missed that. Dunno could we change the `RegisterRunner` function so that we pass it a lambda that return an `error` instead of the what it's like now? So the new implementation would be something like:
```go
func RegisterRunner(r func(cmd *cobra.Command, args []string) error) {
	RootCommand.RunE = r
}
```

oliveromahony: ah nice, will check this

PR #742, opened 2024-07-09
NGINX plus receiver

dhurley:
### Proposed changes

- Collect NGINX Plus metrics
- Update collector configuration to be dynamic depending on want instances are found
- Add build agent with nginx plus image make target
- Add mock OTel collector package for testing purposes

### Checklist


internal/config/types.go
```diff
@@ -103,15 +96,31 @@ type (
...
 		Type string `yaml:"-" mapstructure:"type"`
 	}
-
 	// OTel Collector Receiver configuration.
-	Receiver struct {
+	Receivers struct {
+		OtlpReceivers      []OtlpReceiver      `yaml:"-" mapstructure:"otlp_receivers"`
+		NginxReceivers     []NginxReceiver     `yaml:"-" mapstructure:"nginx_receivers"`
+		NginxPlusReceivers []NginxPlusReceiver `yaml:"-" mapstructure:"nginx_plus_receivers"`
+		HostMetrics        bool                `yaml:"-" mapstructure:"host_metrics"`
```
olli-holmala, line 104: consistency: either we should rename `HostMetrics` to `HostMetricsReceivers` or remove the `Receivers` suffix from all of these altogether. I would opt for the latter approach, as the suffix is redundant because we are within the `Receivers` struct.

.golangci.yml
```diff
@@ -1,6 +1,15 @@
...
     - ".*gen.go"
+    - ".pb.go"
+    - ".pb.validate.go"
+    - internal/metadata/generated_metrics.go
+    - internal/metadata/generated_config.go
+    - internal/metadata/generated_metrics_test.go
+    - internal/metadata/generated_config_test.go
+    - internal/collector/nginxplusreceiver/generated_component_test.go
+    - internal/collector/nginxplusreceiver/generated_package_test.go
+    - internal/collector/nginxplusreceiver/internal/metadata/generated_status.go
```
olli-holmala, line 12: Could these be replaced with something like a `.*/generated_[A-Za-z_]*\.go`?

internal/collector/otel_collector_plugin.go
```diff
@@ -117,19 +121,190 @@ func (oc *Collector) Close(ctx context.Context) error {
...
 }
 
 // Process an incoming Message Bus message in the plugin
-func (oc *Collector) Process(_ context.Context, msg *bus.Message) {
+func (oc *Collector) Process(ctx context.Context, msg *bus.Message) {
+	switch msg.Topic {
+	case bus.NginxConfigUpdateTopic:
+		oc.handleNginxConfigUpdate(ctx, msg)
+	default:
+		slog.DebugContext(ctx, "File plugin unknown topic", "topic", msg.Topic)
```
olli-holmala, line 155:
```suggestion
		slog.DebugContext(ctx, "OTel collector plugin unknown topic", "topic", msg.Topic)
```

internal/collector/otel_collector_plugin.go
```diff
@@ -117,19 +121,190 @@ func (oc *Collector) Close(ctx context.Context) error {
...
+
+	var runCtx context.Context
+	runCtx, oc.cancel = context.WithCancel(ctx)
+
+	go func() {
+		bootErr := oc.bootup(runCtx)
+		if bootErr != nil {
+			slog.ErrorContext(runCtx, "Unable to start OTel Collector", "error", bootErr)
+		}
+	}()
```
olli-holmala, line 209: Don't think we need to run `oc.bootup()` in a goroutine, as `bootup` starts the collector in its own goroutine.

dhurley: There is a for loop in the `oc.bootup()` function which will stop the function from exiting. Thats why we need to run `oc.bootup()` in a goroutine

olli-holmala: Yeah but that for loop only waits for a Running state and then returns a nil error. I would think we would block the `restartCollector` func in the same way that the `bootup` func does. Not a big deal if we keep this as-is though.

internal/collector/otel_collector_plugin.go
```diff
@@ -117,19 +121,190 @@ func (oc *Collector) Close(ctx context.Context) error {
...
+	nginxReceiverFound = false
+
+	for index, nginxPlusReceiver := range oc.config.Collector.Receivers.NginxPlusReceivers {
+		if nginxPlusReceiver.InstanceID == nginxConfigContext.InstanceID {
+			nginxReceiverFound = true
+
+			if nginxPlusReceiver.PlusAPI != nginxConfigContext.PlusAPI {
+				oc.config.Collector.Receivers.NginxPlusReceivers = append(
+					oc.config.Collector.Receivers.NginxPlusReceivers[:index],
+					oc.config.Collector.Receivers.NginxPlusReceivers[index+1:]...,
```
olli-holmala, line 257: Won't this cause a "slice bounds out of range" panic if `nginxPlusReceiver` is the last element in `oc.config.Collector.Receivers.NginxPlusReceivers`?

olli-holmala (approved):

internal/collector/otel_collector_plugin.go
dhurley: oh never mind you are right the extra go routine is redundant. I missed `return nil` when the collector is in a running state

internal/collector/otel_collector_plugin.go
dhurley: I dont think so I have a unit test here that passes where I remove the receiver from the slice https://github.com/nginx/agent/blob/intro-nginx-plus-receiver/internal/collector/otel_collector_plugin_test.go#L288

olli-holmala: I tested this out in Go playground and it does **not** go out of bounds 👍 

So if you subslice with `slice[len(slice):]`, the result is an empty slice, but if you do `slice[len(slice)+1:]` you get the "slice out of bounds" error. Learn something new every day.

Makefile
```diff
@@ -177,12 +177,28 @@ run-mock-management-grpc-server: ## Run mock management plane gRPC server
...
-generate: ## Generate proto files and server and client stubs from OpenAPI specifications
+build-test-plus-image: local-deb-package
+	$(CONTAINER_BUILDENV) $(CONTAINER_CLITOOL) build -t nginx_plus_${IMAGE_TAG} . \
+		--no-cache -f ./scripts/docker/nginx-plus/deb/Dockerfile \
+		--secret id=nginx-crt,src=${CERTS_DIR}/nginx-repo.crt \
+		--secret id=nginx-key,src=${CERTS_DIR}/nginx-repo.key \
+		--build-arg PACKAGE_NAME=${PACKAGE_NAME} \
+		--build-arg PACKAGES_REPO=${OSS_PACKAGES_REPO} \
+		--build-arg BASE_IMAGE=${BASE_IMAGE} \
+		--build-arg ENTRY_POINT=./scripts/docker/entrypoint.sh
```
olli-holmala, line 188: nit: could we use parentheses consistently? Becomes very messy over time if we just randomly use both.
```suggestion
	$(CONTAINER_BUILDENV) $(CONTAINER_CLITOOL) build -t nginx_plus_$(IMAGE_TAG) . \
		--no-cache -f ./scripts/docker/nginx-plus/deb/Dockerfile \
		--secret id=nginx-crt,src=$(CERTS_DIR)/nginx-repo.crt \
		--secret id=nginx-key,src=$(CERTS_DIR)/nginx-repo.key \
		--build-arg PACKAGE_NAME=$(PACKAGE_NAME) \
		--build-arg PACKAGES_REPO=$(OSS_PACKAGES_REPO) \
		--build-arg BASE_IMAGE=$(BASE_IMAGE) \
		--build-arg ENTRY_POINT=./scripts/docker/entrypoint.sh
```

Makefile
```diff
@@ -177,12 +177,28 @@ run-mock-management-grpc-server: ## Run mock management plane gRPC server
...
+
+run-mock-management-otel-collector: ## Run mock management plane OTel collector
+	@echo "🚀 Running mock management plane OTel collector"
+	AGENT_IMAGE=nginx_plus_${IMAGE_TAG}:latest $(CONTAINER_COMPOSE) -f ./test/mock/collector/docker-compose.yaml up -d
```
olli-holmala, line 192:
```suggestion
	AGENT_IMAGE=nginx_plus_$(IMAGE_TAG):latest $(CONTAINER_COMPOSE) -f ./test/mock/collector/docker-compose.yaml up -d
```

Makefile
```diff
@@ -177,12 +177,28 @@ run-mock-management-grpc-server: ## Run mock management plane gRPC server
...
+
+stop-mock-management-otel-collector: ## Stop running mock management plane OTel collector
+	@echo "Stopping mock management plane OTel collector"
+	AGENT_IMAGE=nginx_plus_${IMAGE_TAG}:latest $(CONTAINER_COMPOSE) -f ./test/mock/collector/docker-compose.yaml down
```
olli-holmala, line 196:
```suggestion
	AGENT_IMAGE=nginx_plus_$(IMAGE_TAG):latest $(CONTAINER_COMPOSE) -f ./test/mock/collector/docker-compose.yaml down
```

Makefile
```diff
@@ -177,12 +177,28 @@ run-mock-management-grpc-server: ## Run mock management plane gRPC server
 	@echo "🖲️ Running mock management plane gRPC server"
 	$(GORUN) test/mock/grpc/cmd/main.go -configDirectory=$(MOCK_MANAGEMENT_PLANE_CONFIG_DIRECTORY) -logLevel=$(MOCK_MANAGEMENT_PLANE_LOG_LEVEL) -grpcAddress=$(MOCK_MANAGEMENT_PLANE_GRPC_ADDRESS) -apiAddress=$(MOCK_MANAGEMENT_PLANE_API_ADDRESS)
 
-generate: ## Generate proto files and server and client stubs from OpenAPI specifications
+build-test-plus-image: local-deb-package
```
olli-holmala, line 180:
```suggestion
.PHONY: build-test-plus-image
build-test-plus-image: local-deb-package
```

Makefile
```diff
@@ -177,12 +177,28 @@ run-mock-management-grpc-server: ## Run mock management plane gRPC server
...
+	$(CONTAINER_BUILDENV) $(CONTAINER_CLITOOL) build -t nginx_plus_${IMAGE_TAG} . \
+		--no-cache -f ./scripts/docker/nginx-plus/deb/Dockerfile \
+		--secret id=nginx-crt,src=${CERTS_DIR}/nginx-repo.crt \
+		--secret id=nginx-key,src=${CERTS_DIR}/nginx-repo.key \
+		--build-arg PACKAGE_NAME=${PACKAGE_NAME} \
+		--build-arg PACKAGES_REPO=${OSS_PACKAGES_REPO} \
+		--build-arg BASE_IMAGE=${BASE_IMAGE} \
+		--build-arg ENTRY_POINT=./scripts/docker/entrypoint.sh
+
+run-mock-management-otel-collector: ## Run mock management plane OTel collector
```
olli-holmala, line 190:
```suggestion
.PHONY: run-mock-management-otel-collector
run-mock-management-otel-collector: ## Run mock management plane OTel collector
```

internal/collector/otelcol.tmpl
```diff
@@ -1,15 +1,5 @@
...
-  otlp:
-    protocols:
-      {{- if eq .Server.Type 0 }}
-      grpc:
-      {{- else }}
-      http:
-      {{- end }}
-        endpoint: "{{- .Server.Host -}}:{{- .Server.Port -}}"
-  {{- else if eq .Type "hostmetrics" }}
+{{- if .Receivers.HostMetrics }}
```
oliveromahony, line 2: as discussed, move host under metrics and expose collection_interval and initial_delay

internal/collector/otelcol.tmpl
```diff
@@ -21,22 +11,27 @@ receivers:
       memory:
       network:
       paging:
-      processes:
```
oliveromahony, line 24: can you add back processes and process?

internal/collector/otelcol.tmpl
```diff
@@ -21,22 +11,27 @@ receivers:
...
+      grpc:
+      {{- else }}
+      http:
+      {{- end }}
+        endpoint: "{{- .Server.Host -}}:{{- .Server.Port -}}"
+{{- end }}
+{{- range .Receivers.NginxReceivers }}
+  nginx/{{- .InstanceID -}}:
+    endpoint: "{{- .StubStatus -}}"
     collection_interval: 10s
```
oliveromahony, line 28: can we expose this collection interval?

internal/collector/otel_collector_plugin_test.go
```diff
@@ -33,13 +39,282 @@ func TestInitAndClose(t *testing.T) {
...
+		existingReceivers  config.Receivers
+		expectedReceivers  config.Receivers
+	}{
+		{
+			name: "Test 1: Existing NGINX Receiver",
+			nginxConfigContext: &model.NginxConfigContext{
+				InstanceID: "123",
+				StubStatus: "http://new-test-host:8080/api",
+			},
+			existingReceivers: config.Receivers{
```
aphralG, line 202: Could these be added to the protos to be reused ?

internal/collector/otel_collector_plugin_test.go
```diff
@@ -33,13 +39,282 @@ func TestInitAndClose(t *testing.T) {
...
+			name: "Test 1: NGINX Plus receiver",
+			message: &bus.Message{
+				Topic: bus.NginxConfigUpdateTopic,
+				Data: &model.NginxConfigContext{
+					InstanceID: "123",
+					PlusAPI:    fmt.Sprintf("%s/api", nginxPlusMock.URL),
+				},
+			},
+			receivers: config.Receivers{
+				OtlpReceivers: []config.OtlpReceiver{
```
aphralG, line 81: Could these be added to the protos to be reused ?

internal/collector/otel_collector_plugin.go
```diff
@@ -117,19 +119,182 @@ func (oc *Collector) Close(ctx context.Context) error {
...
-	return []string{}
+	return []string{
+		bus.NginxConfigUpdateTopic,
+	}
+}
+
+func (oc *Collector) handleNginxConfigUpdate(ctx context.Context, msg *bus.Message) {
+	nginxConfigContext, ok := msg.Data.(*model.NginxConfigContext)
+	if !ok {
+		slog.ErrorContext(ctx, "Unable to cast message payload to *model.NginxConfigContext", "payload", msg.Data)
```
olli-holmala, line 167: Probably need a `return` here too.
```suggestion
		slog.ErrorContext(ctx, "Unable to cast message payload to *model.NginxConfigContext", "payload", msg.Data)
```
```suggestion
		slog.ErrorContext(ctx, "Unable to cast message payload to *model.NginxConfigContext", "payload", msg.Data)
		
		return
```

internal/collector/otel_collector_plugin.go
```diff
@@ -62,26 +66,24 @@ func (oc *Collector) Init(ctx context.Context, mp bus.MessagePipeInterface) erro
...
+	}
 
 	return nil
 }
 
 func (oc *Collector) bootup(ctx context.Context) error {
-	oc.appDone = make(chan struct{})
 	errChan := make(chan error)
 
+	oc.stopped = false
```
olli-holmala, line 80: Just a side note, should we toggle this to `true` when we receive an error from `oc.service.Run(ctx)` and/or when state is `otelcol.StateClosed` in the for loop below?

internal/collector/otel_collector_plugin.go
```diff
@@ -117,19 +119,182 @@ func (oc *Collector) Close(ctx context.Context) error {
 	slog.InfoContext(ctx, "Closing OTel Collector plugin")
 
 	if !oc.stopped {
+		slog.InfoContext(ctx, "Shutting down OTel Collector", "state", oc.service.GetState())
 		oc.stopped = true
```
olli-holmala, line 123: Just in the interest of correctness, should we toggle this only when state is `otelcol.StateClosed` in our backoff function?

olli-holmala (approved):

aphralG (approved):

craigell (approved):

PR #743, opened 2024-07-10
Fix data plane response messages and errors

aphralG:
### Proposed changes

- Cleaned up data plane response messages so they are consistent and no longer contain the instanceID
- Errors are cleaned up so Config apply original error is sent with failure status instead of the rollback error, rollbacks status is also included in the message for a config apply failure if rollback has occurred 
- Added additional checks to the unit tests to check the errors being sent in the data plane responses instead of just checking the topic

### Checklist

### DIFF
--- a/internal/file/file_plugin.go
+++ b/internal/file/file_plugin.go
@@ -288,7 +288,7 @@ func (fp *FilePlugin) handleConfigUploadRequest(ctx context.Context, msg *bus.Me
 func (fp *FilePlugin) createDataPlaneResponse(correlationID string, status mpi.CommandResponse_CommandStatus,
 	message, instanceID, err string,
 ) *mpi.DataPlaneResponse {
-	response := &mpi.DataPlaneResponse{
+	return &mpi.DataPlaneResponse{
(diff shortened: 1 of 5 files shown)


dhurley (approved):

internal/file/file_plugin.go
```diff
@@ -301,8 +297,13 @@ func (fp *FilePlugin) createDataPlaneResponse(correlationID string, status mpi.C
 		CommandResponse: &mpi.CommandResponse{
 			Status:  status,
 			Message: message,
-			Error:   err,
 		},
 		InstanceId: instanceID,
 	}
+
+	if err != "" {
+		response.GetCommandResponse().Error = err
```
oliveromahony, line 305: this seems unusual. Why is this being unset here?

aphralG: If there is an error it is being set in the response, can switch back to always setting it but seemed a bit odd to set the err if there was none

aphralG: Reverted the change

oliveromahony (approved):

PR #757, opened 2024-07-25
Add docs-build-push github workflow

nginx-jack:
### Proposed changes
The NGINX docs site has migrated to self hosting. This PR adds support for a workflow to build, push, and provide PR previews. See: https://github.com/nginxinc/docs-actions for details.

### Checklist


.github/workflows/docs-build-push.yml
```diff
@@ -0,0 +1,33 @@
...
+        - prod
+  pull_request:
+    branches:
+      - "*"
+    paths:
+      - site/**
+
+jobs:
+  call-docs-build-push:
+    uses: nginxinc/docs-actions/.github/workflows/docs-build-push.yml@main
```
oliveromahony, line 23: I think the best practice is to put a commit sha here? Are you happy not to pin to a particular version?

nginx-jack: Yeah a commit sha here would be better. I left it at `main` since it would let us bump the theme version without creating a PR across all the repos with docs ( https://github.com/nginxinc/docs-actions/blob/03a9a3808fcb77cd0c19d7fa5d59b25565dd1d6d/.github/workflows/docs-build-push.yml#L42). 
I'll update this to a commit sha, and look at storing the theme version externally so we can still update it cleanly.

oliveromahony (approved):

PR #762, opened 2024-07-31
Introduce the NGINX OSS receiver

olli-holmala:
### Proposed changes

The NGINX OSS Opentelemetry receiver is a modified version of the OTel authors' NGINX receiver (see [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/nginxreceiver)) that also includes access log scraping.

This iteration only has the one new metric, `nginx.http.response.status`, with further metrics being introduced in subsequent PRs.

In the OTel collector config, the NGINX receiver is configured like so:
```yaml
nginx:
  endpoint: "http://localhost:80/status"
  collection_interval: 10s
  access_logs:
      - log_format: "$remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\" \"$http_user_agent\" \"$http_x_forwarded_for\"\"$upstream_cache_status\""
        file_path: "/var/log/nginx/access-custom.conf"
```

In `nginx-agent.conf`, NGINX OSS receivers are configured in the following manner:
```yaml
collector:
  receivers:
    nginx_receivers:
        - instance_id: cd7b8911-c2c5-4daf-b311-dbead151d938
          stub_status: "http://localhost:4321/status"
          access_logs:
            - file_path: "/var/log/nginx/access.log"
              log_format: "$remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\" \"$http_user_agent\" \"$http_x_forwarded_for\"\"$upstream_cache_status\""
```

### Checklist


internal/collector/otelcol.tmpl
```diff
@@ -28,6 +28,15 @@ receivers:
...
     collection_interval: 10s
+    {{- if eq (len .AccessLogs) 1 }}
+    {{- range .AccessLogs }}
+    access_log_format: "{{- .LogFormat -}}"
+    operators:
+      - type: file_input
+        include:
+          - "{{- .FilePath -}}"
+    {{- end }}
+    {{- end }}
```
olli-holmala, line 39: So this is limited to 1 access log for the moment because this NGINX OSS receiver was originally written with the assumption that we'd use the out-of-the-box `file_input` operator (see [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/file_input.md)), which consumes either one log file or a list of log files and then pushes them up to a `stanza` pipeline (see [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/stanza)).

The issue with the above approach is that we cannot pass a log format for each access log file, meaning we the log entries from ALL access log files to the same pipeline and only apply the log format at the other end of the pipeline. At that point we have no way of knowing which entry should use which log format.

So I think we will need to write our own modified `file_input` operator that accepts a log format to circumvent the this problem.

oliveromahony (approved):

aphralG (approved):

PR #784, opened 2024-08-13
Add automatic changelog generation in release workflow

spencerugbo:
### Proposed changes

This PR aims to add automatic changelog generation whenever the release workflow is ran.

### Checklist


dhurley (commented):
Since we are always creating a new commit for the updated changelog for each release, then we will always have to create a PR to merge back into main. At the top of this file there is an input called `createPullRequest`, which we use to determine if we want to automatically create a PR back to main or not. I think you can remove this input and have it that the workflow always creates a PR.

dhurley (approved):

aphralG (approved):

PR #823, opened 2024-09-04
[DRAFT] dont run validation or reload when status is RollbackRequired

sean-breen:
### Checklist

### DIFF
--- a/internal/file/file_plugin.go
+++ b/internal/file/file_plugin.go
@@ -198,14 +198,30 @@ func (fp *FilePlugin) handleConfigApplyRequest(ctx context.Context, msg *bus.Mes
 		)
 
 		fp.messagePipe.Process(ctx, &bus.Message{Topic: bus.DataPlaneResponseTopic, Data: response})
-		fp.handleConfigApplyFailedRequest(ctx, &bus.Message{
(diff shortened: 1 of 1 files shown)


internal/file/file_plugin.go
```diff
@@ -198,14 +198,31 @@ func (fp *FilePlugin) handleConfigApplyRequest(ctx context.Context, msg *bus.Mes
...
+				mpi.CommandResponse_COMMAND_STATUS_FAILURE,
+				"Config apply failed, rollback failed",
+				configApplyRequest.GetOverview().GetConfigVersion().GetInstanceId(),
+				err.Error(),
+			)
+			fp.messagePipe.Process(ctx, &bus.Message{Topic: bus.DataPlaneResponseTopic, Data: response})
+			return
+		}
+
+		response = fp.createDataPlaneResponse(
```
sean-breen, line 216: Sending a `...STATUS_FAILURE` response here with no `error` field seems wrong, but it is to indicate that the rollback was successful

aphralG: the original error that caused the config apply to fail could be sent here possibly, but not sure what others think

PR #831, opened 2024-09-10
Update host metrics receiver

RRashmit:
### Proposed changes

Introduced the option for the users to set the scraper for the Host metrics receiver from the nginx agent conf. 

image

### Checklist

### DIFF
--- a/internal/collector/otelcol.tmpl
+++ b/internal/collector/otelcol.tmpl
@@ -21,7 +21,7 @@ receivers:
       network:
       {{- end }}
       {{- end }}
-    {{- end }}  
+{{- end }}  
 {{- range $index, $otlpReceiver := .Receivers.OtlpReceivers }}
   otlp/{{$index}}:
     protocols:
--- a/internal/collector/otelcol.tmpl
+++ b/internal/collector/otelcol.tmpl
@@ -3,14 +3,10 @@ receivers:
   hostmetrics:
     collection_interval: {{ .Receivers.HostMetrics.CollectionInterval }}
     initial_delay: {{ .Receivers.HostMetrics.InitialDelay }}
-    scrapers:
-      cpu:
-      disk:
-      load:
-      filesystem:
-      memory:
-      network:
-      paging:
+    scrapers: 
+    {{- range $key, $value := .Receivers.HostMetrics.Scrapers }}
+      {{ $key }}:
+    {{- end }}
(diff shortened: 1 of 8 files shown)


dhurley:
Might be good as well to add an example grafana dashboard for host metrics to the mock collector. Similar to what we did for NGINX OSS & Plus (https://github.com/nginx/agent/tree/v3/test/mock/collector/grafana/provisioning/dashboards). It would be useful for testing and demos.

internal/config/types.go
```diff
@@ -136,8 +136,9 @@ type (
 	}
 
 	HostMetrics struct {
-		CollectionInterval time.Duration `yaml:"-" mapstructure:"collection_interval"`
-		InitialDelay       time.Duration `yaml:"-" mapstructure:"initial_delay"`
+		Scrapers           map[string]struct{} `yaml:"-" mapstructure:"scrapers"`
```
dhurley, line 139: Instead of a map could we do the following instead:
```
Scrapers struct {
    Cpu CpuScraper `yaml:"-" mapstructure:"cpu"`
    ...
} 

CpuScraper struct {}
```
This way each scraper has its own struct (which would be an empty struct) so that in the future we can add additional configuration if needed. Like the ones documented here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/README.md

RRashmit: I see, these scrapers can also have the additional config, missed to see that before. This approach definitely makes sense.

internal/collector/otelcol.tmpl
```diff
@@ -4,14 +4,24 @@ receivers:
...
+      {{- if .Receivers.HostMetrics.Scrapers.Memory }}
       memory:
+      {{- end }}
+      {{- if .Receivers.HostMetrics.Scrapers.Network }}
       network:
-      paging:
-{{- end }}
+      {{- end }}
+      {{- end }}
+    {{- end }}
```
sean-breen, line 24: Indentation is a little off
```suggestion
{{- end }}  
```

RRashmit: Will check if putting a yaml linter could help us catching these early.

craigell (approved):

sean-breen (approved):

dhurley (approved):

PR #866, opened 2024-09-25
Publish package for agent

RRashmit:
### Proposed changes

With this change a package will be published every time a commit is merged in the `v3` branch. 

The value of the build number is dynamic which increments each time the workflow is run.

GITHUB_RUN_NUMBER : It's part of the default environment variables that GitHub sets are available to every step in a workflow. A unique number for each run of a particular workflow in a repository. This number begins at 1 for the workflow's first run, and increments with each new run. This number does not change if you re-run the workflow run. For example, 3.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -332,16 +332,30 @@ jobs:
       - name: Push load test result
         if:  ${{ success() && github.ref_name == 'v3'}}
         run: git push 'https://github-actions:${{ secrets.GITHUB_TOKEN }}@github.com/nginx/agent.git' benchmark-results:benchmark-results
+  publish-packages-vars:
+    name: Set workflow variables
+    if: ${{ github.event_name == 'pull_request' &&
+      github.base_ref == 'v3' &&
+      github.event.pull_request.merged == true &&
+      !github.event.pull_request.head.repo.fork }}
+    runs-on: ubuntu-22.04
+    outputs:
+      package_build_num: ${{ steps.get_build_num.outputs.build_num }}
+    steps:
+      - name: Get the build number
+        id: get_build_num
+        run: echo "build_num=${{ github.run_number }}-$(echo ${{ github.sha }} | cut -c1-7)" >> $GITHUB_OUTPUT
+
   publish-packages:
     name: Publish NGINX Agent v3 packages
-    if: ${{ github.event_name =='pull_request' &&
+    if: ${{ github.event_name == 'pull_request' &&
             github.base_ref == 'v3' &&
             github.event.pull_request.merged == true &&
             !github.event.pull_request.head.repo.fork }}
     needs: [ lint, unit-test, performance-tests,
              load-tests, official-oss-image-integration-tests,
              official-plus-image-integration-tests,
-             race-condition-test ]
+             race-condition-test, publish-packages-vars ]
     uses: ./.github/workflows/release-branch.yml
     secrets: inherit
     permissions:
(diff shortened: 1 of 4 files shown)


RRashmit:

.github/workflows/ci.yml
```diff
@@ -334,3 +332,20 @@ jobs:
       - name: Push load test result
         if:  ${{ success() && github.ref_name == 'v3'}}
         run: git push 'https://github-actions:${{ secrets.GITHUB_TOKEN }}@github.com/nginx/agent.git' benchmark-results:benchmark-results
+  publish-packages:
+    name: Publish NGINX Agent v3 packages
+    if: ${{ github.event_name =='pull_request' &&
+            github.base_ref == 'v3' &&
+            github.event.pull_request.merged == true &&
+            !github.event.pull_request.head.repo.fork }}
+    needs: [ lint, unit-test, component-test, performance-test,
```
dhurley, line 341: Think you need to remove `component-test`
The pipeline is failing for this:
```
The workflow is not valid. .github/workflows/ci.yml (Line: 341, Col: 31): Job 'publish-packages' depends on unknown job 'component-test'.
```

RRashmit: Good catch, did you run the pipeline manually to view this error?

RRashmit: well, I see that now in the annotations actions

aphralG (approved):

.github/workflows/ci.yml
```diff
@@ -334,3 +332,20 @@ jobs:
...
+             load-tests, official-oss-image-integration-tests,
+             official-plus-image-integration-tests,
+             race-condition-test ]
+    uses: ./.github/workflows/release-branch.yml
+    with:
+      packageVersion: "3.0.0"
+      packageBuildNo: "${{ github.run_number }}"
+      uploadAzure: true
+      publishPackages: true
+      releaseBranch: "v3"
```
dhurley, line 351: I think you also need to add the following to this job so that it has access to the secrets and has the required permissions
```
    secrets: inherit
    permissions:
      contents: read
      id-token: write
```

.github/workflows/ci.yml
```diff
@@ -334,3 +332,20 @@ jobs:
...
+            github.event.pull_request.merged == true &&
+            !github.event.pull_request.head.repo.fork }}
+    needs: [ lint, unit-test, performance-tests,
+             load-tests, official-oss-image-integration-tests,
+             official-plus-image-integration-tests,
+             race-condition-test ]
+    uses: ./.github/workflows/release-branch.yml
+    with:
+      packageVersion: "3.0.0"
+      packageBuildNo: "${{ github.run_number }}"
```
dhurley, line 348: Could we add the sha id to the end of the build number? I would make it easier to track what packages match to what commit. So something like `<run number>.<short sha id>`

.github/workflows/ci.yml
RRashmit: Updated, initial thought was we also add permissions explicitly to the jobs which need them like [upload_packages](https://github.com/nginx/agent/blob/publish_package/.github/workflows/release-branch.yml#L223)

.github/workflows/ci.yml
RRashmit: Updated, this will now have the value like: <run_number>-<commit_id>

dhurley (approved):

PR #873, opened 2024-09-27
Update collector READMEs

CVanF5:
### Proposed changes
Update collector READMEs with correct directives

### Checklist

### DIFF
--- a/internal/collector/nginxossreceiver/README.md
+++ b/internal/collector/nginxossreceiver/README.md
@@ -33,10 +33,10 @@ Example:
 
 ```yaml
 receivers:
-  nginx:
+  nginx_receivers:
     endpoint: "http://localhost:80/status"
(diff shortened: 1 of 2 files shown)


internal/collector/nginxossreceiver/README.md
```diff
@@ -33,10 +33,10 @@ Example:
 
 ```yaml
 receivers:
-  nginx:
+  nginx_receivers:
```
dhurley, line 36: The names here are correct and don't need to be updated.
In the agent conf they are called `nginx_receivers` but when converted to otel configuration it is called `nginx`.
This README is for documenting the otel config for the receiver not the agent config if that makes sense.

CVanF5: I see thanks!

PR #878, opened 2024-09-30
OpenTelemetry: add resource id

sean-breen:
### Proposed changes

Enables the `attribute` processor, which automatically adds the `resource.id` value to every Otel metric message.

The `resource.id` in this case represents the host system the Nginx Agent instance is running on. This id can be used on the management side to group metrics by Agent id.

If another attribute with the same Key is added, ignore it and don't reload the collector configuration.

### Checklist


internal/config/types.go
```diff
@@ -113,7 +113,12 @@ type (
 
 	// OTel Collector Processors configuration.
 	Processors struct {
-		Batch *Batch `yaml:"-" mapstructure:"batch"`
+		Attribute *Attribute `yaml:"-" mapstructure:"attribute"`
```
dhurley, line 116: I think we want to have this more generic so that we can easily add more attribute processors in the future. Also would allow users to add their own ones as well if they want to.

Something like this:

```
Processors struct {
        Attributes []*Attribute `yaml:"-" mapstructure:"attributes"`
	Batch       *Batch     `yaml:"-" mapstructure:"batch"`
}

Attribute struct {
	key string
        action string
        value string
}
```

sean-breen: Went with a slice of new type `Action` instead of a slice of `*Attributes`

craigell (approved):

internal/collector/otelcol.tmpl
```diff
@@ -106,11 +117,11 @@ exporters:
   prometheus:
     endpoint: "{{ .Exporters.PrometheusExporter.Server.Host -}}:{{- .Exporters.PrometheusExporter.Server.Port }}"
 {{- end }}
-{{- if ne .Exporters.Debug nil }}
   debug:
     verbosity: detailed
     sampling_initial: 5
     sampling_thereafter: 200
+{{- if ne .Exporters.Debug nil }}
```
dhurley, line 124: Can you move this if statement back up?

test/mock/collector/nginx-agent.conf
```diff
@@ -55,6 +55,7 @@ collector:
           key: /tmp/key.pem
           generate_self_signed_cert: true
   processors:
+    attribute: {}
```
dhurley, line 58: Dont think you need to specify an empty attribute in the config since you are programmatically adding it

internal/collector/otel_collector_plugin.go
```diff
@@ -225,7 +232,38 @@ func (oc *Collector) handleNginxConfigUpdate(ctx context.Context, msg *bus.Messa
...
 }
 
+func (oc *Collector) handleResourceUpdate(ctx context.Context, msg *bus.Message) {
+	resourceUpdateContext, ok := msg.Data.(*v1.Resource)
+	if !ok {
+		slog.ErrorContext(ctx, "Unable to cast message payload to *v1.Resource", "payload", msg.Data)
+		return
+	}
+
+	if oc.config.Collector.Processors.Attribute != nil { // attribute processor is enabled
```
dhurley, line 242: If `Attribute` is nil then we should create an `Attribute` object and add the action.
We should also check the slice of actions to see if there is already an action for "resource.id" or not

internal/collector/otelcol.tmpl
sean-breen: Done

internal/collector/otel_collector_plugin.go
```diff
@@ -345,6 +394,37 @@ func (oc *Collector) updateExistingNginxOSSReceiver(
 	return nginxReceiverFound, reloadCollector
 }
 
+// nolint: revive
+func (oc *Collector) updateResourceAttributes(
```
dhurley, line 398:
```suggestion
func (oc *Collector) updateAttributeActions(
```

internal/collector/otel_collector_plugin.go
```diff
@@ -345,6 +394,37 @@ func (oc *Collector) updateExistingNginxOSSReceiver(
...
+	actionsToAdd []config.Action,
+) (reloadCollector bool) {
+	reloadCollector = false
+
+	if oc.config.Collector.Processors.Attribute.Actions != nil {
+		for _, toAdd := range actionsToAdd {
+			duplicateKey := false
+			for _, a := range oc.config.Collector.Processors.Attribute.Actions {
```
dhurley, line 406:
```suggestion
			for _, action := range oc.config.Collector.Processors.Attribute.Actions {
```

dhurley (approved):

internal/collector/otel_collector_plugin.go
```diff
@@ -233,7 +240,49 @@ func (oc *Collector) handleNginxConfigUpdate(ctx context.Context, msg *bus.Messa
...
+		return
+	}
+
+	if oc.config.Collector.Processors.Attribute == nil {
+		oc.config.Collector.Processors.Attribute = &config.Attribute{
+			Actions: make([]config.Action, 0),
+		}
+	}
+
+	var reloadCollector bool
```
aphralG, line 256: Just formatting, could this be moved to the top of the function ?

internal/collector/otel_collector_plugin.go
```diff
@@ -233,7 +240,49 @@ func (oc *Collector) handleNginxConfigUpdate(ctx context.Context, msg *bus.Messa
...
+	if oc.config.Collector.Processors.Attribute != nil {
+		if resourceUpdateContext.GetResourceId() != "" {
```
aphralG, line 258: Could these two be in one if statement ? 
```suggestion
		if oc.config.Collector.Processors.Attribute != nil && resourceUpdateContext.GetResourceId() != "" {
```

internal/collector/otel_collector_plugin.go
sean-breen: Done!

internal/collector/otel_collector_plugin.go
sean-breen: Done!

aphralG (approved):

RRashmit (approved):

PR #912, opened 2024-10-24
Config Apply - Determine File Diff

aphralG:
### Proposed changes

- Config Apply now calculates the file diff using the file overview to determine which action needs to be preformed on a file 
  - If file is in the FileOverview but not in the files on disk - file has been added 
  - If file is not in the FileOverview but is on disk - file has been deleted 
  - If file is in FileOverview and is on disk, check hash of file, if has is different - file has been updated

- Config Apply Successful & Config Apply Complete in the file plugin now as well as clearing the cache, sends the final data plane response. This was done so the data plane response can be used to indicate that every step of the config apply has been completed and the next config apply can start (to be done in future PR). 

- Renamed `RollbackCompleteTopic` to `ConfigApplyCompleteTopic` so it can be used to indicated when config apply is done but hasn't been successful so the watchers can be unlocked. 
- Added DELETE ` /api/v1/responses` to mock management plane to clear the responses during tests

### Checklist


internal/bus/topics.go
```diff
@@ -19,6 +19,6 @@ const (
 	WriteConfigSuccessfulTopic = "write-config-successful"
 	ConfigApplySuccessfulTopic = "config-apply-successful"
 	ConfigApplyFailedTopic     = "config-apply-failed"
-	RollbackCompleteTopic      = "rollback-complete"
+	ConfigApplyCompleteTopic   = "config-apply-complete"
```
RRashmit, line 22: wondering if adding `rollback` in the name make it better?

RRashmit (approved):

internal/bus/topics.go
aphralG: The naming was changed as it’s no longer just used in Rollback, it is one of two ways to clear the cache, and restart the watchers which are paused during config apply. There were scenarios where config apply is done but it wasn’t a rollback or successful scenario. Example would be config apply failing before every writing files like an error when determining the file actions. In this case no rollback is required, config apply wasn’t successful but the config apply is done.

RRashmit: I see, thanks for the example.

dhurley (approved):

PR #956, opened 2024-12-27
Add a `nginxprocess` package for working with NGINX processes

ryepup:
### Proposed changes

Add a new `nginxprocess` package to export useful utilities around working with NGINX processes, and update internals to use that package. [NGINXaaS for Azure] has some proprietary code that also wants to use these utilities.

This comes with a significant speed improvement by filtering for NGINX processes earlier and fetching less data from the OS. See individual commits for more details.

[NGINXaaS for Azure]: https://docs.nginx.com/nginx-for-azure/

### Checklist


ryepup:

ryepup:
I'm not sure why the pipeline is failing; `make unit-test` and `make performance-test` work if I run them locally, and I didn't change the otel collector section.

internal/watcher/health/nginx_health_watcher_operator.go
```diff
@@ -39,7 +36,7 @@ func (nhw *NginxHealthWatcher) Health(ctx context.Context, instance *mpi.Instanc
 	}
 
 	proc, err := nhw.processOperator.Process(ctx, instance.GetInstanceRuntime().GetProcessId())
-	if errors.Is(err, process.ErrorProcessNotRunning) {
+	if nginxprocess.IsNotRunningErr(err) {
```
ryepup, line 39: I'm matching the existing behavior here, but there's a slim chance that `instance.GetInstanceRuntime().GetProcessId()` is a running process but not an NGINX process. Should we add another branch using `nginxprocess.IsNotNginxErr` to catch that case?

pkg/nginxprocess/process.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (f optionFunc) apply(o *options) { f(o) }
+
+// WithStatus runs an additional lookup to load the process status.
+func WithStatus(v bool) Option { //nolint:ireturn // functional options can be opaque
+	return optionFunc(func(o *options) { o.loadStatus = v })
+}
+
+func convert(ctx context.Context, p *process.Process, o options) (*Process, error) {
+	if ctx.Err() != nil { // fail fast if we've canceled
```
nginx-nickc, line 60: nit,


```
if err := ctx.Err(); err != nil {
	return err
}

```

pkg/nginxprocess/process.go
```diff
@@ -0,0 +1,142 @@
...
+	if o.loadStatus {
+		flags, _ := p.StatusWithContext(ctx) // slow: shells out to ps
+		status = strings.Join(flags, " ")
+	}
+
+	// unconditionally run fast lookups
+	var created time.Time
+	if millisSinceEpoch, err := p.CreateTimeWithContext(ctx); err == nil {
+		nanos := millisSinceEpoch * int64(time.Millisecond)
+		created = time.Unix(0, nanos)
```
nginx-nickc, line 85: confused, why not `time. UnixMilli(millisSinceEpoch)`, https://pkg.go.dev/time#UnixMilli

pkg/nginxprocess/process.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, opt := range opts {
+		opt.apply(&o)
+	}
+	processes, err := process.ProcessesWithContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range processes {
+		pr, cerr := convert(ctx, p, o)
+		if errors.Is(cerr, errNotAnNginxProcess) {
```
nginx-nickc, line 113: use the new func for consistency? `IsNotNginxErr`

pkg/nginxprocess/process.go
```diff
@@ -0,0 +1,142 @@
...
+// IsHealthy uses Status flags to judge process health. Only works on processes created using [WithStatus].
+func (p *Process) IsHealthy() bool {
+	return p.Status != "" && !strings.Contains(p.Status, process.Zombie)
+}
+
+type options struct {
+	loadStatus bool
+}
+
+// Option customizes how processes are gathered from the OS.
```
nginx-nickc, line 47: Maybe learning for me, the `interface` for `apply` seems extraneous to me. Why not just 

```
type Option func(opts *options)

func WithStatus(v bool) Option {
	return func(opts *options) {
		opts.loadStatus = v
	}
}
```

pkg/nginxprocess/process.go
ryepup: done

pkg/nginxprocess/process.go
ryepup: done

pkg/nginxprocess/process.go
ryepup: > why not ...

because I didn't know `UnixMilli` existed 😄  

done

pkg/nginxprocess/process.go
ryepup: this is the "functional options" implementation we settled on NGINXaaS. It does take more code, but sets a stronger boundary between exported and unexported symbols. This is a similar approach used by [opentelemetry](https://github.com/open-telemetry/opentelemetry-go/blob/c862d2d69cae9f4af2fe8ab244a3b5054d223b55/trace/config.go#L178).

The downside of `type Option func(opts *options)` is that it exposes the unexported `options` type as part of the exported `Option` type. There's not much a caller can do with that information, but it's exposing details we might want to change in the future without having to think about compatibility.

The other (more hypothetical) benefit is flexibility. In the unlikely event this package grows a lot of more features, we could add another option interface (e.g maybe we split into `FindOption` and `ListOption`). `WithStatus` could return a type that implements two interfaces, but that's not possible with `func` types.

If this code were in an `internal` package, I'd probably just use `type Option func(opts *options)`, as all the callers are known. Since this is a public codebase in a package meant to be consumed by other software, I figured I'd err on the side of paranoia.

ryepup:
> I'm not sure why the pipeline is failing; `make unit-test` and `make performance-test` work if I run them locally, and I didn't change the otel collector section.

`unit-test` fixed itself, and the `performance-test` was my fault; I added a benchmark that failed if there were no nginx processes. I switched that up so hopefully it'll be happy now.

pkg/nginxprocess/process.go
```diff
@@ -0,0 +1,140 @@
...
+	Name    string
+	Cmd     string
+	Exe     string // path to the executable
+	Status  string // process status, only present if this process was created using [WithStatus]
+	PID     int32
+	PPID    int32 // parent PID
+}
+
+// IsWorker returns true if the process is a NGINX worker process.
+func (p *Process) IsWorker() bool { return strings.Contains(p.Cmd, "worker") }
```
ryepup, line 29: this isn't good enough, I can get false positives on the master process using `nginx -c /tmp/workers.conf`

ryepup: got this fixed

pkg/nginxprocess/process_test.go
```diff
@@ -0,0 +1,180 @@
...
+	for status, want := range testcases {
+		t.Run(fmt.Sprintf("status=%q", status), func(t *testing.T) {
+			t.Parallel()
+			p := nginxprocess.Process{Status: status}
+			require.Equal(t, want, p.IsHealthy())
+		})
+	}
+}
+
+func BenchmarkList(b *testing.B) {
```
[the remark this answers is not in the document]
ryepup: This benchmark was useful for development to compare with the existing implementation and verify how expensive additional `gopsutil` calls were. I can see how it might mess up CI due to randomness of the runners.

Which would you prefer:

1. delete it
2. put a `b.Skipf` at the top that devs can comment out if they're working around here
3. put a `if os.Getenv("CI") != "" { b.Skipf(...) }` at the top

pkg/nginxprocess/process_test.go
```diff
@@ -0,0 +1,180 @@
...
+	require.True(t, nginxprocess.IsNotNginxErr(err))
+}
+
+func TestProcess_IsNginxWorker(t *testing.T) {
+	t.Parallel()
+	testcases := map[string]struct {
+		cmd  string
+		want bool
+	}{
+		"nginx_worker": {
```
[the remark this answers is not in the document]
ryepup: Will do.

pkg/nginxprocess/process_test.go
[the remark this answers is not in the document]
dhurley: It they were useful for you during development then maybe we should keep them and put a `b.Skipf` in. Maybe just add a comment as well to the function to say why they are skipped.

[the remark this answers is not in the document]
ryepup: how does that look?

ryepup:
Did a rebase, looks like there's some test flake around the otel collector? `panic: close of closed channel`

ryepup:
Did another rebase, now getting a different seemingly unrelated test failure with `TestConvertToStructs`. I think that one will fail occasionally because `range` over a `map` is non-deterministic, and the test is expecting it to get keys in a certain order.

ryepup:
opened #968 to fix the test thats broken my latest pipeline

dhurley (approved):

aphralG (approved):

RRashmit (approved):

PR #958, opened 2025-01-07
Add token path to Agent config Command block

sean-breen:
### Proposed changes

Allow the data plane key to be loaded from a file by adding a new configuration option `tokenpath` under the Command section of the agent config file. Also refactored some code to create new functions `addTransportCredentials` and `validateTokenFile` which are used when creating the gRPC credentials.

- If both `token` and `tokenpath` are set: 
  - `tokenpath` will take priority
  - `token` field will be used as a fallback when the `tokenpath` file does not exist or is inaccessible

```
command:
  server:
    host: "127.0.0.1"             # Command server host
    port: 8888                    # Command server port
  tls:
    server_name: test-local-server
    ca: /tmp/ca.pem
    cert: /tmp/cert.pem
    key: /tmp/key.pem
    generate_self_signed_cert: true
  auth:
    token: "1a2b3c4d"    
    tokenpath: "path/to/secret-token"    # Path to a file containing the authentication token for the command sever        
```

### Checklist


internal/grpc/grpc.go
```diff
@@ -208,34 +209,83 @@ func GetDialOptions(agentConfig *config.Config, resourceID string) []grpc.DialOp
...
+func addTransportCredentials(agentConfig *config.Config, opts []grpc.DialOption) ([]grpc.DialOption, bool) {
 	transportCredentials, err := getTransportCredentials(agentConfig)
-	if err == nil {
-		slog.Debug("Adding transport credentials to gRPC dial options")
-		opts = append(opts,
-			grpc.WithTransportCredentials(transportCredentials),
-		)
-	} else {
+	if err != nil {
+		slog.Error("Unable to add transport credentials to gRPC dial options", "error", err)
```
dhurley, line 224: This error log is misleading since we are adding the default transport credentials. Maybe it should be ```Unable to get transport credentials from agent configuration, adding default transport credentials to gRPC dial options```

internal/grpc/grpc.go
```diff
@@ -208,34 +209,83 @@ func GetDialOptions(agentConfig *config.Config, resourceID string) []grpc.DialOp
...
 		slog.Debug("Adding default transport credentials to gRPC dial options")
```
dhurley, line 225: Maybe remove the debug log

internal/grpc/grpc.go
```diff
@@ -208,34 +209,81 @@ func GetDialOptions(agentConfig *config.Config, resourceID string) []grpc.DialOp
...
+	if path == "" {
+		return "", errors.New("token file path is empty")
+	}
+
+	slog.Debug("Reading token from file", "path", path)
+	var keyVal string
+	keyBytes, err := os.ReadFile(path)
+	if err != nil {
+		slog.Error("Unable to read token from file", "error", err)
+		return "", err
```
dhurley, line 273:
```suggestion
		return "", fmt.Errorf("unable to read token from file %w",  err)
```

internal/grpc/grpc_test.go
```diff
@@ -352,3 +355,90 @@ func Test_ValidateGrpcError(t *testing.T) {
...
+				if err.Error() != tt.wantErrMsg {
+					t.Errorf("retrieveTokenFromFile() error = %v, wantErr %v", err, tt.wantErrMsg)
+				}
+			}
+			assert.Equalf(t, tt.want, got, "retrieveTokenFromFile(%v)", tt.path)
+		})
+	}
+}
+
+func Test_getTransportCredentials(t *testing.T) {
```
dhurley, line 419: Just wondering if the struct array is needed since there is only one test case

sean-breen: Not really needed, but I was going to leave it there in case of expansion in the future.

craigell (approved):

RRashmit (approved):

PR #971, opened 2025-02-04
Remove the use of File Actions during config apply

aphralG:
### Proposed changes

FileAction in the protos will be removed in a future PR. This PR removes the use of the the FileActions from the agent code first. File Actions were only being used during a config apply and rollback.

Created a FileCache which contains the File proto and an Action which is a string for use during config apply replacing the FileAction protos
### Checklist


internal/model/file.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) F5, Inc.
+//
+// This source code is licensed under the Apache License, Version 2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+
+package model
+
+import mpi "github.com/nginx/agent/v3/api/grpc/mpi/v1"
+
+type FileCache struct {
```
dhurley, line 10: Just wondering would `FileAction` be a better name than `FileCache` for this struct?

internal/file/file_manager_service.go
```diff
@@ -37,7 +37,13 @@ import (
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6@v6.8.1 -generate
 //counterfeiter:generate . fileManagerServiceInterface
 
-const maxAttempts = 5
+const (
+	maxAttempts     = 5
+	addAction       = "add"
+	updateAction    = "update"
+	deleteAction    = "delete"
+	unchangedAction = "unchanged"
```
dhurley, line 45: Would it be better to use an enum here and have it defined in the `internal/model/file.go`?

internal/model/file.go
aphralG: I wasn't really sure what to call it as it's not just the file action its used mostly for rollback so went with cache. I'll rename it though

RRashmit (approved):

sean-breen (approved):

PR #975, opened 2025-02-06
Add Exclude Files to File Watcher & Switch to Regex for Exclude Logs

aphralG:
### Proposed changes

- Added Exclude File to FileWacther 
    - Events that trigger from files that match either an exact file path or regex pattern will be ignored
    - Switched from filepath.Walk to filepath.WalkDir which is more efficient
    - Stopped adding a watcher to each file, a watcher is only added to directories 
    - Default is `^.*(\\.log|.swx|~|.swp)$` which excludes all files ending with `.log`, `.swx`, `.swp` or `~`
- Exclude Logs now uses regex or a full file path instead of using shell pattern matching 
     - example:  `^/var/log/nginx/*.log` now needs to be `^/var/log/nginx/.*.log$` to exclude all log files from /var/log/nginx 

### Checklist


RRashmit (approved):

internal/config/config.go
```diff
@@ -110,6 +110,7 @@ func ResolveConfig() (*Config, error) {
 
 	slog.Debug("Agent config", "config", config)
 	slog.Info("Enabled features", "features", config.Features)
+	slog.Info("Excluded files from file monitoring", "exclude_files", config.Watchers.FileWatcher.ExcludeFiles)
```
dhurley, line 113:
```suggestion
	slog.Info("Excluded files from being watched for file changes", "exclude_files", config.Watchers.FileWatcher.ExcludeFiles)
```

dhurley (approved):

sean-breen (approved):

PR #996, opened 2025-02-24
Update metric names for Otel conventions

CVanF5:
### Proposed changes

In an effort to be more consistent with the use of "count" and pluralization the following changes were made:

1. Where "count" is used, ensure the type is "gauge" and not "sum"
2. Where "count" is used remove pluralization from the name.
3. Pluralization consistency. If the last word is a plural ensure the type is "sum"
4. Fix unit names where applicable
5. Where the unit is "bytes" remove byte from the name
6. Replace abbreviated TX/RX with words

### Checklist


internal/collector/nginxplusreceiver/documentation.md
```diff
@@ -647,7 +647,7 @@ The total number of Stream byte IO.
 | nginx.byte.io.direction | The direction of byte traffic. | Str: ``RX``, ``TX`` |
```
CodeMonkeyF5, line 647: Can we use non-abbreviated values for `receive` and `transmit` instead? That would then be consistent with the system.* ones with IO.

CodeMonkeyF5:
For the nginx.* the only one I can see having IO with both direction would be the [nginx.http.request.byte.io](http://nginx.http.request.byte.io/) , [nginx.stream.byte.io](http://nginx.stream.byte.io/) and [nginx.http.upstream.peer.byte.io](http://nginx.http.upstream.peer.byte.io/) which have a direction of either RX or TX ? But if the request implies it's always one direction only, think we're ok as it wouldn't need an attribute for direction.  Ideally those metrics would be consistent with the OTel and not use the abbreviated attribute values of RX or TX so instead have it as receive and transmit.

Also, the name containing the word "byte" isn't really something we should use since the Unit (UCUM) describes that. So could go through and revise any that have ".byte" appearing in the name?

internal/collector/nginxplusreceiver/documentation.md
CVanF5: Agreed and done!

CodeMonkeyF5:
LGTM!

sean-breen (approved):

craigell (approved):

PR #1002, opened 2025-02-27
Add stream file transfer

nginx-nickc:
### Proposed changes

Added stream data transfer for handle large files. General idea is to add `*Stream` rpc for existing `GetFile` and `UpdateFile`, where the file content is divided into chunks. MP and agent can agree on a size limit to use this stream transfer mechanism. Smaller files may not make sense for this. 

Added `SendChunkedFile` and `RecvChunkedFile` as implementation. 

### Checklist


api/grpc/mpi/v1/files.proto
```diff
@@ -27,6 +27,47 @@ service FileService {
...
     // Update a file from the Agent to the Server
     rpc UpdateFile(UpdateFileRequest) returns (UpdateFileResponse) {}
+
+    // GetFileStream requests the file content in chunks. MP and agent should agree on size to use stream
+    // vs non-stream. For smaller files, it may be more efficient to not-stream.
+    rpc GetFileStream(GetFileRequest) returns (stream FileDataChunk) {}
+
+    // UpdateFileStream uploads the file content in streams. MP and agent should agree on size to use stream
+    // vs non-stream. For smaller files, it may be more efficient to not-stream.
+    rpc UpdateFileStream(stream FileDataChunk) returns (UpdateFileResponse) {}
```
ryepup, line 37: Is there an ordering requirement to this stream? I.e. the first `FileDataChunk` must be a `FileDataChunkHeader`, and then following `FileDataChunkContent`s must be sorted by `chunk_id` (lowest to highest)?

I assume it's invalid for a stream to:
- contain two `FileDataChunkHeader`s
- have two `FileDataChunkContent`s with same `chunk_id`
- have more/fewer `FileDataChunkContent`s than `FileDataChunkHeader.chunks` 
- have a `FileDataChunkContent` with zero-length `data`

Maybe bulk out the doc comments to describe the assumptions / constraints that can't be expressed by protobuf?

api/grpc/mpi/v1/files.proto
```diff
@@ -27,6 +27,47 @@ service FileService {
...
+        FileDataChunkHeader header = 1;
+        // Chunk data
+        FileDataChunkContent content = 2;
+    }
+}
+
+// Represents a chunked resource Header
+message FileDataChunkHeader {
+    // meta regarding the transfer request
+    mpi.v1.MessageMeta meta = 1;
```
ryepup, line 53: would it make sense to move the `MessageMeta` up to `FileDataChunk`?

nginx-nickc:

api/grpc/mpi/v1/files.proto
nginx-nickc: done

api/grpc/mpi/v1/files.proto
nginx-nickc: done, added above and couple more.

api/grpc/mpi/v1/files.proto
```diff
@@ -27,6 +27,55 @@ service FileService {
...
+// - invalid to contain more or less than one FileDataChunkHeaders
+// - invalid to have FileDataChunkContents before FileDataChunkHeaders
+// - invalid to have more/fewer FileDataChunkContents than FileDataChunkHeader.chunks
+// - invalid to have two FileDataChunkContents with same chunk_id
+// - invalid to have FileDataChunkContent with zero-length data
+// - invalid to have FileDataChunk message without either header or content
+// - hash of the combined contents should match FileDataChunkHeader.file_meta.hash
+// - total size of the combined contents should match FileDataChunkHeader.file_meta.size
+// - chunk_size should be less than the gRPC max message size
+message FileDataChunk {
```
ryepup, line 51: future/out-of-scope for the PR: would it make sense to add some golang funcs in `package v1` to help follow these rules consistently across MPI implementations? Thinking stuff like:

```go

// SendChunkedFile reads the src into [FileDataChunkContent]s, and sends a valid sequence of
// [FileDataChunk]s down the stream.
func SendChunkedFile(
	ctx context.Context,
	header FileDataChunk, 
	src io.Reader, 
	dst grpc.ServerStreamingServer[FileDataChunk],
) error

// RecvChunkedFile receives [FileDataChunkContent]s from the stream and writes the file contents
// to the dst.
func RecvChunkedFile(
	ctx context.Context,
	src grpc.ServerStreamingClient[FileDataChunk],
	dst io.Writer,
) (header FileDataChunk, err error)

```

ryepup (approved):

api/grpc/mpi/v1/files.proto
nginx-nickc: sure, sounds sensible to have common implementations.

dhurley (approved):

api/grpc/mpi/v1/files.proto
nginx-nickc: implemented this, didn't take in the ctx since there's no utility for it.

api/grpc/mpi/v1/file_stream.go
```diff
@@ -0,0 +1,156 @@
...
+	"google.golang.org/grpc"
+)
+
+var (
+	ErrInvalidHeader     = errors.New("invalid header")
+	ErrUnexpectedContent = errors.New("unexpected content")
+	ErrSend              = errors.New("send error")
+	ErrWrite             = errors.New("write error")
+	ErrFailedRead        = errors.New("failed to read")
+)
```
ryepup, line 22: nit: package vars are mutable, and the compiler will happily allow moronic code like `v1.ErrInvalidHeader = errors.New("haha, I just broke all your error handling")`

I'm not sure if this repo has conventions around errors, but we can get compiler guarantees with a little more code.

```suggestion

// StreamError is an error related to streaming files. Use the constants for more specific causes.
type StreamError string

func (s StreamError) Error() string { return string(s) }

const (
	ErrInvalidHeader     StreamError = "invalid header"
	ErrUnexpectedContent StreamError = "unexpected content"
	ErrSend              StreamError = "send error"
	ErrWrite             StreamError = "write error"
	ErrFailedRead        StreamError = "failed to read"
)
```

api/grpc/mpi/v1/file_stream.go
```diff
@@ -0,0 +1,156 @@
...
+	chunkSize := int(header.Header.GetChunkSize())
+	total := int(header.Header.GetFileMeta().GetSize())
+	if chunkSize == 0 || chunkCount == 0 || total == 0 {
+		return fmt.Errorf("%w:  %v", ErrInvalidHeader, header.Header)
+	}
+	if err := dst.Send(&FileDataChunk{
+		Meta:  meta,
+		Chunk: &header,
+	}); err != nil {
+		return fmt.Errorf("%w: %s (header)", ErrSend, err)
```
ryepup, line 42: nit: the `%s` might be losing details about the error that might be relevant for callers. We can retain it using [`errors.Join`](https://pkg.go.dev/errors#Join)

```suggestion
		return errors.Join(fmt.Errorf("%w: header", ErrSend), err)
```

Or we make a bigger change to define an error type that supports [`unwrap`](https://pkg.go.dev/errors#Unwrap)

api/grpc/mpi/v1/file_stream.go
```diff
@@ -0,0 +1,156 @@
...
+			return fmt.Errorf("%w: %s (content)", ErrSend, err)
+		}
+	}
+
+	return nil
+}
+
+// RecvChunkedFile receives [FileDataChunkContent]s from the stream and writes the file contents
+// to the dst.
+func RecvChunkedFile(
```
ryepup, line 77: future: `FileMeta.hash` says it's the sha256 hash; we could check that hash as part of this function; can hook up an `io.Multiwriter` with `sha256.New` to calculate the hash as we write to the `dst`, and then compare hashes once the write is all complete.

I don't think that'll work today because https://github.com/nginx/agent/blob/v3/pkg/files/file_helpers.go#L138 is using md5 hashes, and the `uuid` library does some extra bit twiddling. At some point I imagine those file helpers will need to support `io.Reader`s and streams; right now we'd need to load the large files into memory to calculate the hash for the header.

api/grpc/mpi/v1/file_stream_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+//nolint:gocognit,revive
+func TestSendChunkedFile(t *testing.T) {
+	for desc, td := range map[string]struct {
+		clientFunc  func(cl *v1fakes.FakeFileService_GetFileStreamServer)
+		header      v1.FileDataChunk_Header
+		expectedErr error
+		content     []byte
+	}{
+		"empty": {
```
ryepup, line 42: nit: I believe this project has a test naming convention, want to prefix these test cases with "Test 1:", "Test 2:", etc

api/grpc/mpi/v1/file_stream.go
nginx-nickc: probably still do able. Since the `md5` just uses the `hash.Hash` interface with a `io.Writer`, so I imagine that's what we want to do, instead of reading the whole thing into memory. https://github.com/google/uuid/blob/v1.6.0/hash.go#L49

api/grpc/mpi/v1/file_stream.go
ryepup: TBH there's a lot of bikeshedding possible with golang errors. For N4A we kinda settled on:

- if callers want to handle different kinds of errors, define an error type with a handful of methods to differentiate the kind of error (kinda like [`url.Error`](https://pkg.go.dev/net/url@go1.24.1#Error)) and return that. Often with `Is*Err(error) bool` funcs to make it easier on callers.
- else, use `fmt.Errorf("<what failed>: %w", err)` or `errors.New("<what failed>")`

`fmt.Errorf` sounds good to me here; can always make it more complicated later when there's an actual caller.

nginx-nickc: removed, and use `fmt.Errorf`

api/grpc/mpi/v1/file_stream.go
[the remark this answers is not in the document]
nginx-nickc: done

api/grpc/mpi/v1/file_stream.go
nginx-nickc: resolving, will keep in mind for the future. This requires the following:
- adding a stream hash.Hash for MD5
- use io.Multiwriter with the above to calculate the hash
- compare the hash at the end

api/grpc/mpi/v1/files.proto
```diff
@@ -27,6 +27,55 @@ service FileService {
...
+// For any Stream file transfer, following assumptions should be asserted (by implementation):
+// - invalid to contain more or less than one FileDataChunkHeaders
+// - invalid to have FileDataChunkContents before FileDataChunkHeaders
+// - invalid to have more/fewer FileDataChunkContents than FileDataChunkHeader.chunks
+// - invalid to have two FileDataChunkContents with same chunk_id
+// - invalid to have FileDataChunkContent with zero-length data
+// - invalid to have FileDataChunk message without either header or content
+// - hash of the combined contents should match FileDataChunkHeader.file_meta.hash
+// - total size of the combined contents should match FileDataChunkHeader.file_meta.size
+// - chunk_size should be less than the gRPC max message size
```
sjberman, line 50: Trying to understand where/when the size agreement happens between MP and agent, and what the sequence of events is.

MP is going to send a ConfigApplyRequest with a batch of FileOverviews to agent. Currently, agent then calls GetFile for each of those overviews. What is the new process going to look like?

pkg/files/file_stream.go
```diff
@@ -0,0 +1,148 @@
...
+		if err != nil {
+			return fmt.Errorf("%w: failed to read content", err)
+		}
+
+		if err = validateRecvChunk(chunk, chunkSize, chunkCount-1, i); err != nil {
+			return err
+		}
+		data := chunk.GetContent().GetData()
+		if _, err = dst.Write(data); err != nil {
+			return fmt.Errorf("%w: failed write", err)
```
dhurley, line 118:
```suggestion
			return fmt.Errorf("unable to write chunk id %d: %w", i, err)
```

pkg/files/file_stream.go
```diff
@@ -0,0 +1,148 @@
...
+	dst io.Writer,
+	chunkCount int,
+	chunkSize int,
+	totalSize int,
+) error {
+	// receive and write all content chunks
+	for i := 0; i < chunkCount; i++ {
+		chunk, err := src.Recv()
+		if err != nil {
+			return fmt.Errorf("%w: failed to read content", err)
```
dhurley, line 110:
```suggestion
			return fmt.Errorf("unable to receive chunk id %d: %w", i, err)
```

pkg/files/file_stream.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+	if totalSize > 0 {
+		return fmt.Errorf("unexpected content: unexpected end of content, expected additional %d bytes", totalSize)
+	}
+
+	return nil
+}
+
+func validateRecvChunk(chunk *v1.FileDataChunk, chunkSize, lastChunkIndex, i int) error {
```
dhurley, line 132:
```suggestion
func validateRecvChunk(chunk *v1.FileDataChunk, chunkSize, lastChunkIndex, chunkID int) error {
```

dhurley (commented):
looks good. Just have some minor comments around the error messages

api/grpc/mpi/v1/files.proto
nginx-nickc: should be the same? with `GetFileStream` instead of `GetFile` for bigger files. Size probably should be in the agent sdk.

sjberman: Should the FileOverviews in the ConfigApplyRequest have the file sizes? How will agent know when to call GetFileStream versus GetFile if it doesn't know the file size yet?

Or will agent always start with GetFileStream to receive the FileChunkHeader to then determine if it needs to get all files using GetFilestream or GetFile?

nginx-nickc: yup, so I believe the file `size` is already in the `FileMeta` already. 

```
// Meta information about the file, the name (including path) and hash
message FileMeta {
    // The full path of the file
    string name = 1 [(buf.validate.field).string.prefix = "/"];
    // The hash of the file contents sha256, hex encoded
    string hash = 2;
    // Last modified time of the file (created time if never modified)
    google.protobuf.Timestamp modified_time = 3 [(buf.validate.field).timestamp.lt_now = true];
    // The permission set associated with a particular file
    string permissions = 4 [(buf.validate.field).string.pattern = "0[0-7]{3}"];
    // The size of the file in bytes
    int64 size = 5;
    // additional file information
    oneof file_type {
       CertificateMeta certificate_meta = 6;
    }
}
```

So from spec, the `ConfigApplyRequest` -> `FileOverview` -> `File` (multiples) -> `FileMeta`. For each file, either side can determine whether to use `GetFile` or `GetFileStream`.

sjberman: And the MP needs to somehow know if a call to GetFileStream is the first call of the transaction so it knows to initially return the FileChunkHeader? (I'm guessing using the correlation ID).

sjberman: > either side can determine whether to use GetFile or GetFileStream

How does MP determine this if agent is the one making the call? By setting a size that forces a call to GetFileStream? And this threshold is preset in the agent config?

ryepup: > And the MP needs to somehow know if a call to GetFileStream is the first call of the transaction so it knows to initially return the FileChunkHeader? (I'm guessing using the correlation ID).

I'm not following the "first call of the transaction" bit.

When the agent is pulling files, it's iterating over the `FileOverview` and choosing `GetFile` or `GetFileStream` for each file. If it chose `GetFileStream`, that's a new gRPC stream for each file. When the MP gets the request, it sends a header chunk, then the data chunks, then ends the stream. From the MPs perspective it's a new stream per file.

sjberman: Understood. I had thought it was a single Header request, then all the Content request afterwards, but I misunderstood.

api/grpc/mpi/v1/files.proto
ryepup: Looks good!

> didn't take in the ctx since there's no utility for it.

Sometimes I add one in to allow cancellation, even if there's nothing else in my funcs that want a `ctx`. That makes for cleaner shutdown if the agent/server get stopped in the middle of a transfer, and can use it to smuggle otel span contexts around.

I'll use either `ctx.Err() != nil` checks or a `case ctx.Done()` if I've already got some `select`ing.

dhurley (approved):

aphralG (approved):

PR #1013, opened 2025-03-19
Fix panic when updating NGINX config and fix remotely enabling/disabling the metrics feature

bejjrajesh:
### Proposed changes

Agent currently has the support to enable/disable features dynamically from NIM. This MR contains change to support dynamically enable/disable metrics feature from NIM based on a flag in nms.conf file. When customer flip the flag to false in nms.conf, we remove metrics and metrics-sender features from payload and send an update to agent, when customer enables flag we add metrics and metrics-sender feature in the payload and send an update to agent. 

1. This MR will fix the panic error happen when updating the nginx config concurrently.
image
2.  The metrics_sender plugin is updated with access to current configuration and it's updated to set readyToSend flag when metrics and metrics_sender is present in the agent config and set to false when metrics and metrics_sender is not present in the current agent config. These changes are tested with NIM and working properly.

### Checklist


bejjrajesh:

aphralG (approved):

sean-breen (approved):

dhurley:
@bejjrajesh Can you run the following make commands:
```
make install-tools
make deps
```

These will setup some tools, install git hooks and update the vendor folders.

PR #1057, opened 2025-04-24
Fix Add OTel Receiver After Config Apply & Fix Reload Monitoring Old Log

aphralG:
### Proposed changes

Changes to Config Apply: 
- Moved `nginx_config_parser.go` from `internal/watcher/instance` to `/internal/datasource/config `to allow the resource plugin to also have an `NginxConfigParser` to parse the config
- Moved `NginxPlusRuntimeInfoEqual`, `NginxRuntimeInfoEqual` & `UpdateNginxInstanceRuntime` to `/datasource/proto/instance.go` from` instance_watcher_service.go` so they can be used in the `ResourceService` as well as the `InstanceWatcherService`
- `ConfigApplySuccessfulTopic` now sends `ConfigApplySuccess` which contains a `NginxConfigContext` as well as a `DataPlaneResponse` instead of just a `DataPlaneResponse` 
     ```
         type ConfigApplySuccess struct {
	      ConfigContext     *NginxConfigContext
	      DataPlaneResponse *v1.DataPlaneResponse
- During a ConfigApply the resource service before `Validating` and `Reloading` the config will now parse the NGINX config so it can have access to any error logs that might have been added to monitor. If the validate and reload are successful it sends the  `NginxConfigContext` created during the parse along with `DataPlane Response `
- Added enabled to InstanceWatcher, the instance watcher is now stopped during a config apply to stop the NginxConfigContext not being sent. 
- Add new plus config to integration tests as there was problems being caused by changing the plus API to a Stub Status 

Other changes made: 
- Moved logger creation from app.go to when we are resolving the agent config so allowedDir, enabled features etc are logged to agent log file
- Added logs about shutting down and starting the stub status scraper 
- Removed enabled function from CredentialWatcher





### Checklist


john-david3 (approved):

internal/resource/resource_service.go
```diff
@@ -188,17 +194,45 @@ func (r *ResourceService) ApplyConfig(ctx context.Context, instanceID string) er
...
-		return fmt.Errorf("failed to reload NGINX %w", reloadErr)
+		return nil, fmt.Errorf("failed to reload NGINX %w", reloadErr)
 	}
 
-	return nil
+	return configContext, nil
+}
+
+func (r *ResourceService) ParseConfig(ctx context.Context, instance *mpi.Instance) (*model.NginxConfigContext, error) {
+	slog.DebugContext(ctx, "Parsing NGINX instance config", "", instance.GetInstanceMeta().GetInstanceId())
```
dhurley, line 220:
```suggestion
	slog.DebugContext(ctx, "Parsing NGINX instance config", "instance_id", instance.GetInstanceMeta().GetInstanceId())
```

internal/watcher/instance/instance_watcher_service.go
```diff
@@ -108,20 +114,27 @@ func (iw *InstanceWatcherService) Watch(
...
+			}
 		}
 	}
 }
 
 func (iw *InstanceWatcherService) ReparseConfigs(ctx context.Context) {
 	slog.DebugContext(ctx, "Reparsing all instance configurations")
 	for _, instance := range iw.instanceCache {
-		iw.ReparseConfig(ctx, instance.GetInstanceMeta().GetInstanceId())
+		iw.ReparseConfig(ctx, instance.GetInstanceMeta().GetInstanceId(), &model.NginxConfigContext{})
```
dhurley, line 129: Why do you need to pass an empty NginxConfigContext?

test/integration/grpc_management_plane_api_test.go
```diff
@@ -295,21 +294,35 @@ func TestGrpc_ConfigApply(t *testing.T) {
...
+			newConfigFile,
 			fmt.Sprintf("/mock-management-plane-grpc/config/%s/etc/nginx/nginx.conf", nginxInstanceID),
 			0o666,
 		)
 		require.NoError(t, err)
 
 		performConfigApply(t, nginxInstanceID)
 
-		responses = getManagementPlaneResponses(t, 1)
+		time.Sleep(5 * time.Second)
```
dhurley, line 313: Is this sleep needed?

test/integration/grpc_management_plane_api_test.go
```diff
@@ -295,21 +294,35 @@ func TestGrpc_ConfigApply(t *testing.T) {
...
+		responses = getManagementPlaneResponses(t, 2)
 		t.Logf("Config apply responses: %v", responses)
 
+		t.Logf("Config apply responses more than 1")
```
dhurley, line 317: Is this log statement needed?

test/integration/grpc_management_plane_api_test.go
aphralG: I think so, because the config context is now being sent depending on timing the test can fail randomly as it sometimes gets one response other times gets 2. So I added a sleep so it always get 2 response so it doesnt matter if the config context takes a second to be sent

test/integration/grpc_management_plane_api_test.go
aphralG: No removed

internal/resource/resource_service.go
```diff
@@ -188,17 +194,45 @@ func (r *ResourceService) ApplyConfig(ctx context.Context, instanceID string) er
...
+	if parseErr != nil {
+		slog.WarnContext(
+			ctx,
+			"Unable to parse NGINX instance config",
+			"config_path", instance.GetInstanceRuntime().GetConfigPath(),
+			"instance_id", instance.GetInstanceMeta().GetInstanceId(),
+			"error", parseErr,
+		)
+
+		return nil, parseErr
```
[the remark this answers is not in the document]
aphralG: This error is already formatted in the ApplyConfig function which calls the ParseConfig
```
configContext, parseErr := r.ParseConfig(ctx, instance)
	if parseErr != nil || configContext == nil {
		return nil, fmt.Errorf("failed to parse config %w", parseErr)
```

internal/datasource/config/nginx_config_parser_test.go
```diff
@@ -3,7 +3,7 @@
 // This source code is licensed under the Apache License, Version 2.0 license found in the
 // LICENSE file in the root directory of this source tree.
 
-package instance
+package config
```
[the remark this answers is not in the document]
aphralG: I don't know, I thought config was correct as it was the "datasource" for the config and thought nginx was more logs, process etc but I can change it

sean-breen (approved):

internal/resource/resource_service.go
```diff
@@ -188,17 +194,45 @@ func (r *ResourceService) ApplyConfig(ctx context.Context, instanceID string) er
 		}
 	}
 
+	configContext, parseErr := r.ParseConfig(ctx, instance)
```
dhurley, line 197: Is there a need for the `r.ParseConfig` function? Can you call `r.nginxConfigParser.Parse(ctx, instance)` directly?

dhurley (approved):

internal/resource/resource_service.go
aphralG: Sorry yes, there was more in the ParseConfig function before because the linter was giving out but got removed. Fixing now

PR #1069, opened 2025-04-30
Add env variables for make dev target

john-david3:
### Proposed changes

- Added two environment variables to the `make dev` target
    - One for the manifest file directory path and the other for the collector config path
- Created new default value for the manifest directory
- Removed the manifest file path from `file_manager_service.go`

### Checklist


internal/config/types.go
```diff
@@ -42,6 +42,7 @@ type (
 		Version            string           `yaml:"-"`
 		Path               string           `yaml:"-"`
 		UUID               string           `yaml:"-"`
+		ManifestDir        string           `yaml:"manifest_dir"        mapstructure:"manifest_dir"`
```
dhurley, line 45:
```suggestion
		ManifestDir        string           `yaml:"_"
```

dhurley (approved):

Makefile
```diff
@@ -181,7 +183,7 @@ run: build ## Run code
 
 dev: ## Run agent executable
 	@echo "🚀 Running App"
-	$(GORUN) -ldflags=$(DEBUG_LDFLAGS) $(PROJECT_DIR)/$(PROJECT_FILE)
+	COLLECTOR_CONFIG_PATH=$(COLLECTOR_PATH) MANIFEST_DIR_PATH=$(MANIFEST_DIR) $(GORUN) -ldflags=$(DEBUG_LDFLAGS) $(PROJECT_DIR)/$(PROJECT_FILE)
```
dhurley:
```suggestion
	NGINX_AGENT_COLLECTOR_CONFIG_PATH=$(COLLECTOR_PATH) NGINX_AGENT_MANIFEST_DIR_PATH=$(MANIFEST_DIR) $(GORUN) -ldflags=$(DEBUG_LDFLAGS) $(PROJECT_DIR)/$(PROJECT_FILE)
```

Shouldn't the env variables have the `NGINX_AGENT_` prefix?

john-david3: I might be wrong, but does viper not do this automatically in the `registerFlags` method in `config.go`:

`viperInstance.SetEnvPrefix(EnvPrefix)`

Makefile
```diff
@@ -181,7 +183,7 @@ run: build ## Run code
 
 dev: ## Run agent executable
 	@echo "🚀 Running App"
-	$(GORUN) -ldflags=$(DEBUG_LDFLAGS) $(PROJECT_DIR)/$(PROJECT_FILE)
+	NGINX_AGENT_COLLECTOR_CONFIG_PATH=$(COLLECTOR_PATH) NGINX_AGENT_MANIFEST_DIR=$(MANIFEST_DIR) $(GORUN) -ldflags=$(DEBUG_LDFLAGS) $(PROJECT_DIR)/$(PROJECT_FILE)
```
sean-breen, line 186: I don't like the idea of this being configurable. Wouldn't it be better to have a system of discovery during boot up to set reasonable default values for our paths depending on what OS we are running on?

PR #1077, opened 2025-05-14
Add tokenpath parameter to collector

john-david3:
### Proposed changes

- Add tokenpath parameter to collecter
- Collector now accepts a tokenpath as the value in the `header-setter` parameter of the agent config
- Add unit test for `resolveExtensions` method

### Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -918,7 +918,7 @@ func resolveExtensions() Extensions {
 	}
 
 	if headersSetter != nil {
-		headersSetter.Headers = values(headersSetter.Headers)
+		headersSetter.Headers = updateHeaders(headersSetter.Headers)
 	}
 
 	return Extensions{
@@ -927,7 +927,7 @@ func resolveExtensions() Extensions {
 	}
 }
 
-func values(headers []Header) []Header {
+func updateHeaders(headers []Header) []Header {
 	var err error
 	newHeaders := []Header{}
 


internal/config/config.go
```diff
@@ -871,6 +871,7 @@ func processOtlpReceivers(tlsConfig *OtlpTLSConfig) error {
 func resolveExtensions() Extensions {
 	var health *Health
 	var headersSetter *HeadersSetter
```
dhurley: I think we should add a new FilePath field to the Header struct instead of putting file paths in the Value field.

internal/config/config.go
```diff
@@ -917,12 +917,39 @@ func resolveExtensions() Extensions {
 		}
 	}
 
+	if headersSetter != nil {
+		token, err := getToken(headersSetter)
+		if err != nil {
+			slog.Error("error getting token from config", "error", err)
```
```suggestion
			slog.Error("Unable to get token from collector headers setter config", "error", err)
```

internal/config/config.go
```diff
@@ -917,12 +917,39 @@ func resolveExtensions() Extensions {
...
 	return Extensions{
 		Health:        health,
 		HeadersSetter: headersSetter,
 	}
 }
 
+func getToken(headersSetter *HeadersSetter) (string, error) {
+	var err error
+
+	token := headersSetter.Headers[0].Value
```
dhurley: We have to handle the case where there are more than one header so you will need to loop through the list and check the Value and FilePath for each one

john-david3: If there are multiple, which one do we use?

internal/config/config.go
```diff
@@ -917,12 +917,38 @@ func resolveExtensions() Extensions {
...
+		headersSetter.Headers = values(headersSetter.Headers)
+	}
+
 	return Extensions{
 		Health:        health,
 		HeadersSetter: headersSetter,
 	}
 }
 
+func values(headers []Header) []Header {
```
dhurley: Could you rename the function to something like `updateHeaders`?

aphralG (approved):

PR #1081, opened 2025-05-15
Update linter version and install instructions in README.md

UnwashedMeme:
### Proposed changes

Without these changes I can't develop on v3 reliably; the `make lint unit-test` and the pre-push scripts fail. See the individual commits for more details of what/why they do.

### Checklist

   - This is all non-agent changes (tests, docs)
   - I ran `make test` on Ubuntu 24.04 and doubt anyone has run make test on any of the above listed platforms recently because any distro that has an `/etc/os-release` that doesn't _exactly_ match the test expectation will fail.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -77,8 +77,6 @@ linters-settings:
     field-writes-are-uses: true
     # Treat IncDec statement (e.g. `i++` or `i--`) as both read and write operation instead of just wr◊ite. Default: false
     post-statements-are-reads: true
-    # Mark all exported identifiers as used. Default: true
-    exported-is-used: true
     # Mark all exported fields as used. Default: true
     exported-fields-are-used: false
     # Mark all function parameters as used. Default: true
@@ -770,7 +768,7 @@ linters-settings:
     # Enforce using attributes only (overrides no-mixed-args, incompatible with kv-only). Default: false
     attr-only: false
     # Enforce using methods that accept a context. Default: false
-    context-only: false
+    context: ""
     # Enforce using static values for log messages. Default: false
     static-msg: false
     # Enforce using constants instead of raw keys. Default: false
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The following packages need to be installed:
 ```
 git clone https://github.com/open-telemetry/opentelemetry-collector.git
 cd opentelemetry-collector
-git checkout v0.114.0
+git checkout v0.124.0
 cd cmd/mdatagen
 go install
 ```
--- a/README.md
+++ b/README.md
@@ -13,15 +13,19 @@ NGINX Agent is a companion daemon for your NGINX Open Source or NGINX Plus insta
 - Notifications of NGINX events
 
 ## Development Environment Setup
+
 ### Installing Prerequisite Packages
+
 The following packages need to be installed:
- - make
- - golang (https://go.dev/doc/install)
- - protoc (https://grpc.io/docs/protoc-installation/)
- - mdatagen (There is currently an issue installing mdatagen https://github.com/open-telemetry/opentelemetry-collector/issues/9281. See instructions below for workaround.)
+
+- make
+- golang (<https://go.dev/doc/install>)
+- protoc (<https://grpc.io/docs/protoc-installation/>)
+- mdatagen (There is currently an [issue installing mdatagen](https://github.com/open-telemetry/opentelemetry-collector/issues/9281). See instructions below for workaround.)
 
 #### Workaround to install mdatagen
-```
+
+```console
 git clone https://github.com/open-telemetry/opentelemetry-collector.git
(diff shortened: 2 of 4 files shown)


UnwashedMeme:
recheck

UnwashedMeme:

README.md
```diff
@@ -13,64 +13,81 @@ NGINX Agent is a companion daemon for your NGINX Open Source or NGINX Plus insta
...
+- mdatagen (There is currently an [issue installing mdatagen](https://github.com/open-telemetry/opentelemetry-collector/issues/9281). See instructions below for workaround.)
 
 #### Workaround to install mdatagen
-```
+
+```console
 git clone https://github.com/open-telemetry/opentelemetry-collector.git
 cd opentelemetry-collector
-git checkout v0.114.0
+git checkout v0.124.0
```
puneetsarna, line 31: do we need to checkout the higher version or does the lower one work?

internal/datasource/host/info_test.go
```diff
@@ -526,11 +526,12 @@ func TestInfo_ContainerInfo(t *testing.T) {
 			info := NewInfo()
 			info.mountInfoLocation = mountInfoFile.Name()
 			info.exec = execMock
+			info.osReleaseLocation = "/non/existent"
```
puneetsarna, line 529: Does hardcoding this here mean that this mergeHostandOSinfo function never gets unit tested? https://github.com/nginx/agent/blob/c25bb76b7f2bfd2959ac67d2b3cc5b6dd2b2727a/internal/datasource/host/info.go#L348 🤔

puneetsarna (approved):

internal/datasource/host/info_test.go
puneetsarna: Instead of doing this, is it better if we did:

```
info.osReleaseLocation = osReleaseFile
```

That way we have a mocked OS release file and the code also tests the unexported function? except that `osReleaseFile` also hardcodes stuff so that might be a problem now?

.golangci.yml
```diff
@@ -77,8 +77,6 @@ linters-settings:
     field-writes-are-uses: true
     # Treat IncDec statement (e.g. `i++` or `i--`) as both read and write operation instead of just wr◊ite. Default: false
     post-statements-are-reads: true
-    # Mark all exported identifiers as used. Default: true
```
puneetsarna, line 80: is this dropped since it no longer works with the newer version of the linter?

.golangci.yml
```diff
@@ -770,7 +768,7 @@ linters-settings:
     # Enforce using attributes only (overrides no-mixed-args, incompatible with kv-only). Default: false
     attr-only: false
     # Enforce using methods that accept a context. Default: false
-    context-only: false
```
puneetsarna, line 773: is this change since it no longer works with the newer version of the linter and `context` is the newer field that we should be using?

internal/datasource/host/info_test.go
UnwashedMeme: > Does hardcoding this here mean that this mergeHostandOSinfo function never gets unit tested?

Hmm, I believe yes. I was thinking that this is already true. If you run this on a Mac then `mergeHostAndOsReleaseInfo` will be skipped since that file doesn't exist.

However, I believe the github pipeline _does_ test this right now: [.github/workflows/ci.yml#L37](https://github.com/nginx/agent/blob/v3/.github/workflows/ci.yml#L37)

I suspect you are correct that the _intent_ was to set `info.osReleaseLocation = osReleaseFile`; however this also fails because the hardcoded data in there is for ubuntu 20.04 and the expectation is ubuntu 22.04.

UnwashedMeme: It looks like some of the releasinfo tests were added in https://github.com/nginx/agent/pull/977/files#diff-df29d34cf2163999a2f1dfb2567215ed1d099edf1d93cf4f0afd3a0e809378bc; I suspect the github ci and assertions were already a bit out of sync at that time.

.golangci.yml
UnwashedMeme: Correct, `exported-is-used` is deprecated, it is always true now.

.golangci.yml
UnwashedMeme: correct -- `context` config changed, there are now more options than true or
false. "" disables it to match the previous behavior

(this is also in the commit message 😃)

internal/datasource/host/info_test.go
UnwashedMeme: We talked this out; the point of this test is to gather container ids correctly not test mergeHostAndOsReleaseInfo.  It would be good to add another test that focuses on that, but it is outside the scope here.

puneetsarna: so we skip the test then for now?

aphralG (approved):

PR #1082, opened 2025-05-16
Do not send UpdateOverview after handling a Config Update

CodeMonkeyF5:
Removes unnecessary UpdateOverview as N1 doesn't need to be notified by Agent V3 for an update it just applied.

### Proposed changes

Following a Config Update being applied by Agent V3 successfully, there is no need to send a FileOverview update (N1 already knows it and could be in the process of storing it).

### Checklist


CodeMonkeyF5:

PR #1088, opened 2025-05-23
Fix remotely enabling/disabling features

dhurley:
### Proposed changes

- Add fix that prevents the agent from starting plugins more than once when features are being enabling/disabling remotely
- Fix panic that occur when updating NGINX config
- Add more general logging to all plugins

### Checklist

### DIFF
--- a/src/core/pipe.go
+++ b/src/core/pipe.go
@@ -71,7 +71,7 @@ func (p *MessagePipe) Register(size int, plugins []Plugin, extensionPlugins []Ex
 	pluginsRegistered := []string{}
 	extensionPluginsRegistered := []string{}
 
-	for _, plugin := range p.plugins {
+	for _, plugin := range plugins {
 		for _, subscription := range plugin.Subscriptions() {
 			p.regMu.Lock()
 			err := p.bus.Subscribe(subscription, plugin.Process)
@@ -83,7 +83,7 @@ func (p *MessagePipe) Register(size int, plugins []Plugin, extensionPlugins []Ex
 		pluginsRegistered = append(pluginsRegistered, *plugin.Info().name)
 	}
 
-	for _, plugin := range p.extensionPlugins {
+	for _, plugin := range extensionPlugins {
 		for _, subscription := range plugin.Subscriptions() {
 			p.regMu.Lock()
 			err := p.bus.Subscribe(subscription, plugin.Process)
@@ -94,6 +94,7 @@ func (p *MessagePipe) Register(size int, plugins []Plugin, extensionPlugins []Ex
 		}
 		extensionPluginsRegistered = append(extensionPluginsRegistered, *plugin.Info().name)
 	}
+
 	log.Infof("The following core plugins have been registered: %q", pluginsRegistered)
 	log.Infof("The following extension plugins have been registered: %q", extensionPluginsRegistered)
 
@@ -127,6 +128,8 @@ func (p *MessagePipe) DeRegister(pluginNames []string) error {
 					return err
 				}
 			}
+
+			plugin = nil
 		}
 	}
 
--- a/src/core/pipe.go
+++ b/src/core/pipe.go
@@ -128,8 +128,6 @@ func (p *MessagePipe) DeRegister(pluginNames []string) error {
 					return err
 				}
 			}
-
-			plugin = nil
 		}
 	}
 
--- a/src/plugins/config_reader.go
+++ b/src/plugins/config_reader.go
@@ -152,7 +152,9 @@ func (r *ConfigReader) updateAgentConfig(payloadAgentConfig *proto.AgentConfig) 
 		}
 
 		if synchronizeFeatures {
+			log.Debugf("Agent config features changed, synchronizing features")
 			r.synchronizeFeatures(payloadAgentConfig)
+			r.config.Features = payloadAgentConfig.Details.Features
 		}
 
 		r.messagePipeline.Process(core.NewMessage(core.AgentConfigChanged, payloadAgentConfig))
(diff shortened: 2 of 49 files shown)


src/core/pipe.go
```diff
@@ -127,6 +128,8 @@ func (p *MessagePipe) DeRegister(pluginNames []string) error {
 					return err
 				}
 			}
+
+			plugin = nil
```
aphralG, line 132: Why is setting the plugin to nil needed ?

src/plugins/config_reader.go
```diff
@@ -177,12 +184,15 @@ func (r *ConfigReader) synchronizeFeatures(agtCfg *proto.AgentConfig) {
...
-
 		err := r.messagePipeline.DeRegister([]string{agent_config.FeatureFileWatcher, agent_config.FeatureFileWatcherThrottle})
 		if err != nil {
 			log.Warnf("Error De-registering %v Plugin: %v", data, err)
 		}
-
+	} else if data == agent_config.FeatureMetrics {
+		err := r.messagePipeline.DeRegister([]string{agent_config.FeatureMetrics, agent_config.FeatureMetricsThrottle, agent_config.FeatureMetricsSender})
+		if err != nil {
+			log.Warnf("Error De-registering %v Plugin: %v", data, err)
```
aphralG, line 194:
```suggestion
			log.Warnf("Error deregistering %v plugin: %v", data, err)
```

src/plugins/config_reader.go
```diff
@@ -177,12 +184,15 @@ func (r *ConfigReader) synchronizeFeatures(agtCfg *proto.AgentConfig) {
 
 func (r *ConfigReader) deRegisterPlugin(data string) {
 	if data == agent_config.FeatureFileWatcher {
-
 		err := r.messagePipeline.DeRegister([]string{agent_config.FeatureFileWatcher, agent_config.FeatureFileWatcherThrottle})
 		if err != nil {
 			log.Warnf("Error De-registering %v Plugin: %v", data, err)
```
aphralG, line 189:
```suggestion
			log.Warnf("Error deregistering %v plugin: %v", data, err)
```

src/plugins/events.go
```diff
@@ -64,6 +64,7 @@ func (a *Events) Init(pipeline core.MessagePipeInterface) {
 
 func (a *Events) Close() {
 	log.Info("Events is wrapping up")
+	log.Info("Events is closed")
```
aphralG, line 67: Are both logs needed ?

src/plugins/extensions.go
```diff
@@ -37,6 +37,7 @@ func (e *Extensions) Init(pipeline core.MessagePipeInterface) {
 
 func (e *Extensions) Close() {
 	log.Info("Extensions is wrapping up")
+	log.Info("Extensions is closed")
```
aphralG, line 40: Are both logs needed ?

src/plugins/config_reader.go
```diff
@@ -46,6 +47,7 @@ func (r *ConfigReader) Info() *core.Info {
 
 func (r *ConfigReader) Close() {
 	log.Info("ConfigReader is wrapping up")
+	log.Info("ConfigReader is closed")
```
aphralG, line 50: Are both logs needed ?

src/plugins/features.go
```diff
@@ -94,6 +98,7 @@ func (f *Features) Init(pipeline core.MessagePipeInterface) {
 
 func (f *Features) Close() {
 	log.Info("Features is wrapping up")
+	log.Info("Features is closed")
```
aphralG, line 101: Are both logs needed ?

src/plugins/file_watcher_throttling.go
```diff
@@ -59,12 +59,13 @@ func (fwt *FileWatchThrottle) SetStarted(newValue bool) {
...
 
 func (fwt *FileWatchThrottle) Init(pipeline core.MessagePipeInterface) {
-	fwt.messagePipeline = pipeline
 	log.Info("FileWatchThrottle initializing")
+	fwt.messagePipeline = pipeline
 }
 
 func (fwt *FileWatchThrottle) Close() {
 	log.Info("FileWatchThrottle is wrapping up")
+	log.Info("FileWatchThrottle is closed")
```
aphralG, line 68: Are both logs needed ?

src/plugins/metrics_sender.go
```diff
@@ -60,11 +54,13 @@ func (r *MetricsSender) Info() *core.Info {
 
 func (r *MetricsSender) Process(msg *core.Message) {
 	if msg.Exact(core.AgentConnected) {
-		r.readyToSend.Toggle()
+		log.Debug("Metrics sender received agent connected message")
+		r.readyToSend.Store(true)
```
aphralG, line 58: Why the switch from the toggle ?

src/core/pipe.go
dhurley: I was meant to remove that. Let me update that

src/plugins/events.go
dhurley: I just did it for consistency. Other plugins that actually do something in their close functions have the same messages

src/plugins/metrics_sender.go
dhurley: If we get an AgentConnected message we should set it to true always so I replaced the toggle just in case we received the message when wouldn't set the boolean to false

src/plugins/events.go
aphralG: Since it doesn't need to wrap up anything would just the close log do?

src/plugins/metrics_sender.go
aphralG: OK

src/plugins/events.go
dhurley: yeah thats makes sense. Let me make that update

aphralG (approved):

craigell (approved):

sean-breen (approved):

PR #1094, opened 2025-05-28
Fix race conditions

aphralG:
### Proposed changes

Fix Race conditions caused by 
- OTel collector modifying back off settings 
- Instance Cache being updated 
- Process being converted 
### Checklist

### DIFF
--- a/internal/collector/otel_collector_plugin.go
+++ b/internal/collector/otel_collector_plugin.go
@@ -217,9 +217,9 @@ func (oc *Collector) Close(ctx context.Context) error {
 		oc.service.Shutdown()
 		oc.cancel()
 
-		settings := oc.config.Client.Backoff
+		settings := *oc.config.Client.Backoff
 		settings.MaxElapsedTime = maxTimeToWaitForShutdown
-		err := backoff.WaitUntil(ctx, settings, func() error {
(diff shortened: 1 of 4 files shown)


dhurley (approved):

internal/collector/otel_collector_plugin.go
```diff
@@ -160,11 +160,16 @@ func (oc *Collector) processReceivers(ctx context.Context, receivers []config.Ot
...
 }
 
+// nolint: revive, cyclop
 func (oc *Collector) bootup(ctx context.Context) error {
 	slog.InfoContext(ctx, "Starting OTel collector")
 	errChan := make(chan error)
 
 	go func() {
+		if oc.service == nil {
+			errChan <- fmt.Errorf("unable to start OTel collector: service is nil")
```
sean-breen, line 170: Do we need a `return` here to avoid running the rest of the goroutine in this case?

sean-breen (changes requested):

internal/collector/otel_collector_plugin.go
aphralG: No the errChan will return line 182 
`		case err := <-errChan:
			return err`

PR #1134, opened 2025-06-17
Update file watcher to only monitor directories referenced in the nginx configuration

dhurley:
### Proposed changes

- Update file watcher to only monitor directories referenced in the nginx configuration.
- Ensure that any opened files are closed correctly.

### Checklist

Before creating a PR, run through this checklist and mark each as complete.

- [ ] I have read the [`CONTRIBUTING`](https://github.com/nginx/agent/blob/main/docs/CONTRIBUTING.md) document
- [ ] I have run ```make install-tools``` and have attached any dependency changes to this pull request
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works
- [ ] If applicable, I have checked that any relevant tests pass after adding my changes
- [ ] If applicable, I have updated any relevant documentation ([`README.md`](https://github.com/nginx/agent/blob/main/README.md))
- [ ] If applicable, I have tested my cross-platform changes on Ubuntu 22, Redhat 8, SUSE 15 and FreeBSD 13

PR #1144, opened 2025-06-25
Add validation to allowed directories on config load

sean-breen:
### Proposed changes

Adds some validation of paths when loading the allowed directories. 

Allowed paths must:

- be absolute
- not contain any relative portions (`.`,`..`,etc)
- contain any unsupported characters

Any files not satisfying this criteria will be ignored, and logged at WARN level.

When checking if a path is allowed using `isAllowedDir`, we will now check if the path exists and if it does we will perform a symlink check. Any symlinks will be logged and the function will return false.

### Checklist


internal/config/types.go
```diff
@@ -411,12 +414,63 @@ func (c *Config) AreReceiversConfigured() bool {
 		len(c.Collector.Receivers.TcplogReceivers) > 0
 }
 
-func isAllowedDir(dir string, allowedDirs []string) bool {
+// isAllowedDir checks if the given path is in the list of allowed directories.
+// It returns true if the path is allowed, false otherwise.
+// If the path does not exist, it logs a warning and returns false.
+// It also checks if the path is a file, in which case it checks the directory of the file.
```
aphralG: I think the part about if path doesn't exist might cause issues when this PR goes in https://github.com/nginx/agent/pull/1134. From my understanding a directory that does not exist can be refrenced

sean-breen: Removed that portion so I'll update the comment, thanks!

internal/config/config_test.go
```diff
@@ -158,6 +158,84 @@ func TestNormalizeFunc(t *testing.T) {
...
 }
 
+func TestResolveAllowedDirectories(t *testing.T) {
+	tests := []struct {
+		name           string
+		configuredDirs []string
+		expected       []string
+	}{
+		{
+			name:           "Empty path",
```
dhurley: Can you update the format of the test names to 
```
Test 1: Empty path
```

internal/config/types.go
```diff
@@ -414,16 +417,63 @@ func (c *Config) AreReceiversConfigured() bool {
...
 }
 
-func isAllowedDir(dir string, allowedDirs []string) bool {
-	if !strings.HasSuffix(dir, "/") && filepath.Ext(dir) == "" {
-		dir += "/"
+// isAllowedDir checks if the given path is in the list of allowed directories.
+// It returns true if the path is allowed, false otherwise.
+// If the path is allowed but does not exist, it also logs a warning.
+// It also checks if the path is a file, in which case it checks the directory of the file.
+func isAllowedDir(path string, allowedDirs []string) bool {
```
dhurley: Can you pass the context into this function as well so that we can use it in the log statements?

internal/config/types.go
```diff
@@ -414,16 +417,63 @@ func (c *Config) AreReceiversConfigured() bool {
...
+	if len(allowedDirs) == 0 {
+		slog.Warn("No allowed directories configured")
+		return false
+	}
+
+	directoryPath := path
+	isFilePath, err := regexp.MatchString(`\.(\w+)$`, directoryPath)
+	if err != nil {
+		slog.Error("Error matching path", "path", directoryPath, "error", err)
```
dhurley: Should this be a warning?

internal/config/types.go
```diff
@@ -414,16 +431,62 @@ func (c *Config) AreReceiversConfigured() bool {
...
-func isAllowedDir(dir string, allowedDirs []string) bool {
-	if !strings.HasSuffix(dir, "/") && filepath.Ext(dir) == "" {
-		dir += "/"
+// isAllowedDir checks if the given path is in the list of allowed directories.
+// It returns true if the path is allowed, false otherwise.
+// If the path is allowed but does not exist, it also logs a warning.
+// It also checks if the path is a file, in which case it checks the directory of the file.
+func isAllowedDir(path string, allowedDirs []string) (bool, error) {
+	if len(allowedDirs) == 0 {
+		slog.Warn("No allowed directories configured")
```
dhurley: If you are returning an error then there is no need for this log message as the calling function will probably log the error anyways

internal/config/types.go
```diff
@@ -414,16 +431,62 @@ func (c *Config) AreReceiversConfigured() bool {
...
+		return false, errors.New("no allowed directories configured")
+	}
+
+	directoryPath := path
+	isFilePath, err := regexp.MatchString(`\.(\w+)$`, directoryPath)
+	if err != nil {
+		slog.Error("Error matching path", "path", directoryPath, "error", err)
```
dhurley: Same as comment above

john-david3 (approved):

internal/config/types_test.go
```diff
@@ -6,88 +6,104 @@
...
 		t.Run(test.name, func(t *testing.T) {
-			config.AllowedDirectories = test.allowedDirs
-			result := config.IsDirectoryAllowed(test.fileDir)
+			result, err := isAllowedDir(test.filePath, test.allowedDirs)
+			require.NoError(t, err)
 			assert.Equal(t, test.allowed, result)
 		})
 	}
+
+	t.Run("Test 7: Symlink in allowed directory", func(t *testing.T) {
```
aphralG: Could this be moved to a new test TestTypes_isAllowedDir_symlink or something ?

internal/config/types.go
```diff
@@ -432,16 +449,60 @@ func (c *Config) AreReceiversConfigured() bool {
...
+		return false, errors.New("no allowed directories configured")
+	}
+
+	directoryPath := path
+	isFilePath, err := regexp.MatchString(`\.(\w+)$`, directoryPath)
+	if err != nil {
+		return false, errors.New("error matching path" + directoryPath)
+	}
+
+	if isFilePath {
```
aphralG, line 476: Could a unit test be added for this ?

internal/config/types.go
```diff
@@ -432,16 +449,60 @@ func (c *Config) AreReceiversConfigured() bool {
...
+// It returns true if the path is allowed, false otherwise.
+// If the path is allowed but does not exist, it also logs a warning.
+// It also checks if the path is a file, in which case it checks the directory of the file.
+func isAllowedDir(path string, allowedDirs []string) (bool, error) {
+	if len(allowedDirs) == 0 {
+		return false, errors.New("no allowed directories configured")
+	}
+
+	directoryPath := path
+	isFilePath, err := regexp.MatchString(`\.(\w+)$`, directoryPath)
```
aphralG: Could you add a comment for what the regex is check ? And is there a unit test for this specifically ?

internal/config/types.go
sean-breen: The current unit tests in `TestTypes_isAllowedDir` cover handling file paths and directories

internal/config/types.go
sean-breen: It matches if the `directoryPath` ends with what looks like a file extension. However this may not be exact enough as we will also catch directories that start with a `.`, so I'll make it more explicit

internal/config/types.go
```diff
@@ -322,9 +325,13 @@ type (
 
 func (col *Collector) Validate(allowedDirectories []string) error {
 	var err error
-	cleaned := filepath.Clean(col.ConfigPath)
+	cleanedConfPath := filepath.Clean(col.ConfigPath)
 
-	if !isAllowedDir(cleaned, allowedDirectories) {
+	allowed, err := isAllowedDir(cleanedConfPath, allowedDirectories)
```
aphralG, line 344: Would using `IsDirectoryAllowed(directory string) bool` work here instead ? It will log the error saving an err check ?

internal/config/types.go
```diff
@@ -342,9 +349,13 @@ func (nr *NginxReceiver) Validate(allowedDirectories []string) error {
 	}
 
 	for _, al := range nr.AccessLogs {
-		if !isAllowedDir(al.FilePath, allowedDirectories) {
+		allowed, allowedError := isAllowedDir(al.FilePath, allowedDirectories)
```
aphralG, line 366: Same as above

internal/config/types.go
sean-breen: Since that function is a method on the `Config` object, I would need access to the agents `Config` object from the Collector, this seemed easier

aphralG: Should the IsDirectoryAllowed(directory string) bool be updated to not be part of the Config ? Just wondering if there was a reason it was done that way maybe we werent using it outside of there at the time

sean-breen: Not sure of the historical reasoning, but `IsDirectoryAllowed` is used in several places as it is so I tried to reduce the number of locations where it would be changed.

sean-breen: Happy to discuss further about the design of this but it's a little outside the scope of this item

PR #1149, opened 2025-06-30
Fix config format outputted by preinstall script

dhurley:
### Proposed changes

Fix config format outputted by preinstall script.

### Checklist

### DIFF
--- a/scripts/packages/preinstall.sh
+++ b/scripts/packages/preinstall.sh
@@ -130,7 +130,7 @@ command:
         skip_verify: false
 "
 
-        echo "$v3_config_contents" > "$v3_config_file" \
+        echo -e "$v3_config_contents" > "$v3_config_file" \
             || err_exit "Failed to write v3 config"
     else
         echo "Existing NGINX Agent version is not v2, skipping config migration"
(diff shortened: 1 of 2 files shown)


john-david3 (approved):

scripts/packages/preinstall.sh
```diff
@@ -130,8 +130,13 @@ command:
         skip_verify: false
 "
 
-        echo "$v3_config_contents" > "$v3_config_file" \
-            || err_exit "Failed to write v3 config"
+        if [ -n "$( echo -e )" ]; then
```
sean-breen, line 133: Pretty sure this will only fail when `echo` itself isn't found
```
sh-3.2$ echo -e
-e
sh-3.2$ echo $?
0
```

scripts/packages/preinstall.sh
```diff
@@ -130,8 +130,13 @@ command:
...
+            echo "$v3_config_contents" > "$v3_config_file" \
+                || err_exit "Failed to write v3 config"
```
sean-breen, line 135: I think this `err_exit` condition will never be executed since the echo won't fail even with empty string etc.

sean-breen:
```
sh-3.2$ echo "" > i_dont_exist
sh-3.2$ echo $?
0

aphralG (approved):

PR #1160, opened 2025-07-10
Add extra linters

aphralG:
### Proposed changes
Added linters which were added in v2 of golangci-lint 
Left nolintlint commented out for now and will be done in a future PR as it will require a lot of changes and this PR is already big enough. 

- decorder - Check declaration order and count of types, constants, variables and functions
- funcorder - Checks the order of functions, methods, and constructors.
- intrange - Finds places where for loops could make use of an integer range.
- loggercheck - Checks key value pairs for common logger libraries (kitlog,klog,logr,slog,zap).
- perfsprint - Checks that fmt.Sprintf can be replaced with a faster alternative.
- spancheck - Checks for mistakes with OpenTelemetry/Census spans.
- usetesting - Reports uses of functions with replacement inside the testing package. (ie. `os.MkdirTemp("a", "b") -> t.TempDir()`)

// Commented out for now 
nolintlint - Reports ill-formed or insufficient nolint directives. Also can require a comment as to why it is needed
### Checklist


internal/logger/logger.go
```diff
@@ -157,7 +157,6 @@ func CorrelationIDAttr(ctx context.Context) slog.Attr {
 		slog.DebugContext(
 			ctx,
 			"Correlation ID not found in context, generating new correlation ID",
-			"correlation_id",
```
dhurley, line 160: Were you meant to remove this?

aphralG: Yes logger pointed out we were adding the key twice, in GenerateCorrelationID() we return an slog.Attr which already has a key and a value so when we logged it it didn't need a key to be added

john-david3 (approved):

dhurley (approved):

PR #1167, opened 2025-07-15
Add support for configuring mutliple OTel pipelines

dhurley:
### Proposed changes

- Allow multiple pipelines to be configured
- Add default metrics & log pipelines if a command or auxiliary command server are configured

### Checklist


internal/collector/otel_collector_plugin.go
```diff
@@ -59,7 +59,7 @@ var (
 	initMutex            = &sync.Mutex{}
 )
 
-// NewCollector is the constructor for the Collector plugin.
+// New is the constructor for the Collector plugin.
 func New(conf *config.Config) (*Collector, error) {
```
aphralG: Can you change this NewCollector instead

internal/collector/otelcol.tmpl
```diff
@@ -1,10 +1,11 @@
 receivers:
-{{- if ne .Receivers.ContainerMetrics nil }}
+  {{- with .Receivers.ContainerMetrics }}
```
aphralG: should this change be done everywhere ?

internal/config/config.go
```diff
@@ -137,46 +141,164 @@ func ResolveConfig() (*Config, error) {
...
+		collector.Exporters.OtlpExporters = make(map[string]*OtlpExporter)
+	}
+
+	defaultCommandServer := config.Command
+
+	if config.IsAuxiliaryCommandGrpcClientConfigured() {
+		defaultCommandServer = config.AuxiliaryCommand
+	}
+
+	if _, ok := collector.Exporters.OtlpExporters[DefaultExporter]; !ok && defaultCommandServer != nil {
```
aphralG, line 192: Should these checks be done for addDefaultProcessors & addDefaultPipelines

internal/collector/otelcol.tmpl
dhurley: let me undo that change so that we are consistent

internal/config/config.go
dhurley: The check should just be done for addDefaultPipelines. I'll update that

internal/config/config.go
```diff
@@ -138,46 +142,169 @@ func ResolveConfig() (*Config, error) {
...
-			" Using default collector configuration")
-		defaultCollector(collector, config)
+func defaultCollector(collector *Collector, config *Config) {
+	// Always add default host metric receiver and default processor
+	addDefaultHostMetricsReceiver(collector)
+	addDefaultProcessors(collector)
+
+	// Only add default otlp exporter and pipelines if connected to a management plane
+	if config.IsCommandGrpcClientConfigured() || config.IsAuxiliaryCommandGrpcClientConfigured() {
+		addDefaultOtlpExporter(collector, config)
```
aphralG, line 152: Are we adding the default exporter even if an exporter is already configured ?

dhurley: no there is a check in the addDefaultOtlpExporter function to see if there is a default exporter already set

PR #1169, opened 2025-07-16
Filter NAP logs by Severity

aphralG:
### Proposed changes

Added filter operator to TcplogReceiver for NAP Syslog Servers. 
The priority attribute is divided by 8 and if the remainder is greater than 4 the log is dropped.

```
0 | Emergency: system is unusable
1 | Alert: action must be taken immediately
2 | Critical: critical conditions
3 | Error: error conditions
4 | Warning: warning conditions
5 | Notice: normal but significant condition
6 | Informational: informational messages
7 | Debug: debug-level messages
```

### Checklist


internal/collector/otel_collector_plugin.go
```diff
@@ -563,6 +563,15 @@ func (oc *Collector) updateTcplogReceivers(nginxConfigContext *model.NginxConfig
 								"protocol": "rfc3164",
 							},
 						},
+						// filter drops all logs that have a severity above 4
+						// https://docs.secureauth.com/0902/en/how-to-read-a-syslog-message.html#severity-code-table
+						{
+							Type: "filter",
+							Fields: map[string]string{
+								"expr":       "'attributes.priority % 8 > 4'",
```
CVanF5: This is a reasonable solution given the time constraint, but the drawback is it's still processing the entire syslog message, at least I think so if I'm reading https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/parser/syslog/parser.go#L62 correctly.

If CPU usage is high in Agent when processing a high volume of syslogs, we may need to consider writing a custom operator for the TCP Receiver and adding code to just parse the first few bytes, get the PRI, drop the message if > 4 and then allow the other operators to continue

john-david3 (approved):

dhurley (approved):

PR #1182, opened 2025-07-24
Move `datasource/host` to `pkg`

sean-breen:
### Proposed changes

Moves the `internal/datasource/host` to `pkg` so other teams can use the logic for generating the `resource id`.

### Checklist


internal/grpc/grpc.go
```diff
@@ -88,10 +89,13 @@ func NewGrpcConnection(ctx context.Context, agentConfig *config.Config,
 
 	slog.InfoContext(ctx, "Dialing grpc server", "server_addr", serverAddr)
 
+	var err error
 	info := host.NewInfo()
-	resourceID := info.ResourceID(ctx)
+	resourceID, err := info.ResourceID(ctx)
+	if err != nil {
+		slog.WarnContext(ctx, "Failed to get ResourceID from host info", "error", err.Error())
```
dhurley, line 95:
```suggestion
		slog.WarnContext(ctx, "Failed to get ResourceID from host info", "error", err)
```

pkg/host/info.go
```diff
@@ -103,117 +105,146 @@ func NewInfo() *Info {
 	}
 }
 
-func (i *Info) IsContainer() bool {
-	res, err, _ := singleflightGroup.Do(IsContainerKey, func() (interface{}, error) {
```
aphralG, line 107: Since single flight has been removed from the functions does something need to be done where they are being called to prevent duplicate function calls?

sean-breen: I think it's only called once per new resource, so I don't think there's a danger of duplicate calls, AFAIK

dhurley (approved):

craigell (approved):

Issue #1209, opened 2025-08-12
nginx-agent.service using deprecated options


UnwashedMeme:
### Bug Overview

When I install nginx-agent v3 on Ubuntu 24.04 I expect it to work smoothly. While it does work it gives off a number of concerning errors

### Expected Behavior

I expect agent to run without filling my system journal with warnings. 

### Steps to Reproduce the Bug

1. Install agent on Ubuntu 24.04. 
2. Run `systemctl status nginx-agent`

```text
nbird@dpvm-tf2k:~$ journalctl -u nginx-agent.service | head
Aug 11 20:54:43 dpvm-tf2k systemd[1]: /etc/systemd/system/nginx-agent.service:23: PIDFile= references a path below legacy directory /var/run/, updating /var/run/nginx-agent/nginx-agent.pid → /run/nginx-agent/nginx-agent.pid; please update the unit file accordingly.
Aug 11 20:54:43 dpvm-tf2k systemd[1]: /etc/systemd/system/nginx-agent.service:26: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether.
Aug 11 20:54:43 dpvm-tf2k systemd[1]: /etc/systemd/system/nginx-agent.service:27: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether.
```

### Environment Details

```console
nbird@dpvm-tf2k:~$ nginx-agent --version
nginx-agent version v3.1.0-2b6a9f2
nbird@dpvm-tf2k:~$ lsb_release  -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.3 LTS
Release:        24.04
Codename:       noble
```

### Additional Context

A broader perspective is that I suspect the packaging template needs to be refreshed.

Looking at https://github.com/nginx/agent/blob/31f9ef0717a810253bdc61dd9acb6bc0cc43dd35/scripts/packages/nginx-agent.service there's some oddness in here:

- Why is a `Type=simple` service have a pidfile directive at all?
- `StandardOutput`/`StandardError` should probably be journal (or omitted to use system default, which is normally journal)
- `SyslogIdentifier` is not necessary? This should be the default value already since the service is named nginx-agent
[...]

PR #1218, opened 2025-08-19
Update systemd unit file

CVanF5:
### Proposed changes

Added `chown` command in case Agent runs as non-root. The ExecStartPre commands run as root and will create the directories as root. If Agent needs to run as a different user it wouldn't be able to write logs.

Removed StandardOutput and StandardError lines. They will now just fall back to default `journal`

Removed `Environment=` as this will clear all environment variables. There are cases like `HTTP_PROXY` where we may want environment variables not cleared.

Added some optional security directives:
`NoNewPrivilages=true` prevents privilege escalation via setuid 
`ProtectHome=off` Disables home protection (hides /home from Agent) Users may opt to run NGINX in /home/user
 
Fixes: https://github.com/nginx/agent/issues/1209

### Checklist


scripts/packages/nginx-agent.service
```diff
@@ -11,24 +11,20 @@ User=${AGENT_USER}
 Group=${AGENT_GROUP}
 Type=simple
 
-ExecStartPre=/bin/mkdir -p ${AGENT_RUN_DIR}
-ExecStartPre=/bin/mkdir -p ${AGENT_LOG_DIR}
-
+ExecStartPre=/bin/mkdir -p ${AGENT_RUN_DIR} ${AGENT_LOG_DIR}
```
UnwashedMeme: ### run directory

Traditionally this is used to hold a pid file for daemonizing processes and sometimes socket files for listeners. nginx-agent isn't self-daemonizing (double-fork).  

Listen sockets: is there a use case of setting up a syslog listener in `/run/nginx-agent/syslog.socket` or something like that?

Looking for uses of of `run/nginx-agent` in the codebase I can find: https://github.com/nginx/agent/blob/main/internal/watcher/instance/instance_watcher_service.go#L31 that is used as the default for `processPath` (https://github.com/nginx/agent/blob/main/internal/watcher/instance/instance_watcher_service.go#L302) -- but that's supposed to be the path for the binary so this default isn't very sound.  There's a related test file entry.

This feels like an easy cross-off to me.

### Logs Directory

My preference at this point is always to turn file logging off and give that to the journal that has really good built in rotation to not run the disk out of space.  Not everyone loves the journal though; I don't know if you have a policy on this.

I see also that it is used as the default for opentelemetry logging (https://github.com/nginx/agent/blob/main/internal/config/defaults.go#L78). Mostly the same question -- does opentelemetry need to log to a file by default or can that go to stderr (journal) and be user-configurable to send elsewhere at need?

I could see keeping this, but maybe move this directory creation to `postinstall.sh` so that if an admin is trying to limit what directories nginx-agent can access it doesn't trigger with failures on `/var/log`

scripts/packages/nginx-agent.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=NGINX Agent
 Documentation=https://github.com/nginx/agent#readme
-After=network.target
+After=network-online.target
```
UnwashedMeme: If nginx-agent correctly retries connections -- which it should really be doing anyways because networks are never "fully healthy" -- then even if it starts before the network is fully configured  that shouldn't be a problem.  

If we can keep this (and the following `Wants=`) at `network.target` that will avoid bringing in a slowdown on boot; see [network-online.target](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#network-online.target)

Maybe split out a separate ticket to test nginx-agent reacts well (either directly or through a service restart) if the connections can't be made on startup, fix where necessary, and remove this directive then?

scripts/packages/nginx-agent.service
```diff
@@ -11,24 +11,20 @@ User=${AGENT_USER}
...
-
+Restart=on-failure
 PermissionsStartOnly=true
-PIDFile=${AGENT_RUN_DIR}/nginx-agent.pid
-Environment=
 
-StandardOutput=syslog
-StandardError=syslog
-SyslogIdentifier=nginx-agent
+KillSignal=SIGINT
```
UnwashedMeme: This [defaults to](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillSignal=) `SIGTERM`; which I believe nginx-agent [reacts to the same as SIGINT](https://github.com/nginx/agent/blob/main/cmd/agent/main.go#L31). Am I missing something, is this necessary?

CVanF5: I had it in my head that Agent reacted differently to SIGINT. I'll remove this

scripts/packages/nginx-agent.service
UnwashedMeme: I see the new version is both targetting `network.target`, going to assume you went with this.

scripts/packages/nginx-agent.service
CVanF5: I'll review this with the Agent team and follow up. I can think of 1 or 2 cases where Agent might want to open a listening socket. One of them as you said is a syslog receiver

UnwashedMeme: If there's a use-case then by all means keep it-- I'm just generally trying to shed privileges.

I don't want to get too hung up here; if you want to just move forward don't let me stop you. Here's a bunch more though:

### systemd option

You can also get systemd to build these directories with [RuntimeDirectory=](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=) (and LogsDirectory=).  This has a small benefit that I don't need to grant this service permission to `/var/run` | `/run` directly that this `ExecStartPre=` needs right now

E.g. on my _ubuntu 24.04_ system right now:

```ini
# Clear out nginx-agent trying to mkdir /var/log/nginx-agent /var/run/nginx-agent
ExecStartPre=

# otel tries to log here
LogsDirectory=nginx-agent
# not particularly used
RuntimeDirectory=nginx-agent
# store manifest information
StateDirectory=nginx-agent

#- /var/ construction
TemporaryFileSystem=/var:ro
BindPaths=/var/www
BindPaths=/var/log/nginx

# similar tmpfs construct for /etc:ro omitted
```

I'm blocking off nginx-agent from seeing anything else in `/var` or `/etc` that I'm not explicitly granting it.


E.g.
```console
root@dpvm-h2tm:~# systemctl status nginx-agent | grep PID
   Main PID: 59236 (nginx-agent)

root@dpvm-h2tm:~# cat /proc/59236/mounts | grep var
tmpfs /var tmpfs ro,nosuid,nodev,size=98148k,nr_inodes=409600,mode=755,inode64 0 0
/dev/root /var/lib/nginx-agent ext4 rw,nosuid,relatime,discard,errors=remount-ro,commit=30 0 0
/dev/root /var/log/nginx ext4 rw,nosuid,relatime,discard,errors=remount-ro,commit=30 0 0
/dev/root /var/log/nginx-agent ext4 rw,nosuid,relatime,discard,errors=remount-ro,commit=30 0 0
/dev/root /var/tmp ext4 rw,nosuid,relatime,discard,errors=remount-ro,commit=30 0 0
/dev/root /var/www ext4 rw,nosuid,relatime,discard,errors=remount-ro,commit=30 0 0

# Create a file just to demo this is a shared bind
root@dpvm-h2tm:~# touch /var/log/nginx-agent/foo

# run a shell in that mount namespace
root@dpvm-h2tm:~# nsenter --target 59236 --mount /bin/bash

root@dpvm-h2tm:/# ls -hl /var/
total 8.0K
drwxr-xr-x 3 root root   60 Aug 21 14:13 lib
drwxr-xr-x 4 root root   80 Aug 21 14:13 log
drwxrwxrwt 2 root root 4.0K Aug 21 14:13 tmp
drwxr-xr-x 2 root root 4.0K Aug 20 17:22 www

root@dpvm-h2tm:/# ls -hl /var/log/nginx-agent
total 0
-rw-r--r-- 1 root root        0 Aug 21 14:14 foo
-rw------- 1 root nginx-agent 0 Aug 21 14:13 otel.log

root@dpvm-h2tm:/# rm /var/log/nginx-agent/foo 
```

If I remove the `ExecStartPre=` that blanks out the upstream version, it fails because it can't create a directory on a read-only filesystem (the /var/ tmpfs):
```console
root@dpvm-h2tm:~# sed -i '/ExecStartPre=/d' /etc/systemd/system/nginx-agent.service.d/fixups.conf
root@dpvm-h2tm:~# systemctl daemon-reload
root@dpvm-h2tm:~# systemctl restart nginx-agent
Job for nginx-agent.service failed because the control process exited with error code.
See "systemctl status nginx-agent.service" and "journalctl -xeu nginx-agent.service" for details.
root@dpvm-h2tm:~# systemctl status nginx-agent
● nginx-agent.service - require nginx-agent-vars.service
     Loaded: loaded (/etc/systemd/system/nginx-agent.service; enabled; preset: enabled)
    Drop-In: /etc/systemd/system/nginx-agent.service.d
             └─fixups.conf, sandbox.conf, vars.conf
     Active: activating (auto-restart) (Result: exit-code) since Thu 2025-08-21 14:28:28 UTC; 1s ago
       Docs: https://github.com/nginx/agent#readme
    Process: 59718 ExecStartPre=/bin/mkdir -p /var/run/nginx-agent (code=exited, status=1/FAILURE)
        CPU: 18ms
```

UnwashedMeme (approved):

aphralG (approved):

PR #1241, opened 2025-08-27
Update NGINX config context after NGINX reload

dhurley:
### Proposed changes

Update NGINX config context after NGINX reload. This means that API action requests can be handled straight away after a config apply since agent will know the address of the NGINX Plus API.

### Checklist


internal/datasource/config/nginx_config_parser_test.go
```diff
@@ -805,353 +805,272 @@ func TestNginxConfigParser_checkLog(t *testing.T) {
...
+		oss  *model.APIDetails
+		plus *model.APIDetails
 		name string
 		conf string
-		oss  []*model.APIDetails
-		plus []*model.APIDetails
 	}{
 		{
-			plus: []*model.APIDetails{
-				{
```
aphralG, line 819: Why were the tests that checked `127.0.0.1` removed ?

dhurley: I simplified the logic to always point to localhost. Agent is always installed next to NGINX so it would always use localhost. We dont need to have logic to determine the hostname to use.

PR #1242, opened 2025-08-28
CPU Profiling

sean-breen:
### Proposed changes

Adding CPU Profiling to our Load test

### Checklist


Makefile.containers
```diff
@@ -52,6 +52,6 @@ CONTAINER_REGISTRY = registry.access.redhat.com
 BASE_IMAGE = $(CONTAINER_REGISTRY)/ubi$(OS_VERSION)/ubi:latest
 endif
 
-CONTAINER_NGINX_IMAGE_REGISTRY ?=  registry.nginx.com
+CONTAINER_NGINX_IMAGE_REGISTRY ?=  docker-registry.nginx.com
```
aphralG, line 55: Why is this change needed ?

Makefile
```diff
@@ -288,26 +288,66 @@ local-rpm-package: ## Create local rpm package
...
+	@$(GOTOOL) pprof -proto -output=default.pgo build/test/profiles/merged.pprof build/test/load-cpu-profiling/load/metrics_load_cpu.pprof \
+		|| { echo "Failed to generate PGO profile"; exit 1; }
 
 # run under sudo locally
 load-test-image: ## Build performance load testing image
-	@echo "🚚 Running load tests"
-	$(CONTAINER_BUILDENV) $(CONTAINER_CLITOOL) build -t $(IMAGE_TAG)_load_test . \
-		--no-cache -f ./test/docker/load/Dockerfile \
+	@echo "🚚  Building load test image"
+# add 		--no-cache to the command below to avoid using cache
```
aphralG: Is this comment needed ?

scripts/performance/profiling.sh
```diff
@@ -0,0 +1,54 @@
...
+for pkg in $packages; do
+    echo "Running tests in package: ${pkg}"
+    go test \
+      -count 10 -timeout 3m \
+      -cpuprofile "${PROFILES_DIR}/$(basename $pkg)_watcher_cpu.pprof" \
+      "./${pkg}" || { echo "Tests failed in package: ${pkg}, but continuing..."; continue; }
+    echo "Profile saved to: ${PROFILES_DIR}/$(basename $pkg)_watcher_cpu.pprof"
+done
+
+### Run integration tests with CPU profiling for each package
```
aphralG: Was this supposed to be removed ?

scripts/performance/profiling.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -e
+set -o pipefail
+
+PROFILES_DIR="build/test/profiles"
+mkdir -p ${PROFILES_DIR}
+
+# Run watcher tests with CPU profiling for each package
+echo "Starting watcher tests with cpu profiling..."
+packages=$(find internal/watcher -type f -name '*_test.go' -exec dirname {} \; | sort -u)
```
aphralG, line 23: Do we need to add a script? Could this be done in the make file ?

Makefile.containers
sean-breen: `registry.nginx.com` isn't used anymore. I figured if we set a default value it should at least be a real registry

Makefile
sean-breen: I'll add this back to avoid using the cache, such is the behaviour for our other docker images

scripts/performance/profiling.sh
sean-breen: It probably could but it's much nicer to do it in bash vs Makefile. If we expand our profiling later I don't want to bloat these Makefiles

scripts/performance/profiling.sh
sean-breen: Removed the commented code. I planned originally to run the integration tests as part of the profiling but turns out it doesn't actually work as we thought. In order to run the profiling with the integration test we'll need to implement the profiling inside the agent binary via `pprof` library.

scripts/performance/profiling.sh
aphralG: Ok I'm just not the biggest fan of adding scripts

PR #1261, opened 2025-09-09
Restart Subscribe on Connection Reset 

aphralG:
### Proposed changes
- Cancel and restart subscribe on connection reset.
- Update `FilePlugin` to update the `FileManagerService` connection instead of creating a new one.
- Update `UpdateManifestFile` so if the `previousManifestFiles` map is empty to use the `currentFiles`
- Add `RollbackManifest` bool to determine if the manifest needs to be rollback after a failed config apply
- More logging 

### Checklist


internal/file/file_manager_service.go
```diff
@@ -161,6 +170,7 @@ func (fms *FileManagerService) ConfigApply(ctx context.Context,
 
 	fileErr := fms.executeFileActions(ctx)
 	if fileErr != nil {
+		fms.rollbackManifest = false
```
dhurley, line 174: When does this boolean get reset to true?

aphralG: Good spot, I forgot adding now

PR #1265, opened 2025-09-15
Perform atomic writes during a config apply

dhurley:
### Proposed changes

Before update any NGINX config files, all files will be downloaded to a tmp directory before the NGINX config files are replaced.

### Checklist


internal/file/file_manager_service.go
```diff
@@ -474,37 +485,127 @@ func (fms *FileManagerService) manifestFile() (map[string]*model.ManifestFile, m
...
+
+	// Move/Delete files
 	for _, fileAction := range fms.fileActions {
 		switch fileAction.Action {
 		case model.Delete:
-			slog.DebugContext(ctx, "File action, deleting file", "file", fileAction.File.GetFileMeta().GetName())
+			slog.DebugContext(ctx, "Deleting file", "file", fileAction.File.GetFileMeta().GetName())
 			if err := os.Remove(fileAction.File.GetFileMeta().GetName()); err != nil && !os.IsNotExist(err) {
-				return fmt.Errorf("error deleting file: %s error: %w",
+				actionError = fmt.Errorf("error deleting file: %s error: %w",
```
aphralG, line 540: Why the switch from returning the error ? Asking as if one fails we will have to rollback so not sure continuing with other actions is needed.

internal/file/file_manager_service.go
```diff
@@ -474,37 +485,127 @@ func (fms *FileManagerService) manifestFile() (map[string]*model.ManifestFile, m
...
+	// Remove temp files if there is a failure moving or deleting files
+	if actionError != nil {
+		fms.deleteTempFiles(ctx, tempDir)
+	}
+
+	return actionError
 }
 
-func (fms *FileManagerService) fileUpdate(ctx context.Context, file *mpi.File) error {
+func (fms *FileManagerService) moveFilesFromTempDirectory(
```
aphralG: Should the functions for the temp location and moving be in an operator? Either a new one or the file operator since I guess it is for writing files?

internal/file/file_manager_service.go
dhurley: oh my bad I'm meant to be breaking out of the for loop here not the switch. Let me update that

internal/file/file_manager_service.go
dhurley: yeah I can move these to the operator

aphralG (approved):

sean-breen (approved):

PR #1276, opened 2025-09-16
Fix isAllowedDir matching by prefix

sean-breen:
### Proposed changes

Stops isAllowedDir function from matching by prefix. Originally this function would allow any path which was prefixed by an allowed directory, i.e 
- if `/var/log/nginx` was present in allowed dirs, then the path `/var/log/nginx-test` would also be allowed. Not good!

This PR stops this match from taking place, but still matches if the provided path is a subdirectory of an already allowed directory, i.e
- if `/var/log/nginx` is in allowed dirs, `/var/log/nginx/mysubdir` will be allowed but `/var/log/nginx-test` will be blocked.

### Checklist


internal/config/types.go
```diff
@@ -480,29 +482,38 @@ func (c *Config) IsCommandServerProxyConfigured() bool {
...
-		directoryPath = filepath.Dir(directoryPath)
+	// Does the path exist?
+	info, err := os.Stat(path)
+	if err == nil && !info.IsDir() {
+		// Path exists and is not a directory
+		// It's a file so get the parent directory
+		path = filepath.Dir(path)
+	}
+	if err != nil {
+		if !os.IsNotExist(err) {
```
UnwashedMeme: As said above, I think this whole Stat and err handling isn't useful, but if I'm missing the reason then can we please handle the errors as one block rather than as three separate ifs, one of which is needlessly nested?


```golang
switch {
  case os.IsNotExists(err):  // nothing to do so why did we do this?
  case err != nil:
     return false, fmt.Errorf("PathError %s %w", path, err) // disk i/o error surfaced in a weird location
  case !info.IsDir():
     // we don't trust our code to handle files that _do_ exist, so get the parent directory
     path = filepath.Dir(path)
}
```

internal/config/types.go
```diff
@@ -480,29 +482,38 @@ func (c *Config) IsCommandServerProxyConfigured() bool {
...
+		} // else the path does not exist, we will check if it's in an allowed directory
 	}
 
-	for _, allowedDirectory := range allowedDirs {
-		// Check if the directoryPath starts with the allowedDirectory
-		// This allows for subdirectories within the allowed directories.
-		if strings.HasPrefix(directoryPath, allowedDirectory) {
+	for _, dir := range allowedDirs {
+		// Check if the path is a direct match, or is a subdirectory of the allowed directory
+		if slices.Contains(allowedDirs, path) || strings.HasPrefix(path, dir+"/") {
```
UnwashedMeme: I don't think this is a good check
- for dir in alloweddirs, do a slices.Contains on the alloweddirs list; this is needlessly O(n^2). 
- if the AllowedDirs list is `["/"]` this is going to check if the path has the prefix `//`


Look at [this implementation](https://gitlab.com/f5/nginx/nginxazurelb/azure-resource-provider/-/blob/890feee6ec62463945b5c8a21edaf2f407c370ca/pkg/validation/vfs/vfs.go#L277-290)

We walk the given path up the tree ensuring that it is an exact match against the allowed prefix. we use `strings.HasPrefix` as a shortcircuit out -- e.g. `/var/www2/foo/bar/bam/jack/`  checked against `/var/www`  exits out once it is comparing `/var/www2` against `/var/www`.  Thinking about it right now, that hasprefix shortcircuit _might_ take more computation then just continuing to walk up the directory tree-- would need a benchmark to really evaluate.

See https://go.dev/play/p/tYMSZ6FbL31 as an example of walking up the tree

internal/config/types.go
```diff
@@ -480,29 +482,38 @@ func (c *Config) IsCommandServerProxyConfigured() bool {
...
 
-	if isFilePath {
-		directoryPath = filepath.Dir(directoryPath)
+	// Does the path exist?
+	info, err := os.Stat(path)
+	if err == nil && !info.IsDir() {
+		// Path exists and is not a directory
+		// It's a file so get the parent directory
+		path = filepath.Dir(path)
+	}
```
UnwashedMeme: Why do we think this stat and parent is useful?

If this function is called with `isAllowedDir("/etc/nginx/non-existant-file.conf", ["/etc/nginx"])` then it will skip this branch and needs to work.

If this function is called with `isAllowedDir("/etc/nginx/exists.conf", ["/etc/nginx"])` then it will be converted to "/etc/nginx" and it needs to work.  

What are we gaining on this `Stat` and `IsDir` check? The only case I can think of is to try and catch a case where the file does exist and this process doesn't have read permission to it.  But that's not the purpose of this function, and surfacing the error here feels really odd...

internal/config/types_test.go
```diff
@@ -66,6 +66,14 @@ func TestTypes_isAllowedDir(t *testing.T) {
...
 			filePath: "/not-nginx-test/idontexist.conf",
 		},
+		{
+			name:    "Test 7: Prefix match not allowed",
+			allowed: false,
+			allowedDirs: []string{
+				"/etc/nginx",
+			},
+			filePath: "/etc/nginx-test/nginx.conf",
+		},
```
UnwashedMeme, line 76:
```suggestion
		},
	    {
			name:    "Test 8: allow root should allow all",
			allowed: true,
			allowedDirs: []string{"/"},
			filePath: "/etc/nginx-test/nginx.conf",
		},
```

internal/config/config_test.go
```diff
@@ -1037,8 +1037,8 @@ func agentConfig() *Config {
 			},
 		},
 		AllowedDirectories: []string{
-			"/etc/nginx/", "/etc/nginx-agent/", "/usr/local/etc/nginx/", "/var/run/nginx/", "/var/log/nginx/",
-			"/usr/share/nginx/modules/", "/etc/app_protect/",
+			"/etc/nginx/", "/etc/nginx-agent/", "/usr/local/etc/nginx", "/var/run/nginx", "/var/log/nginx",
```
UnwashedMeme: Why are we removing the trailing slash on _some_ of the paths in this test?

UnwashedMeme (changes requested):

internal/config/config_test.go
UnwashedMeme: When I played with this briefly it looked like these paths were normlaized and had the trailing slash stripped as config was being parsed and processed. 

This change is confusing though, why remove it from some and not all?

internal/config/types_test.go
sean-breen: Using this as the base case for recursive function now, so if we hit the root directory it is assumed to not be allowed. Not sure if there is any concern here, or if a customer may want to add the root dir and allow agent to write anywhere on the machine?

internal/config/types.go
```diff
@@ -480,32 +470,22 @@ func (c *Config) IsCommandServerProxyConfigured() bool {
...
+func isAllowedDir(path string, allowedDirs []string) bool {
+	return checkDirIsAllowed(filepath.Clean(path), allowedDirs)
+}
 
-	if isFilePath {
-		directoryPath = filepath.Dir(directoryPath)
+func checkDirIsAllowed(path string, allowedDirs []string) bool {
+	slog.Info(filepath.Clean(path))
+	if slices.Contains(allowedDirs, path) {
+		slog.Info("Path is allowed", "path", path)
```
UnwashedMeme: I feel like these logs are leftover from debugging? If you think the logging is good longterm maybe move the logs  up to `isAllowedDir` and have the helper `checkDirIsAllowed` return the index so you could log the allowed path? Just a thought.

UnwashedMeme: I might also like to see this helper as a closure inside `isAllowedDir` since that is doing the initial `filepath.Clean`. By putting it inside `isAllowedDir` you ensure no one accidently calls `checkDirIsAllowed` that has the same visible signature but doesn't handle that initial clean.

internal/config/types_test.go
UnwashedMeme: I see you're using it as the base case, but since the slices.contains is before the check for `=="/"` then shouldn't this case (and test) still work?

internal/config/types_test.go
```diff
@@ -66,12 +66,61 @@ func TestTypes_isAllowedDir(t *testing.T) {
...
+			name:    "Test 10: Multiple allowed directories, file not in any of them",
+			allowed: false,
+			allowedDirs: []string{
+				"/etc/nginx",
+				"/usr/local/nginx",
+			},
+			filePath: "/opt/nginx/nginx.conf",
+		},
+		{
+			name:    "Test 11: Allowed directory is root",
```
UnwashedMeme: This doesn't describe the test...

UnwashedMeme (approved):

internal/config/types_test.go
UnwashedMeme: >  allow agent to write anywhere on the machine?

E.g. its in a docker container and the customer just wants this working as fast as possible without futzing with the paths.

sean-breen: Gotcha, thanks Nathan

internal/config/types.go
sean-breen: I've removed the logging, it was mainly for debugging

internal/config/types.go
```diff
@@ -480,32 +469,19 @@ func (c *Config) IsCommandServerProxyConfigured() bool {
...
-	}
-
-	directoryPath := path
-	// Check if the path is a file, regex matches when end of string is /<filename>.<extension>
-	isFilePath, err := regexp.MatchString(`/(\w+)\.(\w+)$`, directoryPath)
-	if err != nil {
-		return false, errors.New("error matching path" + directoryPath)
-	}
+// It recursively checks the parent directories of the path, until it finds a match or reaches the root directory.
+func isAllowedDir(path string, allowedDirs []string) bool {
```
aphralG, line 473: Could `isAllowedDir` & `IsDirectoryAllowed` be merged into one function ?

sean-breen: `IsDirectoryAllowed` is a method on the Config object, and it seems `isAllowedDir` was created to be available to the whole package in cases where we don't pass the actual Config object around

aphralG: Ya I think that was the original intention and makes sense if we want to allow that but I couldn't see anywhere that we use isAllowedDir outside of` types.go` so I don't know if there is a reason to have it anymore. It's just a small thing so I don't mind either way

PR #1284, opened 2025-09-22
Prevent execute permissions from being set

spencerugbo:
### Proposed changes

This PR aims to prevent any files from having execute permissions set during config apply.

### Checklist

### DIFF
--- a/internal/file/file_manager_service.go
+++ b/internal/file/file_manager_service.go
@@ -13,6 +13,8 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"strconv"
+	"strings"
 	"sync"
 
 	"google.golang.org/grpc"
@@ -150,6 +152,11 @@ func (fms *FileManagerService) ConfigApply(ctx context.Context,
 	allowedErr := fms.checkAllowedDirectory(fileOverview.GetFiles())
 	if allowedErr != nil {
 		return model.Error, allowedErr
+	}
+
+	permissionErr := fms.validateAndFixPermissions(ctx, fileOverview.GetFiles())
+	if permissionErr != nil {
+		return model.PermissionChange, permissionErr
 	}
 
 	diffFiles, fileContent, compareErr := fms.DetermineFileActions(
@@ -513,6 +520,58 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
 		if !allowed {
 			return fmt.Errorf("file not in allowed directories %s", file.GetFileMeta().GetName())
 		}
+	}
+
+	return nil
+}
+
+func (fms *FileManagerService) validateAndFixPermissions(ctx context.Context, fileList []*mpi.File) error {
+	var permissionIssues []string
+
+	for _, file := range fileList {
+		if err := fms.checkFilePermissions(file); err != nil {
+			permissionIssues = append(permissionIssues, file.GetFileMeta().GetName())
+
+			if resetErr := fms.resetFilePermissions(file); resetErr != nil {
+				return fmt.Errorf("failed to reset permissions for %s: %w", file.GetFileMeta().GetName(), resetErr)
+			}
+
+			slog.InfoContext(ctx, "Reset execute permissions", "file", file.GetFileMeta().GetName())
+		}
+	}
+
+	if len(permissionIssues) > 0 {
+		return fmt.Errorf("reset execute permissions for files: %s", strings.Join(permissionIssues, ", "))
+	}
+
+	return nil
+}
+
+func (fms *FileManagerService) checkFilePermissions(file *mpi.File) error {
+	filePermission := file.GetFileMeta().GetPermissions()
+
+	permissionCodes := filePermission[1:]
+
+	for _, digit := range permissionCodes {
+		singleCode := digit - '0'
+
+		if singleCode&1 != 0 {
+			return fmt.Errorf("file %s has execute permissions", file.GetFileMeta().GetName())
+		}
+	}
+
+	return nil
+}
+
+func (fms *FileManagerService) resetFilePermissions(file *mpi.File) error {
+	perm, err := strconv.ParseUint("0644", 8, 32)
+	if err != nil {
+		return fmt.Errorf("error parsing file permissions: %w", err)
+	}
+
+	err = os.Chmod(file.GetFileMeta().GetName(), os.FileMode(perm))
+	if err != nil {
+		return fmt.Errorf("failed to set file permissions: %w", err)
 	}
 
 	return nil
--- a/internal/file/file_plugin.go
+++ b/internal/file/file_plugin.go
@@ -347,6 +347,26 @@ func (fp *FilePlugin) handleConfigApplyRequest(ctx context.Context, msg *bus.Mes
 		}
 
 		fp.messagePipe.Process(ctx, &bus.Message{Topic: bus.WriteConfigSuccessfulTopic, Data: data})
+	case model.PermissionChange:
+		slog.WarnContext(ctx, "Files with execute permissions found and reset",
+			"details", err.Error())
+		dpResponse := fp.createDataPlaneResponse(
+			correlationID,
+			mpi.CommandResponse_COMMAND_STATUS_OK,
+			"Config apply successful, files updated with permission changes",
+			instanceID,
+			"",
+		)
+
+		successMessage := &model.ConfigApplySuccess{
+			ConfigContext:     &model.NginxConfigContext{},
+			DataPlaneResponse: dpResponse,
+		}
+
+		fp.fileManagerService.ClearCache()
+		fp.messagePipe.Process(ctx, &bus.Message{Topic: bus.ConfigApplySuccessfulTopic, Data: successMessage})
+
+		return
 	}
 }
 
--- a/internal/model/config.go
+++ b/internal/model/config.go
@@ -75,6 +75,7 @@ const (
 	NoChange
 	Error
 	OK
+	PermissionChange = 5
 )
 
 type ConfigApplySuccess struct {


internal/file/file_manager_service.go
```diff
@@ -595,6 +602,58 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+
+	permissionCodes := filePermission[1:]
+
+	for _, digit := range permissionCodes {
+		singleCode := digit - '0'
+
+		if singleCode&1 != 0 {
+			return fmt.Errorf("file %s has execute permissions", file.GetFileMeta().GetName())
+		}
+	}
```
UnwashedMeme: This looks like we're expecting the permissions being represented as an octal string.

- use strconv? This will also tell us if we're dealing with some unexpected representation.
- do the mask all at once?

```suggestion
	pv, err := strconv.ParseUint(filePermission, 8, 32) 
	if err != nil {
	    return fmt.Errorf("file %s has malformed permissions", file.GetFileMeta().GetName())
	}
	if pv&0111 > 0 {
    	return fmt.Errorf("file %s has execute permissions", file.GetFileMeta().GetName())
	}
```

internal/file/file_manager_service.go
```diff
@@ -595,6 +602,58 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+
+	return nil
+}
+
+func (fms *FileManagerService) resetFilePermissions(file *mpi.File) error {
+	perm, err := strconv.ParseUint("0644", 8, 32)
```
UnwashedMeme: Just use a fixed literal if that's what you want:
```suggestion
	perm := 0o644
```

But I'm a bit suspicious of that. If the user requested `0o700`, it's going to end up as `0o644` and unintentionally made world readable?

While I like having little functions that do one thing, this might be a case for integrating it a bit more:
- parse permissions, if err, append error and use `p := 0o644`
- check permission with `p&0111 > 0`, if err append error and use `p = p&0o666`
- set permission to p.

This way you preserve the permissions the user gave except for the permission bit.

internal/file/file_manager_service.go
spencerugbo: Good idea, I've made changes to use strconv to parse the file permissions

internal/file/file_manager_service.go
spencerugbo: Thats a better way of dealing with the permissions, now the new permission without the execute bits in now based on the original permission

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+	filePermission := file.GetFileMeta().GetPermissions()
+
+	permissionOctal, err := strconv.ParseUint(filePermission, 8, 32)
+	if err != nil {
+		chmodErr := os.Chmod(file.GetFileMeta().GetName(), os.FileMode(filePerm))
+		if chmodErr != nil {
+			return fmt.Errorf("failed to set file permissions: %w", err)
+		}
+
+		return fmt.Errorf("falied to parse file permissions, permissions set to 0o644: %w", err)
```
UnwashedMeme: I've recently been on a bit of an "anti-constant" kick because of cases like this.   The prevailing wisdom, for good reason, is don't use magic numbers in your code.  The _purpose_ of that is that a number by itself isn't obvious and _by assigning it a name_ you document what it is.  

This is a case of following the rule but not helping. I'm reading the code and I go "what the hell permissions are being set?" The name--it's pretty much just a type name--doesn't really tell me or reassure me that we're using the right the permission value, there is no docstring on it, and it is 600 lines away from this single use.

And to really emphasize that disconnect the error message on L653 is now wrong but it is difficult to spot because the distance between definition and use. Of course the constant could formatted into the error string...

Consider:
```suggestion
        defaultPerm := 0o600
		chmodErr := os.Chmod(file.GetFileMeta().GetName(), os.FileMode(defaultPerm))
		if chmodErr != nil {
			return fmt.Errorf("failed to set file permissions: %w", err)
		}
		return fmt.Errorf("falied to parse file permissions, permissions set to %O: %w", defaultPerm, err)
```
- The name now tells us a bit more; it's the default
- The value is now local to the use, so it is easy to know.  I think that also makes it easier to spot that the fmt.Errorf _should_ use it.

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+	}
+
+	newPermission := permissionOctal &^ executePerm
```
UnwashedMeme: Same as above, do it here:


```suggestion
    executeMask := 0o111
	newPermission := permissionOctal &^ executeMask
```

Like the use of `&^` here.

UnwashedMeme: This one is at least in two places; but my preference is still to have it locally because then the question of "are we talking executeMask for user, group, other, or all" is immediately obvious rather than 600 lines away.

craigell (approved):

internal/model/config.go
```diff
@@ -75,6 +75,7 @@ const (
 	NoChange
 	Error
 	OK
+	PermissionChange = 5
```
aphralG: I don't think you need to specify the value, iota + 1 will assign the values 
```suggestion
	PermissionChange
```

internal/file/file_plugin.go
```diff
@@ -347,6 +347,26 @@ func (fp *FilePlugin) handleConfigApplyRequest(ctx context.Context, msg *bus.Mes
 		}
 
 		fp.messagePipe.Process(ctx, &bus.Message{Topic: bus.WriteConfigSuccessfulTopic, Data: data})
+	case model.PermissionChange:
```
aphralG: Do we need a separate case for this? Could the failure to fix permissions go under rollbackRequired ?

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+
+			if resetErr := fms.resetFilePermissions(file); resetErr != nil {
+				return fmt.Errorf("failed to reset permissions for %s: %w", file.GetFileMeta().GetName(), resetErr)
+			}
+
+			slog.InfoContext(ctx, "Reset execute permissions", "file", file.GetFileMeta().GetName())
+		}
+	}
+
+	if len(permissionIssues) > 0 {
```
aphralG: Will this always be greater than 0 even if the permissions have been reset successfully so will always error?

internal/file/file_manager_service.go
```diff
@@ -156,6 +159,11 @@ func (fms *FileManagerService) ConfigApply(ctx context.Context,
 		return model.Error, allowedErr
 	}
 
+	permissionErr := fms.validateAndFixPermissions(ctx, fileOverview.GetFiles())
+	if permissionErr != nil {
+		return model.PermissionChange, permissionErr
```
aphralG: Could this be the below instead ?

```suggestion
		return return model.Error, allowedErr
```

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+		if chmodErr != nil {
+			return fmt.Errorf("failed to set file permissions: %w", err)
+		}
+
+		return fmt.Errorf("falied to parse file permissions, permissions set to 0o644: %w", err)
+	}
+
+	newPermission := permissionOctal &^ executePerm
+
+	err = os.Chmod(file.GetFileMeta().GetName(), os.FileMode(newPermission))
```
aphralG: Could we instead of changing the file itself change file.GetFileMeta.Permissions  so its updated in the file meta and then written and changed in the config apply

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+
+	for _, file := range fileList {
+		if err := fms.checkFilePermissions(file); err != nil {
+			permissionIssues = append(permissionIssues, file.GetFileMeta().GetName())
+
+			if resetErr := fms.resetFilePermissions(file); resetErr != nil {
+				return fmt.Errorf("failed to reset permissions for %s: %w", file.GetFileMeta().GetName(), resetErr)
+			}
+
+			slog.InfoContext(ctx, "Reset execute permissions", "file", file.GetFileMeta().GetName())
```
dhurley: Can you remove this log message?

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+		return fmt.Errorf("file %s has execute permissions", file.GetFileMeta().GetName())
+	}
+
+	return nil
+}
+
+func (fms *FileManagerService) resetFilePermissions(file *mpi.File) error {
+	filePermission := file.GetFileMeta().GetPermissions()
+
+	permissionOctal, err := strconv.ParseUint(filePermission, 8, 32)
```
dhurley, line 692: Can you add a debug log message to say what the file permissions are being updated from?

internal/file/file_manager_service.go
```diff
@@ -595,6 +603,66 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+	if err != nil {
```
dhurley, line 693: If the file permissions are invalid coming from the management plane then we should return an error straight away and not try to set a default file permission for the file.

internal/file/file_manager_service_test.go
```diff
@@ -322,6 +323,158 @@ func TestFileManagerService_checkAllowedDirectory(t *testing.T) {
...
+	fileManagerService := NewFileManagerService(nil, types.AgentConfig(), &sync.RWMutex{})
+
+	tests := []struct {
+		name        string
+		permissions string
+		errorMsg    string
+		expectError bool
+	}{
+		{
+			name:        "File with read and write permissions for owner",
```
dhurley: Can you add more test cases like 0007 and 0070? Just to cover more scenarios

dhurley: Can you update the test names to follow this format?
```
name:        "Test 1: File with read and write permissions for owner",
```

internal/file/file_manager_service.go
spencerugbo: Defining this value in both locations triggers a lint error. To meet with our coding standards both locations should reference a single constant instead of duplicating the definition.

internal/file/file_manager_service.go
```diff
@@ -595,6 +602,54 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+		return fmt.Errorf("falied to parse file permissions: %w", err)
+	}
+
+	permissionOctal &^= executePerm
+
+	newPermission := "0" + strconv.FormatUint(permissionOctal, 8)
+	file.FileMeta.Permissions = newPermission
+
+	slog.DebugContext(ctx, "Permissions have been changed", "file", file.GetFileMeta().GetName(),
+		"old permissions", filePermission, "new permissions", newPermission)
```
```suggestion
		"old_permissions", filePermission, "new_permissions", newPermission)
```

internal/file/file_manager_service_test.go
```diff
@@ -322,6 +322,170 @@ func TestFileManagerService_checkAllowedDirectory(t *testing.T) {
...
+		expectError bool
+	}{
+		{
+			name:        "Test 1: File with read and write permissions for owner",
+			permissions: "0600",
+			expectError: false,
+		},
+		{
+			name:        "Test 2: File with read/write and execute permissions for owner",
+			permissions: "0700",
```
aphralG, line 378: Could you add a few more test cases with execute permissions

aphralG (approved):

internal/file/file_manager_service.go
```diff
@@ -595,6 +602,54 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
 	return nil
 }
 
+func (fms *FileManagerService) validateAndFixPermissions(ctx context.Context, fileList []*mpi.File) error {
```
```suggestion
func (fms *FileManagerService) validateAndUpdateFilePermissions(ctx context.Context, fileList []*mpi.File) error {
```

internal/file/file_manager_service.go
```diff
@@ -595,6 +602,54 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
...
+	for _, file := range fileList {
+		err := fms.checkFilePermissions(file)
+		if err != nil {
+			resetErr := fms.resetFilePermissions(ctx, file)
```
```suggestion
			resetErr := fms.removeExecuteFilePermissions(ctx, file)
```

internal/file/file_manager_service.go
```diff
@@ -595,6 +602,54 @@ func (fms *FileManagerService) checkAllowedDirectory(checkFiles []*mpi.File) err
 	return nil
 }
 
+func (fms *FileManagerService) validateAndFixPermissions(ctx context.Context, fileList []*mpi.File) error {
+	for _, file := range fileList {
+		err := fms.checkFilePermissions(file)
```
```suggestion
		err := fms.areExecuteFilePermissionsSet(file)
```

PR #1285, opened 2025-09-22
Use temp files for Rollback 

aphralG:
### Proposed changes
- Removed `rollbackFileContents map[string][]byte` which was used to rollback files during a failed config apply
- Instead write files that have beed updated or deleted to a temp Dir `rollback<some-number>` in `/var/lib/nginx-agent/ `
- Clear all temp files and temp dirs after a config apply is complete in the ClearCache function 
### Checklist

### DIFF
--- a/internal/file/file_operator.go
+++ b/internal/file/file_operator.go
@@ -193,12 +193,19 @@ func (fo *FileOperator) MoveFile(ctx context.Context, sourcePath, destPath strin
 	if openErr != nil {
 		return openErr
 	}
+	defer inputFile.Close()
+
+	fileInfo, statErr := inputFile.Stat()
+	slog.Info("File Stat", "file", inputFile.Name(), "mode", fileInfo.Mode())
+	if statErr != nil {
+		return statErr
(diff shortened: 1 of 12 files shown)


internal/file/file_operator.go
```diff
@@ -188,28 +189,33 @@ func (fo *FileOperator) WriteManifestFile(
...
 
-	outputFile, err := os.Create(destPath)
-	if err != nil {
-		return err
+	if dirErr := os.MkdirAll(filepath.Dir(destPath), dirPerm); dirErr != nil {
+		return fmt.Errorf("failed to create directories for %s: %w", destPath, dirErr)
+	}
+
+	outputFile, createErr := os.Create(destPath)
+	if createErr != nil {
```
UnwashedMeme: - we're destroying any file permissions it had when it gets moved. Read from sourcepath and apply to destpath?
- It's often better to create a temp file in the target location and then do a rename once the data is written.  This way other processes don't see partially written files at expected locations.

PR #1302, opened 2025-09-26
Rollback Using Temp Files & Fix Atomic Writes

aphralG:
### Proposed changes

Changes:

- Update Atomic writes to create the temp directory in the location of the NGINX config or a default of `"/etc/nginx"` if we can't get the config dir. Instead of the file being moved it is now renamed preserving the file permissions etc 

- Rollback has changed to before a config apply starts create a copy of the file to a temp directory (need to use copy here to not delete original) at the location of the NGINX config. If a rollback occurs these files are renamed to rollback the files 

- File watchers are now enabled by a specific topic sent after we have cleaned up the temp directories, File Watcher now when disabled removes all directories it is watching and then they are added back when it is enabled. This is to prevent an issue where the events that happened while the watcher was disabled were still on the channel when the watcher was enabled causing agent to process those events unnecessarily. 

### Checklist

### DIFF
--- a/internal/file/file_manager_service.go
+++ b/internal/file/file_manager_service.go
@@ -71,7 +71,7 @@ type (
 			fileToUpdate *mpi.File,
 		) error
 		SetIsConnected(isConnected bool)
-		renameFile(ctx context.Context, hash, fileName, tempDir string) error
+		RenameFile(ctx context.Context, hash, fileName, tempDir string) error
 		UpdateClient(ctx context.Context, fileServiceClient mpi.FileServiceClient)
 	}
 
@@ -610,7 +610,7 @@ actionsLoop:
 			continue
 		case model.Add, model.Update:
 			fileMeta := fileAction.File.GetFileMeta()
-			err := fms.fileServiceOperator.renameFile(ctx, fileMeta.GetHash(), fileMeta.GetName(), tempDir)
+			err := fms.fileServiceOperator.RenameFile(ctx, fileMeta.GetHash(), fileMeta.GetName(), tempDir)
 			if err != nil {
 				actionError = err
 
--- a/internal/file/file_service_operator.go
+++ b/internal/file/file_service_operator.go
@@ -277,7 +277,7 @@ func (fso *FileServiceOperator) UpdateFile(
 }
 
 // renameFile, renames (moves) file from tempDir to new location to update file.
-func (fso *FileServiceOperator) renameFile(
+func (fso *FileServiceOperator) RenameFile(
 	ctx context.Context, hash, fileName, dir string,
 ) error {
 	slog.DebugContext(ctx, "Renaming file", "file", fileName)


internal/file/file_manager_service.go
```diff
@@ -193,13 +211,51 @@ func (fms *FileManagerService) ConfigApply(ctx context.Context,
 	return model.OK, nil
 }
 
+func (fms *FileManagerService) RollbackTempFiles(ctx context.Context) error {
```
```suggestion
func (fms *FileManagerService) backupFiles(ctx context.Context) error {
```
I think this might be a more descriptive name for this function. Also I dont think it needs to be a public function

internal/file/file_manager_service.go
```diff
@@ -193,13 +211,51 @@ func (fms *FileManagerService) ConfigApply(ctx context.Context,
...
+
 func (fms *FileManagerService) ClearCache() {
-	clear(fms.rollbackFileContents)
+	slog.Debug("Clearing cache and temp files after config apply")
 	clear(fms.fileActions)
 	clear(fms.previousManifestFiles)
+
+	configErr := os.RemoveAll(fms.tempConfigDir)
+	if configErr != nil {
+		slog.Error("error removing temp config directory", "path", fms.tempConfigDir, "err", configErr)
```
```suggestion
		slog.Error("Error removing temp config directory", "path", fms.tempConfigDir, "err", configErr)
```

internal/file/file_manager_service.go
```diff
@@ -193,13 +211,51 @@ func (fms *FileManagerService) ConfigApply(ctx context.Context,
...
+	}
+
+	rollbackErr := os.RemoveAll(fms.tempRollbackDir)
+	if rollbackErr != nil {
+		slog.Error("error removing temp rollback directory", "path", fms.tempRollbackDir, "err", rollbackErr)
```
```suggestion
		slog.Error("Error removing temp rollback directory", "path", fms.tempRollbackDir, "err", rollbackErr)
```

internal/file/file_manager_service.go
```diff
@@ -218,10 +274,24 @@ func (fms *FileManagerService) Rollback(ctx context.Context, instanceID string)
 			continue
 		case model.Delete, model.Update:
 			fileMeta := fileAction.File.GetFileMeta()
-			content := fms.rollbackFileContents[fileMeta.GetName()]
-			err := fms.fileOperator.Write(ctx, content, fileMeta.GetName(), fileMeta.GetPermissions())
-			if err != nil {
-				return err
+			fileName := fileMeta.GetName()
```
dhurley: Could this logic be moved into its own function?
Something like `restoreFiles`

internal/file/file_plugin.go
```diff
@@ -133,11 +133,28 @@ func (fp *FilePlugin) Subscriptions() []string {
 		bus.ConfigUploadRequestTopic,
 		bus.ConfigApplyRequestTopic,
 		bus.ConfigApplyFailedTopic,
-		bus.ConfigApplySuccessfulTopic,
+		bus.ReloadSuccessfulTopic,
 		bus.ConfigApplyCompleteTopic,
 	}
 }
 
+func (fp *FilePlugin) CleanUpConfigApply(ctx context.Context,
```
dhurley:
```suggestion
func (fp *FilePlugin) enableFileWatcher(ctx context.Context,
```
Would it be better to move out the `fp.fileManagerService.ClearCache()` and just have this function enable the watcher?

So that it looks something like this:
```
fp.fileManagerService.ClearCache()
fp.enableFileWatcher(ctx, successMessage.ConfigContext, successMessage.DataPlaneResponse.GetInstanceId())
```

internal/file/file_service_operator.go
```diff
@@ -277,7 +277,7 @@ func (fso *FileServiceOperator) UpdateFile(
 	return fso.sendUpdateFileStream(ctx, fileToUpdate, fso.agentConfig.Client.Grpc.FileChunkSize)
 }
 
-func (fso *FileServiceOperator) MoveFilesFromTempDirectory(
+func (fso *FileServiceOperator) MoveFileFromTempDirectory(
 	ctx context.Context, fileAction *model.FileCache, tempDir string,
 ) error {
```
dhurley:
```suggestion
func (fso *FileServiceOperator) RenameFile(
ctx context.Context, sourceFile, destination string,
) error {
```

I think calling the function rename instead of move makes it more clear in what its doing

internal/watcher/file/file_watcher_service.go
```diff
@@ -97,8 +97,25 @@ func (fws *FileWatcherService) Watch(ctx context.Context, ch chan<- FileUpdateMe
 	}
 }
 
-func (fws *FileWatcherService) SetEnabled(enabled bool) {
-	fws.enabled.Store(enabled)
+func (fws *FileWatcherService) DisableWatcher(ctx context.Context) {
+	if fws.watcher != nil && fws.watcher.WatchList() != nil {
```
dhurley, line 102: Can you add a debug message here to say disabling watcher? Same with the EnableWatcher function.

internal/watcher/file/file_watcher_service_test.go
```diff
@@ -265,7 +266,8 @@ func TestFileWatcherService_Watch(t *testing.T) {
 		defer os.Remove(skippableFile.Name())
 
 		select {
-		case <-channel:
+		case file := <-channel:
+			slog.Info("Skippable file updated", "", file)
```
dhurley: Can you remove this log?

internal/file/file_plugin.go
aphralG: It's enabling multiple watchers so switched it to enableWatchers

internal/file/file_service_operator.go
aphralG: Renamed the function but I need the hash to validate the file.

sean-breen:
> > `if we can't get the config dir`
> > What does this mean? We will write into the `/etc/nginx` directory by default? Shouldn't we stop if we can't find the config dir? What happens when we can't find the directory and someone has set it to a non-default location?
> 
> I can change it to fail just after discussing with Donal we were just trying to set it to a default in case something happened where it wouldn't be set. But in reality it will probably always bet set as it is sent with the config context.

Fair, I'm more concerned with us writing the intermediary into the `/etc/nginx` directory, would it be possible to write the temp files to `/etc/nginx-agent` or another location that we know will exist and we have access to?

internal/file/file_manager_service.go
```diff
@@ -71,7 +71,7 @@ type (
 			fileToUpdate *mpi.File,
 		) error
 		SetIsConnected(isConnected bool)
-		MoveFilesFromTempDirectory(ctx context.Context, fileAction *model.FileCache, tempDir string) error
+		renameFile(ctx context.Context, hash, fileName, tempDir string) error
```
dhurley: Think you can remove this from the interface?

craigell (approved):

PR #1312, opened 2025-09-30
Update Dependencies 

aphralG:
### Proposed changes
Update OTel Dependencies
Update Plus Go Client
Update other Dependencies 

### Checklist


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/nginx/agent/v3
 
-go 1.24.0
+go 1.24.2
```
dhurley, line 3: are we forced to bump the patch version here?

aphralG: Yes, this is caused by the plus go client. It's not specific to version 3 this will also happen if we update to 2.4.0

Akshay2191 (approved):

PR #1315, opened 2025-10-02
Prioritise API with api write=on

Akshay2191:
Issue:
Currently, if more than one Plus API is defined in NGINX config, Agent will always select the first working API URL, and if that API has  "api write=off", that leads to "MethodDisabled" error while trying to update upstream using an API.

Fix:
With this change agent is parsing all the Plus API's defined in the NGINX config to find if any of them have "api write=on" and if yes, then we are selecting that as the default api; else we are selecting the first API defined in the NGINX config as the default one, which is the current behaviour.

### Checklist

### DIFF
--- a/api/grpc/mpi/v1/command.proto
+++ b/api/grpc/mpi/v1/command.proto
@@ -345,6 +345,8 @@ message NGINXPlusRuntimeInfo {
     repeated string dynamic_modules = 5;
     // the plus API details
     APIDetails plus_api = 6;
+    // to store all the endpoints details
+    repeated APIDetails plus_apis = 7;
 }
 
 message APIDetails {
@@ -354,6 +356,8 @@ message APIDetails {
     string listen =  2;
     // the API CA file path
     string Ca = 3;
+    // flag to know if this API location was configured with 'api write=on;'
+    bool write_enabled = 4;
 }
 
 // A set of runtime NGINX App Protect settings
--- a/docs/proto/protos.md
+++ b/docs/proto/protos.md
@@ -697,6 +697,7 @@ Perform an associated API action on an instance
 | location | [string](#string) |  | the API location directive |
 | listen | [string](#string) |  | the API listen directive |
 | Ca | [string](#string) |  | the API CA file path |
+| write_enabled | [bool](#bool) |  | flag to know if this API location was configured with &#39;api write=on;&#39; |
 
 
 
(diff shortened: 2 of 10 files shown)


internal/datasource/config/nginx_config_parser.go
```diff
@@ -888,3 +921,19 @@ func (ncp *NginxConfigParser) isDuplicateFile(nginxConfigContextFiles []*mpi.Fil
 
 	return false
 }
+
+func (ncp *NginxConfigParser) sortPlusAPIs(apis []*model.APIDetails) []*model.APIDetails {
```
aphralG: Could you add a comment or update the naming to explain that this is trying to prioritise and find apis with write enabled?

internal/datasource/config/nginx_config_parser_test.go
```diff
@@ -150,7 +150,7 @@ var (
     listen [::]:8000;
 	server_name _;
     location /api/ {
-        api write=on;
+        api write=off;
```
aphralG, line 153: Could you add unit test for 
-  the case where there are multiple apis and the first one doesn't have write enabled but another does
- the function that prioritises the api with write on

api/grpc/mpi/v1/command.proto
```diff
@@ -354,6 +354,8 @@ message APIDetails {
     string listen =  2;
     // the API CA file path
     string Ca = 3;
+    // flag to know API is configured with 'write=on;'
+    bool write_enabled = 4;
```
aphralG: Does a change need to happen to the MPI ? Do the management planes care about the write status ? Could the change instead only be made to the `APIDetails struct` thats internal to the Agent?

docs/proto/protos.md
```diff
@@ -697,6 +697,7 @@ Perform an associated API action on an instance
 | location | [string](#string) |  | the API location directive |
 | listen | [string](#string) |  | the API listen directive |
 | Ca | [string](#string) |  | the API CA file path |
+| write_enabled | [bool](#bool) |  | flag to know API is configured with &#39;write=on;&#39; |
```
sean-breen: Similar to Aphral's question above, is it required that we modify the MPI and send this info via our protos? Would it be possible to keep it internal to avoid having to modify the API?

internal/datasource/config/nginx_config_parser.go
```diff
@@ -888,3 +912,47 @@ func (ncp *NginxConfigParser) isDuplicateFile(nginxConfigContextFiles []*mpi.Fil
...
+	foundWriteEnabled := false
+	for _, api := range apis {
+		if api.WriteEnabled {
+			foundWriteEnabled = true
+			break
+		}
+	}
+
+	if !foundWriteEnabled && len(apis) > 0 {
+		slog.WarnContext(ctx, "No write-enabled NGINX Plus API found. Defaulting to read-only API")
```
dhurley: Maybe Info level is better for this log

spencerugbo (approved):

PR #1317, opened 2025-10-03
Add default syslog server

aphralG:
### Proposed changes
Added syslog server port option in agent config
```
syslog_server:
    port: 1514
```

Default value is `1514`, agent will parse the NGINX config for Nap syslog server directives, when one is found agent will check if the host is valid and then check if the port matches the configured syslog server port in the config (or default if not set) 
If the agent fails to find a valid syslog server it will log the following warning, which contains the configured port 
```
level=WARN msg="Could not find available local NGINX App Protect syslog server configured on port 565. Security violations will not be collected."
```

Changed NapSyslogServers in NginxConfigContext to a single value as we no longer allow multiple syslog servers. 

### Checklist


internal/collector/otel_collector_plugin.go
```diff
@@ -705,40 +705,30 @@ func (oc *Collector) updateResourceAttributes(
 	return actionUpdated
 }
 
-func (oc *Collector) findAvailableSyslogServers(ctx context.Context, napSyslogServers []string) string {
-	napSyslogServersMap := make(map[string]bool)
-	for _, server := range napSyslogServers {
-		napSyslogServersMap[server] = true
+func (oc *Collector) findAvailableSyslogServers(ctx context.Context, napSyslogServer string) string {
```
dhurley:
```suggestion
func (oc *Collector) findAvailableSyslogServer(ctx context.Context, napSyslogServer string) string {
```

internal/config/config.go
```diff
@@ -420,6 +421,12 @@ func registerFlags() {
 		"A comma-separated list of features enabled for the agent.",
 	)
 
+	fs.String(
+		SyslogServerPort,
+		DefSyslogServerPort,
+		"The port Agent will start the syslog server for Nginx App Protect Security violations on",
```
```suggestion
		"The port Agent will start the syslog server on for logs collection",
```

internal/collector/otel_collector_plugin.go
```diff
@@ -705,40 +705,30 @@ func (oc *Collector) updateResourceAttributes(
...
+	}
 
-		slog.DebugContext(ctx, "Found valid NAP syslog server", "address", napSyslogServer)
+	closeError := ln.Close()
+	if closeError != nil {
+		slog.DebugContext(ctx, "Failed to close syslog server", "address", napSyslogServer, "error", closeError)
+	}
 
-		oc.previousNAPSysLogServer = napSyslogServer
+	slog.InfoContext(ctx, "Found valid NAP syslog server", "address", napSyslogServer)
```
dhurley: can we remove this info message and have it in the config parser instead?

internal/datasource/config/nginx_config_parser.go
```diff
@@ -251,15 +250,10 @@ func (ncp *NginxConfigParser) createNginxConfigContext(
...
-		if len(napSyslogServersFound) > 0 {
-			var napSyslogServer []string
-			for server := range napSyslogServersFound {
-				napSyslogServer = append(napSyslogServer, server)
-			}
-			nginxConfigContext.NAPSysLogServers = napSyslogServer
-		} else if napEnabled {
-			slog.WarnContext(ctx, "Could not find available local NGINX App Protect syslog server. "+
-				"Security violations will not be collected.")
+		if napEnabled && nginxConfigContext.NAPSysLogServer == "" {
```
dhurley: can you move this if statement outside the for loop so that its not logged multiple times?
also can you can an else statement to log on info message saying that you found a syslog server configured for NAP?

internal/collector/otel_collector_plugin.go
aphralG: There is a log message already in the parser that it found one. I had this on as the config parser could find a NAP server but when the otel plugin goes to use it the port is unavailable but I can remove it if its causing problems

PR #1342, opened 2025-10-16
Check for deprecated/unknown environment variables

Akshay2191:
Implemented a function to validate the environment variables passed are part of the new v3 allowed variables if not it should throw a warning message.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ permissions:
 
 env:
   NFPM_VERSION: 'v2.35.3'
-  GOPROXY: "https://${{ secrets.ARTIFACTORY_USER }}:${{ secrets.ARTIFACTORY_TOKEN }}@azr.artifactory.f5net.com/artifactory/api/go/f5-nginx-go-dev"
+  GOPROXY: "direct"
 
 jobs:
   proxy-sanity-check:


internal/config/config.go
```diff
@@ -63,6 +63,40 @@ func Execute(ctx context.Context) error {
 func Init(version, commit string) {
 	setVersion(version, commit)
 	registerFlags()
+	checkDeprecatedEnvVars()
+}
+
+func checkDeprecatedEnvVars() {
+	allViperKeys := make(map[string]struct{})
```
sean-breen, line 70: Is the map structure necessary here? From what I see we could use a []string and use `slices.Contains` to determine if it's in the list.

internal/config/config_test.go
```diff
@@ -163,6 +165,100 @@ func TestNormalizeFunc(t *testing.T) {
...
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			runDeprecatedEnvVarsTest(t, tc)
+		})
+	}
+}
+
+func runDeprecatedEnvVarsTest(t *testing.T, tc deprecatedEnvVarsTest) {
```
aphralG, line 225: Can this instead just be done inside the unit tests instead of splitting it up

internal/config/config_test.go
```diff
@@ -163,6 +165,100 @@ func TestNormalizeFunc(t *testing.T) {
 	assert.Equal(t, expected, result)
 }
 
+type deprecatedEnvVarsTest struct {
```
aphralG, line 168: Can this be moved into the test

internal/config/config_test.go
Akshay2191: That increases the Cyclomatic complexity of the Test function.

internal/config/config_test.go
Akshay2191: I had to define it outside as the linter gives error if we have more than  5 parameter in a function call.

internal/config/config.go
Akshay2191: I chose the map over slices as it will be the most performant and scalable solution. In slices, the time taken would be leaner O(n), whereas in a map, it will be constant O(1).

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ permissions:
 
 env:
   NFPM_VERSION: 'v2.35.3'
-  GOPROXY: "direct"
+  GOPROXY: "https://${{ secrets.ARTIFACTORY_USER }}:${{ secrets.ARTIFACTORY_TOKEN }}@azr.artifactory.f5net.com/artifactory/api/go/f5-nginx-go-dev"
```
sean-breen: You can revert this change, main branch will use the dev goproxy by default

PR #1345, opened 2025-10-16
Validating the label values against regex

Akshay2191:
Implemented a function to validate the label value against regex to align it with N1C.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ permissions:
 
 env:
   NFPM_VERSION: 'v2.35.3'
-  GOPROXY: "https://${{ secrets.ARTIFACTORY_USER }}:${{ secrets.ARTIFACTORY_TOKEN }}@azr.artifactory.f5net.com/artifactory/api/go/f5-nginx-go-dev"
+  GOPROXY: "direct"
 
 jobs:
   proxy-sanity-check:


internal/config/config.go
```diff
@@ -940,6 +942,17 @@ func resolveLabels() map[string]interface{} {
 	return result
 }
 
+func validateLabel(labelValue string) bool {
+	const maxLength = 256
+	labelPattern := regexp.MustCompile("^[a-zA-Z0-9]([a-zA-Z0-9-_]{0,254}[a-zA-Z0-9])?$")
```
sean-breen: Could you add this regex as a constant, similar to `regexInvalidFile` on line 49?

internal/config/testdata/nginx-agent.conf
```diff
@@ -13,7 +13,7 @@ watchers:
                     - \.*log$
 
 labels:
-  label1: label 1
+  label1: label-1
```
aphralG, line 16: Why the change in the label, will spaces no longer be allowed ?

internal/config/config.go
```diff
@@ -940,6 +943,17 @@ func resolveLabels() map[string]interface{} {
 	return result
 }
 
+func validateLabel(labelValue string) bool {
+	const maxLength = 256
+	labelPattern := regexp.MustCompile(regexLabelPattern)
+	if len(labelValue) > maxLength || !labelPattern.MatchString(labelValue) {
+		slog.Warn("Label value contains unsupported character ", "label_value", labelValue)
```
aphralG: The reason for failure here could be that the label is too long, could this message me updated to mention that too

internal/config/config_test.go
```diff
@@ -1410,3 +1410,54 @@ func createDefaultCollectorConfig() *Collector {
 		},
 	}
 }
+
+func TestValidateLabel(t *testing.T) {
+	tests := []struct {
+		name      string
+		input     string
+		wantValid bool
```
aphralG:
```suggestion
		expected bool
```

internal/config/config_test.go
```diff
@@ -1410,3 +1410,54 @@ func createDefaultCollectorConfig() *Collector {
...
+		{
+			name:      "Test 7: Invalid label - empty",
+			input:     "",
+			wantValid: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := validateLabel(tt.input)
```
aphralG:
```suggestion
			actual := validateLabel(tt.input)
```

internal/config/testdata/nginx-agent.conf
dhurley: yeah thats correct. We are enforcing more strict rules on what values labels can have.

internal/config/config_test.go
```diff
@@ -1410,3 +1410,54 @@ func createDefaultCollectorConfig() *Collector {
...
+		{
+			name:      "Test 6: Invalid label - ends with dash",
+			input:     "label-",
+			wantValid: false,
+		},
+		{
+			name:      "Test 7: Invalid label - empty",
+			input:     "",
+			wantValid: false,
+		},
```
sean-breen, line 1459: Can we add a test for checking that a label containing spaces is also invalid?

spencerugbo (approved):

PR #1389, opened 2025-11-12
NGINX Configs to Reference Remote External Files

Akshay2191:
Extends the ConfigApply capability to manage NGINX configurations that reference external resources hosted at remote URLs. The agent now handles downloading of these files into the NGINX configuration directory before applying the new configuration.

### Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1622,6 +1622,7 @@ func resolveExternalDataSource() *ExternalDataSource {
 	}
 
 	if err := validateAllowedDomains(externalDataSource.AllowedDomains); err != nil {
+		slog.Error("External data source not configured due to invalid configuration", "error", err)
 		return nil
 	}
 
@@ -1636,7 +1637,6 @@ func validateAllowedDomains(domains []string) error {
 	for _, domain := range domains {
 		// Validating syntax using the RFC-compliant regex
 		if !domainRegex.MatchString(domain) || domain == "" {
-			slog.Error("domain specified in allowed_domains is invalid", "domain", domain)
 			return errors.New("invalid domain found in allowed_domains")
 		}
 	}
--- a/internal/file/external_file_operator.go
+++ b/internal/file/external_file_operator.go
@@ -45,7 +45,7 @@ func (efo *ExternalFileOperator) DownloadExternalFile(ctx context.Context, fileA
 	permission := fileAction.File.GetFileMeta().GetPermissions()
 	fileName := fileAction.File.GetFileMeta().GetName()
 
-	slog.InfoContext(ctx, "Downloading external file from", "location", location)
+	slog.InfoContext(ctx, "Downloading external file", "location", location)
 
 	var contentToWrite []byte
 	var downloadErr, updateError error
@@ -150,7 +150,8 @@ func (efo *ExternalFileOperator) downloadFileContent(ctx context.Context, file *
 		headers.ETag = resp.Header.Get("ETag")
 		headers.LastModified = resp.Header.Get("Last-Modified")
 	case http.StatusNotModified:
-		slog.DebugContext(ctx, "File content unchanged (304 Not Modified)", "file_name", fileName)
+		slog.InfoContext(ctx, "Received 304 status code in response from external data source, "+
+			"file has not been modified", "file_name", fileName)
 		// return empty content but preserve headers if present
 		header := DownloadHeader{
 			ETag:         resp.Header.Get("ETag"),
--- a/internal/file/file_manager_service.go
+++ b/internal/file/file_manager_service.go
@@ -703,7 +703,7 @@ actionsLoop:
 		tempFilePath := tempFilePath(fileMeta.GetName())
 		switch fileAction.Action {
 		case model.Delete:
-			slog.DebugContext(ctx, "Deleting file", "file", fileMeta.GetName())
+			slog.InfoContext(ctx, "Deleting file", "file", fileMeta.GetName())
 			if err = os.Remove(fileMeta.GetName()); err != nil && !os.IsNotExist(err) {
 				actionError = fmt.Errorf("error deleting file: %s error: %w",
 					fileMeta.GetName(), err)
--- a/internal/file/file_plugin.go
+++ b/internal/file/file_plugin.go
@@ -7,6 +7,7 @@ package file
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
@@ -352,14 +353,16 @@ func (fp *FilePlugin) handleConfigApplyRequest(ctx context.Context, msg *bus.Mes
 		if rollbackErr != nil {
 			// include both the original apply error and the rollback error so the management plane
 			// receives actionable information about what failed during apply and what failed during rollback
-			combinedErr := fmt.Sprintf("apply error: %s; rollback error: %s", err.Error(), rollbackErr.Error())
+			applyErr := fmt.Errorf("config apply error: %w", err)
+			rbErr := fmt.Errorf("rollback error: %w", rollbackErr)
+			combinedErr := errors.Join(applyErr, rbErr)
 
 			rollbackResponse := fp.createDataPlaneResponse(
 				correlationID,
 				mpi.CommandResponse_COMMAND_STATUS_FAILURE,
 				"Config apply failed, rollback failed",
 				instanceID,
-				combinedErr,
+				combinedErr.Error(),
 			)
 
 			fp.messagePipe.Process(ctx, &bus.Message{Topic: bus.ConfigApplyCompleteTopic, Data: rollbackResponse})


internal/config/config.go
```diff
@@ -628,6 +648,11 @@ func registerClientFlags(fs *flag.FlagSet) {
 		DefMaxFileSize,
 		"Max file size in bytes.",
 	)
+	fs.Duration(
+		ClientFileDownloadTimeoutKey,
+		DefClientFileDownloadTimeout,
+		"Timeout value in seconds, to downloading file for config apply.",
```
```suggestion
		"Timeout value in seconds, to download a file during a config apply request.",
```

internal/config/config.go
```diff
@@ -1560,3 +1586,35 @@ func areCommandServerProxyTLSSettingsSet() bool {
...
+}
+
+func validateAllowedDomains(domains []string) error {
+	if len(domains) == 0 {
+		return nil
+	}
+
+	for _, domain := range domains {
+		if strings.ContainsAny(domain, "/\\ ") || domain == "" {
+			slog.Error("domain is not specified in allowed_domains")
```
dhurley: no need to log error here if you are returning the error anyways

internal/config/config.go
```diff
@@ -1560,3 +1586,35 @@ func areCommandServerProxyTLSSettingsSet() bool {
...
+	return externalDataSource
+}
+
+func validateAllowedDomains(domains []string) error {
+	if len(domains) == 0 {
+		return nil
+	}
+
+	for _, domain := range domains {
+		if strings.ContainsAny(domain, "/\\ ") || domain == "" {
```
dhurley: Is there a better way to validate if the string is a valid domain name or not? Maybe there is a regex we could use?

internal/config/defaults.go
```diff
@@ -114,6 +116,9 @@ const (
 
 	// File defaults
 	DefLibDir = "/var/lib/nginx-agent"
+
+	DefExternalDataSourceProxyUrl = ""
+	DefExternalDataSourceMaxBytes = 100 * 1024 * 1024
```
dhurley: can you add a comment to say what the default size is in MB

internal/file/file_service_operator.go
```diff
@@ -276,6 +282,23 @@ func (fso *FileServiceOperator) UpdateFile(
 	return fso.sendUpdateFileStream(ctx, fileToUpdate, fso.agentConfig.Client.Grpc.FileChunkSize)
 }
 
+func (fso *FileServiceOperator) RenameExternalFile(
```
dhurley: instead of duplicating the RenameFile function can we update the RenameFile function to just rename the file and have the hash validation in a separate function like this:

```
func (fso *FileServiceOperator) RenameAndValidateFile(
  ctx context.Context, source, destination string,
) error {
  err := fso.RenameFile(ctx, source, destination)
  if err != nil {
    return err
  }

 return fso.validateFileHash(desination, hash)
}
```

internal/file/file_manager_service.go
```diff
@@ -40,6 +44,11 @@ const (
 	executePerm = 0o111
 )
 
+type DownloadHeaders struct {
```
```suggestion
type DownloadHeader struct {
```

internal/file/file_manager_service.go
```diff
@@ -396,13 +408,16 @@ func (fms *FileManagerService) DetermineFileActions(
...
 			fileDiff[fileName] = modifiedFile
 
-			continue
 			// if file currently exists and file hash is different, file has been updated
 			// copy contents, set file action
 		} else if ok && modifiedFile.File.GetFileMeta().GetHash() != currentFile.GetFileMeta().GetHash() {
 			modifiedFile.Action = model.Update
 			fileDiff[fileName] = modifiedFile
 		}
+		if modifiedFile.File.GetExternalDataSource() != nil || currentFile.GetExternalDataSource() != nil {
```
dhurley: can you move this if statement up to just after line 404? It will save you from doing unnecessary `os.Stat`

internal/config/config.go
aphralG:
```suggestion
		"Timeout value in seconds, for downloading a file during a config apply.",
```

internal/config/config.go
```diff
@@ -489,6 +491,24 @@ func registerFlags() {
 	})
 }
 
+func registerExternalDataSourceFlags(fs *flag.FlagSet) {
+	fs.String(
+		ExternalDataSourceProxyUrlKey,
+		DefExternalDataSourceProxyUrl,
+		"Url to the proxy service to fetch the external file.",
```
aphralG:
```suggestion
		"Url to the proxy service for fetching external files.",
```

internal/config/config_test.go
```diff
@@ -1569,3 +1576,73 @@ func TestValidateLabel(t *testing.T) {
 		})
 	}
 }
+
```
aphralG, line 1633: Can you add a test for external Data source and add the external data source to TestResolveConfig() which checks all the values that can be set in the agent config

internal/file/file_manager_service.go
```diff
@@ -104,26 +114,28 @@ type FileManagerService struct {
...
 	fileActions map[string]*model.FileCache // key is file path
 	// map of the files currently on disk, used to determine the file action during config apply
-	currentFilesOnDisk    map[string]*mpi.File // key is file path
-	previousManifestFiles map[string]*model.ManifestFile
-	manifestFilePath      string
-	rollbackManifest      bool
-	filesMutex            sync.RWMutex
+	currentFilesOnDisk     map[string]*mpi.File // key is file path
+	previousManifestFiles  map[string]*model.ManifestFile
+	newExternalFileHeaders map[string]DownloadHeaders
```
```suggestion
	externalFileHeaders map[string]DownloadHeaders
```

internal/file/file_manager_service.go
```diff
@@ -396,13 +408,16 @@ func (fms *FileManagerService) DetermineFileActions(
 			modifiedFile.Action = model.Add
 			fileDiff[fileName] = modifiedFile
 
-			continue
 			// if file currently exists and file hash is different, file has been updated
```
aphralG: Why was the continue removed ?

internal/file/file_manager_service.go
```diff
@@ -233,7 +245,7 @@ func (fms *FileManagerService) Rollback(ctx context.Context, instanceID string)
 			delete(fms.currentFilesOnDisk, fileAction.File.GetFileMeta().GetName())
 
 			continue
-		case model.Delete, model.Update:
+		case model.Delete, model.Update, model.ExternalFile:
```
aphralG, line 256: With the external files are the files still considered updated, added or deleted? Just wondering instead should it be a field added to the file rather than an action this might fix a few of the things the linter had an issue with which required the checking of Adding and updating in the switch statement

internal/file/file_manager_service.go
```diff
@@ -792,3 +818,217 @@ func tempBackupFilePath(fileName string) string {
...
+		return nil, DownloadHeaders{}, fmt.Errorf("failed to execute download request for %s: %w", downloadURL, err)
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		headers.ETag = resp.Header.Get("ETag")
+		headers.LastModified = resp.Header.Get("Last-Modified")
+	case http.StatusNotModified:
+		slog.InfoContext(ctx, "File content unchanged (304 Not Modified)", "file_name", fileName)
```
```suggestion
		slog.DebugContext(ctx, "File content unchanged (304 Not Modified)", "file_name", fileName)
```

internal/file/file_manager_service.go
```diff
@@ -792,3 +818,217 @@ func tempBackupFilePath(fileName string) string {
...
+		if fileMeta.LastModified != "" {
+			req.Header.Set("If-Modified-Since", fileMeta.LastModified)
+		}
+	} else {
+		slog.DebugContext(ctx, "File not found in manifest or missing metadata; skipping conditional headers.",
+			"file", fileName)
+	}
+}
+
+func isWildcardMatch(hostname, pattern string) bool {
```
aphralG: Can the naming of this be changed

internal/file/file_manager_service.go
```diff
@@ -792,3 +818,217 @@ func tempBackupFilePath(fileName string) string {
...
+		slog.Debug("Failed to parse download URL for domain check", "url", downloadURL, "error", err)
+		return false
+	}
+
+	hostname := u.Hostname()
+	if hostname == "" {
+		return false
+	}
+
+	for _, pattern := range allowedDomains {
```
```suggestion
	for _, domain := range allowedDomains {
```

internal/file/file_manager_service_test.go
```diff
@@ -29,6 +35,39 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type fakeFSO struct {
```
aphralG: We already have a fake file operator in agent/internal/file/filefakes/fake_file_operator.go

internal/file/file_manager_service.go
```diff
@@ -792,3 +818,217 @@ func tempBackupFilePath(fileName string) string {
 	tempFileName := "." + filepath.Base(fileName) + ".agent.backup"
 	return filepath.Join(filepath.Dir(fileName), tempFileName)
 }
+
+func (fms *FileManagerService) handleExternalFileDownload(ctx context.Context, fileAction *model.FileCache,
+	tempFilePath string,
+) error {
```
```suggestion
func (fms *FileManagerService) downloadExternalFile(ctx context.Context, fileAction *model.FileCache,
	filePath string,
) error {
```

internal/file/file_manager_service.go
```diff
@@ -792,3 +818,219 @@ func tempBackupFilePath(fileName string) string {
...
+func (fms *FileManagerService) writeContentToTempFile(
+	ctx context.Context,
+	content []byte,
+	path string,
+) error {
+	writeErr := fms.fileOperator.Write(
+		ctx,
+		content,
+		path,
+		"0600",
```
dhurley: the file permissions should come from the file meta in the File proto message

internal/file/file_manager_service.go
Akshay2191: The initial check for ADD action caused the loop to exit early therefore I removed this statement.
However, with Donal's suggestion to prioritize the external file check first, the loop's control flow changed and have re added the continue statement back.

internal/file/file_manager_service.go
Akshay2191: Yes, external files still involve Add and Update actions. The challenge is that these actions are only confirmed after a call to the third-party server. Because we must make this external call every time a reference to the third-party file is encountered, we introduced a new flow that triggers based on the presence of the External File URL.

internal/config/config.go
Akshay2191: I have used this  regex now.
`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`

internal/file/file_service_operator.go
Akshay2191: The core issue with the current implementation is that the function call lacks necessary attributes to make a decision. Specifically, we should only proceed with the hash validation if File.GetExternalDataSource() returns an empty value.

internal/file/file_service_operator_test.go
```diff
@@ -187,3 +187,86 @@ func TestFileManagerService_UpdateFile_LargeFile(t *testing.T) {
...
 }
+
+func TestFileServiceOperator_RenameExternalFile(t *testing.T) {
+	tests := []struct {
+		prepare    func(t *testing.T) (src, dst string)
+		name       string
+		wantErrMsg string
+		wantErr    bool
+	}{
+		{
```
craigell: The following scenarios could increase coverage: 

-  No destination specified
-  Restricted directory. 
- Two files to the same destination.

internal/file/file_service_operator_test.go
```diff
@@ -187,3 +187,86 @@ func TestFileManagerService_UpdateFile_LargeFile(t *testing.T) {
...
+			name: "Test 1: success",
+			prepare: func(t *testing.T) (string, string) {
```
aphralG: Can this be changed to follow the way other tests are set up and formatted

internal/config/defaults.go
```diff
@@ -114,6 +116,10 @@ const (
 
 	// File defaults
 	DefLibDir = "/var/lib/nginx-agent"
+
+	DefExternalDataSourceProxyUrl = ""
+	// Default allow external data sources up to 100 MB
+	DefExternalDataSourceMaxBytes = 100 * 1024 * 1024
```
aphralG:
```suggestion
	DefExternalDataSourceMaxBytes = 100 * 1024 * 1024 // default 100MB
```

internal/file/file_manager_service_test.go
```diff
@@ -1173,3 +1180,337 @@ rQHX6DP4w6IwZY8JB8LS
...
+	diff, err := fileManagerService.DetermineFileActions(ctx, make(map[string]*mpi.File), modifiedFiles)
+	require.NoError(t, err)
+
+	fc, ok := diff[fileName]
+	require.True(t, ok, "expected file to be present in diff")
+	assert.Equal(t, model.ExternalFile, fc.Action)
+}
+
+//nolint:gocognit,revive,govet // cognitive complexity is 22
+func TestFileManagerService_downloadExternalFiles_Cases(t *testing.T) {
```
aphralG: Can this test be simplified and changed to match the way other tests are set up and structured

internal/file/file_manager_service_test.go
```diff
@@ -1173,3 +1180,337 @@ rQHX6DP4w6IwZY8JB8LS
...
+
+	assert.Equal(t, 1, fakeFSO.RenameExternalFileCallCount(), "RenameExternalFile should be called once")
+
+	_, srcArg, dstArg := fakeFSO.RenameExternalFileArgsForCall(0)
+
+	assert.Equal(t, tempPath, srcArg, "RenameExternalFile source argument mismatch")
+	assert.Equal(t, fileName, dstArg, "RenameExternalFile destination argument mismatch")
+}
+
+func TestDownloadFileContent_MaxBytesLimit(t *testing.T) {
```
aphralG: Can these tests please be combined

internal/file/file_manager_service_test.go
```diff
@@ -1173,3 +1180,337 @@ rQHX6DP4w6IwZY8JB8LS
...
+	_, _, err := fms.downloadFileContent(ctx, file)
+	require.Error(t, err)
+	if !strings.Contains(err.Error(), "invalid proxy URL configured") &&
+		!strings.Contains(err.Error(), "failed to execute download request") &&
+		!strings.Contains(err.Error(), "proxyconnect") {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
+func TestIsDomainAllowed_EdgeCases(t *testing.T) {
```
aphralG: Can this be changed to use structs and loop through instead of calling isDomainAllowed multiple times

internal/file/file_manager_service.go
```diff
@@ -792,3 +829,222 @@ func tempBackupFilePath(fileName string) string {
 	tempFileName := "." + filepath.Base(fileName) + ".agent.backup"
 	return filepath.Join(filepath.Dir(fileName), tempFileName)
 }
+
+func (fms *FileManagerService) downloadExternalFile(ctx context.Context, fileAction *model.FileCache,
```
aphralG: Could we add an extra operator maybe like the file operator to handle the external file download as this file is getting a little too large and complex

internal/file/file_manager_service.go
```diff
@@ -792,3 +829,222 @@ func tempBackupFilePath(fileName string) string {
...
+
+	fileName := fileAction.File.GetFileMeta().GetName()
+	fms.externalFileHeaders[fileName] = headers
+
+	updateErr := fms.writeContentToTempFile(ctx, contentToWrite, filePath, permission)
+
+	return updateErr
+}
+
+func (fms *FileManagerService) writeContentToTempFile(
```
aphralG: Does this need to be a separate function just on its own ?

internal/file/file_manager_service.go
```diff
@@ -792,3 +829,222 @@ func tempBackupFilePath(fileName string) string {
...
+	if hostname == "" {
+		return false
+	}
+
+	for _, domain := range allowedDomains {
+		if domain == "" {
+			continue
+		}
+
+		if domain == hostname {
```
aphralG: Would the below work to reduce the number of if statements 
```suggestion
		if domain == hostname || sMatchesWildcardDomain(hostname, domain) {
```

internal/file/file_manager_service_test.go
Akshay2191: I tried to consolidate these divergent test scenarios (MaxBytes vs. Proxy Errors) but it spiked the cognitive complexity to 23 due the very different validation steps. Therefore I kept them seperate.

internal/file/file_manager_service.go
```diff
@@ -614,10 +649,13 @@ func (fms *FileManagerService) downloadUpdatedFilesToTempLocation(ctx context.Co
 func (fms *FileManagerService) moveOrDeleteFiles(ctx context.Context, actionError error) error {
 actionsLoop:
 	for _, fileAction := range fms.fileActions {
+		var err error
+		fileMeta := fileAction.File.GetFileMeta()
+		tempFilePath := tempFilePath(fileAction.File.GetFileMeta().GetName())
```
spencerugbo: Could `fileMeta` be used more instead of `fileAction.File.GetFileMeta()` throughout this for loop?
```suggestion
		tempFilePath := tempFilePath(fileMeta.GetName())
```

internal/config/config.go
```diff
@@ -50,6 +50,10 @@ const (
 	regexLabelPattern = "^[a-zA-Z0-9]([a-zA-Z0-9-_]{0,254}[a-zA-Z0-9])?$"
 )
 
+var domainRegex = regexp.MustCompile(
```
spencerugbo, line 53: Could we group the two var declarations into one block?
```suggestion
var (
	domainRegex = regexp.MustCompile(
		`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`,
	)
	viperInstance = viper.NewWithOptions(viper.KeyDelimiter(KeyDelimiter))
)
```

dhurley (approved):

internal/config/config_test.go
```diff
@@ -1622,3 +1630,73 @@ func TestValidateLabel(t *testing.T) {
...
+func TestValidateAllowedDomains(t *testing.T) {
+	tests := []struct {
+		name    string
+		domains []string
+		wantErr bool
```
aphralG: Could this be changed to error so we check the specific error and its nil if no error is expected?

internal/file/file_manager_service_test.go
```diff
@@ -1197,3 +1197,31 @@ rQHX6DP4w6IwZY8JB8LS
 		})
 	}
 }
+
+func TestFileManagerService_DetermineFileActions_ExternalFile(t *testing.T) {
+	ctx := context.Background()
+	tempDir := t.TempDir()
```
aphralG, line 1229: Can a test for this be added to the current Determine file action tests ?

internal/file/file_manager_service.go
```diff
@@ -114,17 +125,23 @@ type FileManagerService struct {
 func NewFileManagerService(fileServiceClient mpi.FileServiceClient, agentConfig *config.Config,
 	manifestLock *sync.RWMutex,
 ) *FileManagerService {
-	return &FileManagerService{
+	fms := &FileManagerService{
```
```suggestion
	fileManagerService := &FileManagerService{
```

internal/file/file_manager_service.go
```diff
@@ -392,6 +409,16 @@ func (fms *FileManagerService) DetermineFileActions(
 			continue
 		}
 
+		// If it's external, we DON'T care about disk state or hashes here.
```
```suggestion
		// If it's external, we don't care about disk state or hashes here.
```

internal/file/file_manager_service.go
```diff
@@ -392,6 +409,16 @@ func (fms *FileManagerService) DetermineFileActions(
...
+		// We tag it as ExternalFile and let the downloader handle the rest.
+		if modifiedFile.File.GetExternalDataSource() != nil || (ok && currentFile.GetExternalDataSource() != nil) {
+			slog.DebugContext(ctx, "External file detected - flagging for fetch", "file_name", fileName)
```
```suggestion
			slog.DebugContext(ctx, "External file requires downloading", "file_name", fileName)
```

internal/file/file_manager_service.go
```diff
@@ -634,45 +662,63 @@ func (fms *FileManagerService) downloadUpdatedFilesToTempLocation(ctx context.Co
...
 			tempFilePath := tempFilePath(fileAction.File.GetFileMeta().GetName())
 
-			slog.DebugContext(
-				errGroupCtx,
-				"Downloading file to temp location",
-				"file", tempFilePath,
-			)
+			switch fileAction.Action {
+			case model.ExternalFile:
+				// delegate to the new operator
```
```suggestion
				
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+	"github.com/nginx/agent/v3/internal/model"
+)
+
+type ExternalFileOperator struct {
+	fms *FileManagerService
+}
+
+const (
+	sniffMaxBytes = 500
+	mimeSplitN    = 2
```
aphralG, line 32:
```suggestion
    sniffMaxBytes = 500
	mimeSplitN    = 2
```
Can the naming be updated to be more clear what these are ?

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/gabriel-vasile/mimetype"
+	mpi "github.com/nginx/agent/v3/api/grpc/mpi/v1"
+	"github.com/nginx/agent/v3/internal/model"
+)
+
+type ExternalFileOperator struct {
+	fms *FileManagerService
```
aphralG:
```suggestion
	fileManagerService *FileManagerService
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+) {
+	fileName := file.GetFileMeta().GetName()
+	downloadURL := file.GetExternalDataSource().GetLocation()
+	externalConfig := efo.fms.agentConfig.ExternalDataSource
+
+	if !efo.isDomainAllowed(downloadURL, externalConfig.AllowedDomains) {
+		return nil, DownloadHeader{}, fmt.Errorf("download URL %s is not in the allowed domains list", downloadURL)
+	}
+
+	u, err := url.Parse(downloadURL)
```
aphralG: Can a better name be given to this ?

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		headers.ETag = resp.Header.Get("ETag")
+		headers.LastModified = resp.Header.Get("Last-Modified")
+	case http.StatusNotModified:
+		slog.DebugContext(ctx, "File content unchanged (304 Not Modified)", "file_name", fileName)
+		// return empty content but preserve headers if present
+		h := DownloadHeader{
```
aphralG:
```suggestion
		header := DownloadHeader{
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+			LastModified: resp.Header.Get("Last-Modified"),
+		}
+
+		return nil, h, nil
+	default:
+		const maxErrBody = 4096
+		var bodyMsg string
+
+		limited := io.LimitReader(resp.Body, maxErrBody)
+		b, readErr := io.ReadAll(limited)
```
aphralG:
```suggestion
		body, readErr := io.ReadAll(limited)
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+		return nil, DownloadHeader{}, fmt.Errorf("failed to read content from response body: %w", err)
+	}
+
+	slog.InfoContext(ctx, "Successfully downloaded file content", "file_name", fileName, "size", len(content))
+
+	return content, headers, nil
+}
+
+func (efo *ExternalFileOperator) isDomainAllowed(downloadURL string, allowedDomains []string) bool {
+	u, err := url.Parse(downloadURL)
```
aphralG: Can another name be given to this parsedURL or something more clear

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,363 @@
...
+func (efo *ExternalFileOperator) validateDownloadedFile(content []byte, fileName string) error {
+	sniff := content
+	if len(sniff) > sniffMaxBytes {
+		sniff = sniff[:sniffMaxBytes]
+	}
+
+	detected := mimetype.Detect(sniff).String()
+	ext := strings.ToLower(filepath.Ext(fileName)) // includes leading dot or empty
+
+	// extension <-> detected MIME compatibility
```
aphralG: Can this comment be made more clear or removed ?

internal/config/config.go
```diff
@@ -1574,3 +1609,37 @@ func areCommandServerProxyTLSSettingsSet() bool {
...
+	}
+	externalDataSource := &ExternalDataSource{
+		ProxyURL:         proxyURLStruct,
+		AllowedDomains:   viperInstance.GetStringSlice(ExternalDataSourceAllowDomainsKey),
+		AllowedFileTypes: viperInstance.GetStringSlice(ExternalDataSourceAllowedFileTypesKey),
+		MaxBytes:         viperInstance.GetInt64(ExternalDataSourceMaxBytesKey),
+	}
+
+	if err := validateAllowedDomains(externalDataSource.AllowedDomains); err != nil {
+		return nil
```
valyria257, line 1626: should we be returning the error here?

internal/config/config.go
```diff
@@ -1574,3 +1609,37 @@ func areCommandServerProxyTLSSettingsSet() bool {
...
+func validateAllowedDomains(domains []string) error {
+	if len(domains) == 0 {
+		return nil
+	}
+
+	for _, domain := range domains {
+		// Validating syntax using the RFC-compliant regex
+		if !domainRegex.MatchString(domain) || domain == "" {
+			slog.Error("domain specified in allowed_domains is invalid", "domain", domain)
+			return errors.New("invalid domain found in allowed_domains")
```
valyria257: If we let the caller return the error, then this will already get logged [here](https://github.com/nginx/agent/blob/main/internal/app.go#L44)

```suggestion
			return fmt.Errorf("invalid domain %q found in allowed_domains", domain)
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,362 @@
...
+		// persist headers if present
+		if headers.ETag != "" || headers.LastModified != "" {
+			fileName := fileAction.File.GetFileMeta().GetName()
+			efo.fileManagerService.externalFileHeaders[fileName] = headers
+		}
+
+		return nil
+	}
+
+	fileName := fileAction.File.GetFileMeta().GetName()
```
valyria257: nit: can declare `fileName` earlier to avoid extra calls to `fileAction.File.GetFileMeta().GetName()`

internal/file/file_manager_service.go
```diff
@@ -97,15 +106,17 @@ type (
...
+	agentConfig          *config.Config
+	externalFileOperator *ExternalFileOperator
+	fileOperator         fileOperator
+	fileServiceOperator  fileServiceOperatorInterface
 	// map of files and the actions performed on them during config apply
 	fileActions map[string]*model.FileCache // key is file path
 	// map of the files currently on disk, used to determine the file action during config apply
 	currentFilesOnDisk    map[string]*mpi.File // key is file path
 	previousManifestFiles map[string]*model.ManifestFile
+	externalFileHeaders   map[string]DownloadHeader
```
valyria257, line 119: what is this map used for?

internal/config/config.go
Akshay2191: I designed the logic to ensure that an invalid domain prevents external variables from loading without crashing the agent. Since an MPI call without an external file makes these variables redundant, the agent should only fail when there is an external file URL  rather than its entire operation. This keeps core functionality intact even if the variable setup is incorrect.

internal/config/config.go
Akshay2191: To ensure the Agent remains operational despite incorrect external variable setup, I am intentionally handling the error locally rather than returning it to the main execution flow.

internal/file/file_manager_service.go
Akshay2191: Following our initial design discussion, these fields track the ETAG and Last-Modified timestamps for external files. While originally external file code was part of the file_manager_service, this logic was not moved during refactoring and kept at file level is now utilised by the external_file_operator.

internal/config/config.go
```diff
@@ -1574,3 +1609,37 @@ func areCommandServerProxyTLSSettingsSet() bool {
...
+
+func validateAllowedDomains(domains []string) error {
+	if len(domains) == 0 {
+		return nil
+	}
+
+	for _, domain := range domains {
+		// Validating syntax using the RFC-compliant regex
+		if !domainRegex.MatchString(domain) || domain == "" {
+			slog.Error("domain specified in allowed_domains is invalid", "domain", domain)
```
dhurley: Can you remove this log line and have a log message in the resolveExternalDataSource function instead?

internal/config/config.go
dhurley, line 1626: Can you add a log line here like this
```
slog.Error("External data source not configured due to invalid configuration", "error", err)
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (efo *ExternalFileOperator) DownloadExternalFile(ctx context.Context, fileAction *model.FileCache,
+	filePath string,
+) error {
+	location := fileAction.File.GetExternalDataSource().GetLocation()
+	permission := fileAction.File.GetFileMeta().GetPermissions()
+	fileName := fileAction.File.GetFileMeta().GetName()
+
+	slog.InfoContext(ctx, "Downloading external file from", "location", location)
```
dhurley:
```suggestion
	slog.InfoContext(ctx, "Downloading external file", "location", location)
```

internal/file/external_file_operator.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil, DownloadHeader{}, fmt.Errorf("failed to execute download request for %s: %w", downloadURL, err)
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		headers.ETag = resp.Header.Get("ETag")
+		headers.LastModified = resp.Header.Get("Last-Modified")
+	case http.StatusNotModified:
+		slog.DebugContext(ctx, "File content unchanged (304 Not Modified)", "file_name", fileName)
```
dhurley:
```suggestion
		slog.InfoContext(ctx, "Received 304 status code in response from external data source, file has not been modified", "file_name", fileName)
```
I think this would be better as info instead for debug

internal/file/file_manager_service.go
```diff
@@ -644,45 +672,62 @@ func (fms *FileManagerService) downloadUpdatedFilesToTempLocation(ctx context.Co
...
 actionsLoop:
 	for _, fileAction := range fms.fileActions {
+		var err error
+		fileMeta := fileAction.File.GetFileMeta()
+		tempFilePath := tempFilePath(fileMeta.GetName())
 		switch fileAction.Action {
 		case model.Delete:
-			slog.InfoContext(ctx, "Deleting file", "file", fileAction.File.GetFileMeta().GetName())
-			if err := os.Remove(fileAction.File.GetFileMeta().GetName()); err != nil && !os.IsNotExist(err) {
+			slog.DebugContext(ctx, "Deleting file", "file", fileMeta.GetName())
```
```suggestion
			slog.InfoContext(ctx, "Deleting file", "file", fileMeta.GetName())
```

internal/file/file_plugin.go
```diff
@@ -349,12 +350,16 @@ func (fp *FilePlugin) handleConfigApplyRequest(ctx context.Context, msg *bus.Mes
 			instanceID,
 		)
 		if rollbackErr != nil {
+			// include both the original apply error and the rollback error so the management plane
+			// receives actionable information about what failed during apply and what failed during rollback
+			combinedErr := fmt.Sprintf("apply error: %s; rollback error: %s", err.Error(), rollbackErr.Error())
```
dhurley: Would it be better to use `errors.Join` function here?

PR #1395, opened 2025-11-18
Ensure only files are in file overview in config apply request

dhurley:
### Proposed changes

Ensure only files are in file overview in config apply request.

### Checklist

### DIFF
--- a/internal/file/file_manager_service.go
+++ b/internal/file/file_manager_service.go
@@ -398,7 +398,7 @@ func (fms *FileManagerService) DetermineFileActions(
 		// If file doesn't exist on disk.
 		// Treat it as adding a new file.
 		if errors.Is(statErr, os.ErrNotExist) {
-			slog.DebugContext(ctx, "New untracked file to created", "file_name", fileName)
+			slog.DebugContext(ctx, "New untracked file needs to be created", "file_name", fileName)
 			modifiedFile.Action = model.Add
 			fileDiff[fileName] = modifiedFile
 


internal/file/file_manager_service.go
```diff
@@ -373,26 +375,52 @@ func (fms *FileManagerService) DetermineFileActions(
...
 			fileDiff[fileName] = modifiedFile
 
 			continue
-			// if file currently exists and file hash is different, file has been updated
-			// copy contents, set file action
-		} else if ok && modifiedFile.File.GetFileMeta().GetHash() != currentFile.GetFileMeta().GetHash() {
-			modifiedFile.Action = model.Update
+		}
+
+		// If file doesn't exist on disk.
```
vrmare, line 398:
```suggestion
	// File doesn't exist on disk → ADD
    fileStats, statErr := os.Stat(fileName)
    if errors.Is(statErr, os.ErrNotExist) {
        slog.DebugContext(ctx, "New untracked file to created", "file_name", fileName)
        modifiedFile.Action = model.Add
        fileDiff[fileName] = modifiedFile
        continue
    }
    
    // Early return on stat error (not just NotExist)
    if statErr != nil {
        return nil, fmt.Errorf("unable to stat file %s: %w", fileName, statErr)
    }
    
    // Directory conflict → error
    if fileStats.IsDir() {
        return nil, fmt.Errorf("unable to create file %s since a directory with the same name already exists", fileName)
    }
    
    // Untracked file with different hash → UPDATE
    metadataOfFileOnDisk, err := files.FileMeta(fileName)
    if err != nil {
        return nil, fmt.Errorf("unable to get file metadata for %s: %w", fileName, err)
    }
    
    if metadataOfFileOnDisk.GetHash() != modifiedFile.File.GetFileMeta().GetHash() {
        slog.DebugContext(ctx, "Untracked file requires updating", "file_name", fileName)
        modifiedFile.Action = model.Update
        fileDiff[fileName] = modifiedFile
    }
		
```

vrmare: Could reduce the nesting for improved cognitive complexity. Each branch could end with continue instead of deeper nesting (else if). Early returns are great to have.

internal/file/file_manager_service.go
```diff
@@ -373,26 +375,52 @@ func (fms *FileManagerService) DetermineFileActions(
 	for _, modifiedFile := range modifiedFiles {
 		fileName := modifiedFile.File.GetFileMeta().GetName()
 		currentFile, ok := filesMap[fileName]
-		// default to unchanged action
 		modifiedFile.Action = model.Unchanged
 
-		// if file is unmanaged, action is set to unchanged so file is skipped when performing actions
+		// If file is unmanaged, action is set to unchanged so file is skipped when performing actions.
```
vrmare, line 380: I think this function can be refactored for simplicity? I imagine existing manifest (source of truth) is maintained as a virtual FS in memory. Then single loop over modified files can be applied to the VFS. 

Something like this:

```
type VirtualFS struct {
    files map[string]SomeFileCache
    mu    sync.RWMutex
}

func (vfs *VirtualFS) ApplyChanges(modifiedFiles map[string]SomeFileCache) ([]FileAction, error) {}
```

The idea is that VFS holds canonical state.

vrmare: A bigger change I believe and happy to discuss outside this PR.

internal/file/file_manager_service.go
dhurley: Thanks for the feedback. I'll make that update as I was missing one of the error scenarios

internal/file/file_manager_service.go
dhurley: Yeah we can take a look at this and do any changes in a separate PR

internal/file/file_manager_service.go
```diff
@@ -373,24 +375,58 @@ func (fms *FileManagerService) DetermineFileActions(
...
+
+			continue
+		}
+
+		fileStats, statErr := os.Stat(fileName)
+
+		// If file doesn't exist on disk.
+		// Treat it as adding a new file.
+		if errors.Is(statErr, os.ErrNotExist) {
+			slog.DebugContext(ctx, "New untracked file to created", "file_name", fileName)
```
```suggestion
			slog.DebugContext(ctx, "New untracked file needs creating", "file_name", fileName)
```

PR #1400, opened 2025-11-19
[CI/CD] Improve assertion doc generation

sean-breen:
### Proposed changes

Adds the assertion document generation as an option when running the release branch workflow.

For the assertion doc, we need to generate a few extra files to pass along, mainly:

- sha checksum
- build start time
- build end time

These get stored in files alongside the binary which is created in `make package`, and these files along with the binary itself are stored to GitHub artifacts.

When 'Store assertion document' option is selected:
- The Assertion document workflow then downloads the artifacts from the calling pipeline, runs the dependency analysis and generates the document.

When running the Assertion doc workflow manually:
- Supply the `Run ID` of the pipeline which created the build artifacts, if they are still in GitHub the workflow will download these, and proceed with the workflow as normal.

### Checklist


.github/workflows/release-branch.yml
```diff
@@ -189,24 +193,21 @@ jobs:
...
-        if: ${{ needs.vars.outputs.tag_release == 'true' }}
-        run: |
           git push origin "v${{ inputs.packageVersion }}"
 
-  upload-packages:
-    name: Upload packages
-    runs-on: ubuntu-22.04-amd64
+  build-and-upload-packages:
+    name: Build and upload release packages
+    runs-on: ubuntu-22.04
```
dhurley: any reason for changing the runner?

sean-breen: Whoops, will change this back to our self-hosted one 👍

sean-breen: done

PR #1402, opened 2025-11-20
Remote Labels Configuration 

aphralG:
### Proposed changes

- Instead of sending a message on the bus to inform the plugins of changes to the agent config, added a Reconfigure function to the plugins. This allows the message bus before it processes a message to check if its a config update, if it is it validates and updates the new changes, then call reconfigure on each of the plugins and handles the data plane response. 
- Add ability to modify labels when a change is made to the labels the otel collector if one exists is also updated and restarted with the new changes 

### Checklist

### DIFF
--- a/api/grpc/mpi/v1/helpers.go
+++ b/api/grpc/mpi/v1/helpers.go
@@ -61,7 +61,7 @@ func ConvertToMap(input []*structpb.Struct) map[string]any {
 			case *structpb.Value_BoolValue:
 				convertedMap[key] = field.GetBoolValue()
 			default:
-				slog.Warn("unknown type for map conversion", "value", kind)
+				slog.Warn("Unknown type for map conversion", "value", kind)
 			}
 		}
 	}
--- a/internal/collector/otel_collector_plugin.go
+++ b/internal/collector/otel_collector_plugin.go
@@ -226,6 +226,7 @@ func (oc *Collector) Reconfigure(ctx context.Context, agentConfig *config.Config
 			oc.restartCollector(ctx)
 		}
 	} else {
+		slog.InfoContext(ctx, "OTel collector headers have been added, restarting collector")
 		oc.config = agentConfig
 		oc.restartMutex.Lock()
 		defer oc.restartMutex.Unlock()


api/grpc/mpi/v1/helpers_test.go
```diff
@@ -70,3 +70,45 @@ func TestConvertToStructs(t *testing.T) {
...
 }
+
+func TestConvertToMaps(t *testing.T) {
+	tests := []struct {
+		name     string
+		expected map[string]any
+		input    []*structpb.Struct
+	}{
+		{
+			name: "Test 1: Valid input with simple key-value pairs",
```
sean-breen, line 81: Missing cases for slices, and structs

craigell: Should there be a test for invalid values? Empty values, blank "" values and special characters. 

For valid scenario, should we include "config-sync-group" key?

sean-breen: Tests for this function should be pretty generic, the exact strings we place in the tests shouldn't really matter

spencerugbo (approved):

api/grpc/mpi/v1/helpers_test.go
aphralG: This function is only to covert to a map validation will be done elsewhere

api/grpc/mpi/v1/helpers.go
```diff
@@ -45,3 +47,24 @@ func ConvertToStructs(input map[string]any) ([]*structpb.Struct, error) {
...
+			kind := field.GetKind()
+			switch kind.(type) {
+			case *structpb.Value_StringValue:
+				convertedMap[key] = field.GetStringValue()
+			case *structpb.Value_NumberValue:
+				convertedMap[key] = int(field.GetNumberValue())
+			case *structpb.Value_BoolValue:
+				convertedMap[key] = field.GetBoolValue()
+			default:
+				slog.Warn("unknown type for map conversion", "value", kind)
```
dhurley:
```suggestion
				slog.Warn("Unknown type for map conversion", "key", key, "value", kind)
```

internal/collector/otel_collector_plugin.go
```diff
@@ -206,10 +206,35 @@ func (oc *Collector) Subscriptions() []string {
...
+		if !reflect.DeepEqual(oc.config.Collector.Extensions.HeadersSetter.Headers,
+			agentConfig.Collector.Extensions.HeadersSetter.Headers) {
+			slog.InfoContext(ctx, "OTel collector headers have changed, restarting collector")
+			oc.config = agentConfig
+			oc.restartMutex.Lock()
+			defer oc.restartMutex.Unlock()
+			oc.restartCollector(ctx)
+		}
+	} else {
+		oc.config = agentConfig
```
dhurley, line 230: Can you add an info log here saying that new headers have being added and the collector is bing restarted

PR #1404, opened 2025-11-21
Add script to check container images' agent versions

sean-breen:
### Proposed changes

Adds a script `check-images.sh` which can check the agent version and nginx version of all image tags matching a pattern.

Example:

```shell
> scripts/check-images.sh agentv3 alpine
Checking images in docker-registry.nginx.com/nginx/agentv3
892 tags fetched.
165 tags after filtering.
tags matching 'alpine': 81
:: docker-registry.nginx.com/nginx/agentv3:1-alpine
nginx version: nginx/1.29.3
nginx-agent version v3.5.0-34a13e6
...
```
will return all the tags found for the `agentv3` image which match the pattern `alpine`, and run the commands:
- `nginx -v`
- `nginx-agent -v`

### Checklist

### DIFF
--- a/.github/actions/az-sync/action.yml
+++ b/.github/actions/az-sync/action.yml
@@ -1,5 +1,0 @@
-name: az-sync
-author: s.breen
-description: az-sync
-runs:
-  using:


scripts/check-images.sh
```diff
@@ -0,0 +1,41 @@
...
+# Filter out tags that end with three or more digits (nightly/build tags)
+grep -Ev '\d{3,}$' all_tags.txt | sort -u > filtered_tags.txt
+echo $(wc -l < filtered_tags.txt) "tags after filtering."
+
+FOUND=($(grep -E "${search}" filtered_tags.txt | sort)) || { echo "No tags found matching '${search}'"; exit 1; }
+echo "tags matching '${search}':" ${#FOUND[@]}
+
+for tag in "${FOUND[@]}"; do
+    echo ":: ${REGISTRY_URL}/${IMAGE_NAME}:$tag"
+    podman pull ${REGISTRY_URL}/${IMAGE_NAME}:$tag > /dev/null 2>&1
```
dhurley: Can you add a check to see if docker or podman is installed? Some people might use docker over podman

scripts/check-images.sh
```diff
@@ -0,0 +1,41 @@
...
+}
+
+if [[ -z "${IMAGE_NAME}" ]]; then
+    usage
+fi
+
+echo "Checking images in ${REGISTRY_URL}/${IMAGE_NAME}"
+
+# Fetch all tags from the remote registry
+skopeo list-tags docker://${REGISTRY_URL}/${IMAGE_NAME} | jq -r '.Tags[]' > all_tags.txt
```
dhurley: Would it be possible to use the skopeo docker image instead? https://github.com/containers/skopeo/blob/main/install.md#container-images

Would save people having to install skopeo on their machines

sean-breen: True yeah I'll add that

scripts/check-images.sh
sean-breen: Sure

scripts/check-images.sh
sean-breen: Do we want this implemented as a Github workflow in that case?

aphralG (approved):

scripts/check-images.sh
dhurley: dont think we need a github workflow. The script by itself is fine

sean-breen: Added the skopeo image

scripts/check-images.sh
sean-breen: Defaults to `docker` if available, if not it checks for `podman`

scripts/check-images.sh
```diff
@@ -0,0 +1,82 @@
...
+        echo "Neither docker nor podman could be found. Please install one of them to proceed."
+        exit 1
+    fi
+fi
+
+echo "Using container tool: ${CONTAINER_TOOL}"
+${CONTAINER_TOOL} --version
+
+echo "Getting skopeo tool..."
+${CONTAINER_TOOL} pull docker://${SKOPEO_IMAGE}
```
dhurley: Getting this error when pulling image using docker
```
Getting skopeo tool...
invalid reference format
```

PR #1417, opened 2025-11-27
Ensure Unmanaged Field Is Stored Correctly In Manifest 

aphralG:
### Proposed changes
If config applys were preformed in quick succession there was a chance that unmanaged files would be marked as managed. 
This updates the resource plugin so that when it creates the config context it then reads the manifest to determine what files are unmanaged and updates them in the config context.

Also fixes issue where file watcher wasn't properly disabled. 

### Checklist

### DIFF
--- a/api/grpc/mpi/v1/command.pb.go
+++ b/api/grpc/mpi/v1/command.pb.go
@@ -85,7 +85,7 @@ func (x InstanceHealth_InstanceHealthStatus) Number() protoreflect.EnumNumber {
 
 // Deprecated: Use InstanceHealth_InstanceHealthStatus.Descriptor instead.
 func (InstanceHealth_InstanceHealthStatus) EnumDescriptor() ([]byte, []int) {
-	return file_mpi_v1_command_proto_rawDescGZIP(), []int{8, 0}
+	return file_mpi_v1_command_proto_rawDescGZIP(), []int{9, 0}
 }
 
 // the types of instances possible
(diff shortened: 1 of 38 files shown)


internal/watcher/file/file_watcher_service.go
```diff
@@ -83,7 +84,11 @@ func (fws *FileWatcherService) Watch(ctx context.Context, ch chan<- FileUpdateMe
 
 			return
 		case <-instanceWatcherTicker.C:
-			fws.checkForUpdates(ctx, ch)
+			if fws.enabled.Load() {
+				fws.checkForUpdates(ctx, ch)
+			} else {
+				slog.DebugContext(ctx, "Skipping check for file updates, file watcher is disabled")
```
sean-breen, line 90: Do we need to log this every tick? It might be more beneficial to just log once whenever the watcher is enabled or disabled

internal/resource/resource_service.go
```diff
@@ -332,6 +334,51 @@ func (r *ResourceService) UpdateHTTPUpstreamServers(ctx context.Context, instanc
...
+				}
+			}
+		}
+	}
+
+	return nginxConfigContext
+}
+
+func (r *ResourceService) manifestFile() (map[string]*model.ManifestFile, error) {
+	if _, err := os.Stat(r.agentConfig.LibDir + "/manifest.json"); err != nil {
```
sean-breen: Could the manifest file name be stored in a variable, maybe as a const somewhere? It will save us having to change it in a lot of places if the name changes in future

internal/watcher/file/file_watcher_service.go
aphralG: I copied the same behaviour as the instance watcher, this wont be logged every tick it will only be logged if we are skipping a file update that hasnt cleared the channel by the time we disable it.

internal/resource/resource_service.go
aphralG: This whole function will be removed in a PR that is coming up soon to remove the file plugin, I also don't see us changing the name of the manifest as it would cause a lot of confusion

internal/resource/resource_service.go
```diff
@@ -332,6 +334,51 @@ func (r *ResourceService) UpdateHTTPUpstreamServers(ctx context.Context, instanc
 	return added, updated, deleted, createPlusAPIError(updateError)
 }
 
+func (r *ResourceService) updateConfigContextFiles(ctx context.Context,
```
Akshay2191, line 339: Can we add test for this?

PR #1418, opened 2025-11-28
Set a limit on the number of access logs Agent can tail

spencerugbo:
### Proposed changes

Limit the number of access logs that Agent tails by default in order to reduce overhead. The limit can be configured in the agent conf as seen in the following example:

```
log:
  # set log level (error, warn, info, debug; default "info")
  level: info
  # set log path. if empty, don't log to file.
  path: /var/log/nginx-agent/

allowed_directories: 
    - /etc/nginx
    - /etc/app_protect
    - /usr/local/etc/nginx
    - /usr/share/nginx/modules
    - /var/run/nginx
    - /var/log/nginx

data_plane_config:
    nginx:
        # set the max access log files to be watched (default 10)
        max_access_log_files: 5
```

### Checklist


internal/config/config.go
```diff
@@ -158,6 +158,7 @@ func ResolveConfig() (*Config, error) {
 		Labels:             resolveLabels(),
 		LibDir:             viperInstance.GetString(LibDirPathKey),
 		SyslogServer:       resolveSyslogServer(),
+		MaxAccessLogFiles:  viperInstance.GetInt(MaxAccessLogFilesKey),
```
dhurley: can you move this field under the NginxDataPlaneConfig struct instead?

internal/datasource/config/nginx_config_parser.go
```diff
@@ -228,7 +228,13 @@ func (ncp *NginxConfigParser) createNginxConfigContext(
 					if !ncp.ignoreLog(directive.Args[0]) {
 						accessLog := ncp.accessLog(directive.Args[0], ncp.accessLogDirectiveFormat(directive),
 							formatMap)
-						nginxConfigContext.AccessLogs = ncp.addAccessLog(accessLog, nginxConfigContext.AccessLogs)
+						if ncp.agentConfig.IsMaxAccessLogFilesConfigured() {
```
dhurley: no need for the if statement here. Viper will set a default for MaxAccessLogFiles if none is set in the agent config

internal/datasource/config/nginx_config_parser.go
```diff
@@ -344,7 +350,7 @@ func (ncp *NginxConfigParser) parseIncludeDirective(
 }
 
 func (ncp *NginxConfigParser) addAccessLog(accessLog *model.AccessLog,
-	accessLogs []*model.AccessLog,
+	accessLogs []*model.AccessLog, maxAccessLogFiles int,
```
dhurley: you can remove the extra argument here since you can access the maxAccessLogFiles from the ncp.agentConfig.MaxAccessLogFiles

internal/datasource/config/nginx_config_parser.go
```diff
@@ -361,6 +367,11 @@ func (ncp *NginxConfigParser) addAccessLog(accessLog *model.AccessLog,
 		}
 	}
 
+	if len(accessLogs) >= maxAccessLogFiles {
+		slog.Warn("Maximum access log files have been reached, additional logs will be skipped")
```
dhurley:
```suggestion
		slog.Warn("Maximum access log files have been reached, unable to monitor access log", "access_log", accessLog.Name)
```

aphralG (commented):
Could you update `TestResolveConfig()` to also check for MaxAccessLogFiles

PR #1442, opened 2025-12-11
Write manifest file updates to a temp file first

dhurley:
### Proposed changes

Write manifest file updates to a temp file first, then validate that the temp file contains validate JSON before replacing the current manifest file with the temp one.

### Checklist

### DIFF
--- a/internal/file/file_operator.go
+++ b/internal/file/file_operator.go
@@ -174,13 +174,13 @@ func (fo *FileOperator) WriteManifestFile(
 	tempManifestFilePath := manifestPath + ".tmp"
 	tempFile, err := os.OpenFile(tempManifestFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
 	if err != nil {
-		return fmt.Errorf("failed to open temporary manifest file: %w", err)
+		return fmt.Errorf("failed to open file %s: %w", tempManifestFilePath, err)
 	}
 
 	if _, err = tempFile.Write(manifestJSON); err != nil {
 		closeFile(ctx, tempFile)
 
-		return fmt.Errorf("failed to write to temporary manifest file: %w", err)
+		return fmt.Errorf("failed to write to file %s: %w", tempManifestFilePath, err)
 	}
 
 	closeFile(ctx, tempFile)
@@ -188,7 +188,7 @@ func (fo *FileOperator) WriteManifestFile(
 	// Verify the contents of the temporary file is JSON
 	file, err := os.ReadFile(tempManifestFilePath)
 	if err != nil {
-		return fmt.Errorf("failed to read temporary manifest file: %w", err)
+		return fmt.Errorf("failed to read file %s: %w", tempManifestFilePath, err)
 	}
 
 	var manifestFiles map[string]*model.ManifestFile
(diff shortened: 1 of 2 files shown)


internal/file/file_operator.go
```diff
@@ -164,28 +164,50 @@ func (fo *FileOperator) WriteManifestFile(
...
 	}
 
-	// 0600 ensures only root can read/write
-	newFile, err := os.OpenFile(manifestPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	// Write to a temporary file first to ensure atomicity
+	tempManifestFilePath := manifestPath + ".tmp"
+	tempFile, err := os.OpenFile(tempManifestFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
 	if err != nil {
-		return fmt.Errorf("failed to read manifest file: %w", err)
+		return fmt.Errorf("failed to open temporary manifest file: %w", err)
```
sean-breen: Would it be better if we added the file path to the log message like so? Removes reference to the `temporary file` part as the file will have the `.tmp` extension anyway.
```suggestion
		return fmt.Errorf("failed to open file %s: %w", tempManifestFilePath, err)
```

internal/file/file_operator.go
```diff
@@ -164,28 +164,50 @@ func (fo *FileOperator) WriteManifestFile(
...
+		if len(file) == 0 {
+			return fmt.Errorf("temporary manifest file is empty: %w", err)
+		}
+
+		return fmt.Errorf("failed to parse temporary manifest file: %w", err)
+	}
+
+	// Rename the temporary file to the actual manifest file path
+	if renameError := os.Rename(tempManifestFilePath, manifestPath); renameError != nil {
+		return fmt.Errorf("failed to rename temporary manifest file: %w", renameError)
```
sean-breen: Could we add some tests around these error cases?
- OpenFile fails
- Write fails - ensure file is closed and original is unmodified
- Rename fails
- ensure .tmp file is cleaned up

- Manifest file is empty
- Manifest file is missing
- Manifest file exists, is named `manifest.json` but contains JSON objects that do not match our expectations
- Manifest file exists, is named `manifest.json` but is not actually JSON format (e.g .YAML)

dhurley: added more tests

sean-breen (approved):

internal/file/file_operator.go
```diff
@@ -164,28 +164,50 @@ func (fo *FileOperator) WriteManifestFile(
...
+	if err != nil {
+		if len(file) == 0 {
+			return fmt.Errorf("file %s is empty: %w", tempManifestFilePath, err)
+		}
+
+		return fmt.Errorf("failed to parse file %s: %w", tempManifestFilePath, err)
+	}
+
+	// Rename the temporary file to the actual manifest file path
+	if renameError := os.Rename(tempManifestFilePath, manifestPath); renameError != nil {
```
Akshay2191, line 206: Can we use the RenameFile function from file service operator here?

dhurley: that RenameFile function is doing extra things that I don't require here thats why I didn't use it

sean-breen (approved):

internal/file/file_operator.go
```diff
@@ -164,28 +164,50 @@ func (fo *FileOperator) WriteManifestFile(
...
+
 	// 0755 allows read/execute for all, write for owner
 	if err = os.MkdirAll(manifestDir, dirPerm); err != nil {
 		return fmt.Errorf("unable to create directory %s: %w", manifestDir, err)
 	}
 
-	// 0600 ensures only root can read/write
-	newFile, err := os.OpenFile(manifestPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	// Write to a temporary file first to ensure atomicity
+	tempManifestFilePath := manifestPath + ".tmp"
```
aphralG, line 174: does this need to be agent.tmp like other temp files /

internal/file/file_operator_test.go
```diff
@@ -44,13 +45,89 @@ func TestFileOperator_Write(t *testing.T) {
 	assert.Equal(t, fileContent, data)
 }
 
+func TestFileOperator_WriteManifestFile(t *testing.T) {
```
aphralG, line 48: Could you add a test around the parsing failing

internal/file/file_operator.go
dhurley: I'd didn't do that since this is a nginx-agent file and not a nginx file.

internal/file/file_operator_test.go
dhurley: Wasn't able to write a test for that scenario since the WriteManifestFile function is the one writing and reading the data. I have no way to send bad data to the function

aphralG (approved):

PR #1447, opened 2025-12-15
Refactor agent logging

dhurley:
### Proposed changes

- Refactor what gets logged when log level is set to info.
- Add RequestType to DataPlaneResponse proto message.
- Log health requests as debug instead of info.

### Checklist


internal/file/file_manager_service.go
```diff
@@ -623,7 +633,7 @@ func (fms *FileManagerService) downloadUpdatedFilesToTempLocation(ctx context.Co
 	}
 
 	if len(downloadFiles) == 0 {
-		slog.DebugContext(ctx, "No updated files to download")
+		slog.InfoContext(ctx, "No files are required to be downloaded")
```
```suggestion
		slog.InfoContext(ctx, "No files require downloading")
```

internal/file/file_manager_service.go
aphralG, line 674: Can we add a log for updating and adding for consistency ?

internal/file/file_service_operator.go
```diff
@@ -283,7 +290,7 @@ func (fso *FileServiceOperator) UpdateFile(
 func (fso *FileServiceOperator) RenameFile(
 	ctx context.Context, hash, source, desination string,
 ) error {
-	slog.DebugContext(ctx, fmt.Sprintf("Renaming file %s to %s", source, desination))
+	slog.InfoContext(ctx, fmt.Sprintf("Renaming file %s to %s", source, desination))
```
aphralG, line 293: Does this need to be info ? seems like if there was a big config apply it could spam the logs unnecessarily

internal/grpc/grpc.go
```diff
@@ -83,7 +83,7 @@ func NewGrpcConnection(ctx context.Context, agentConfig *config.Config,
 
 	serverAddr := serverAddress(ctx, commandConfig)
 
-	slog.InfoContext(ctx, "Dialing grpc server", "server_addr", serverAddr)
+	slog.InfoContext(ctx, "Creating connection to management plane server", "server_address", serverAddr)
```
aphralG, line 86: Could the type also be logged, auxiliary or command ?

internal/watcher/instance/instance_watcher_service.go
```diff
@@ -270,6 +273,18 @@ func (iw *InstanceWatcherService) instanceUpdates(ctx context.Context) (
 		return instanceUpdates, err
 	}
 
+	if !slices.EqualFunc(iw.processCache, nginxProcesses, func(a, b *nginxprocess.Process) bool {
```
aphralG, line 276: is this still needed ?

internal/file/file_manager_service.go
dhurley: The RenameFile function is already logging that

internal/file/file_service_operator.go
dhurley: Yeah I think its better to log this at info so that the user knows exactly what files have being changed on the data plane by the agent. Even if a user is doing a big config apply I don't think the extra logging will be an issue.

internal/grpc/grpc.go
dhurley: Yeah I'll add the server type to the context so that it gets logged here

internal/watcher/instance/instance_watcher_service.go
dhurley: yeah I think it is useful to log the NGINX process info when in debug if there is any changes. This will help with debugging issues in the future

internal/file/file_manager_service.go
aphralG: Ok, just not sure renaming makes it clear that the file is being updated or added the way that its clear a file is being deleted. I wouldn't be able to tell immediately the action being preformed

PR #1493, opened 2026-01-19
Fixed Agent crashing in AWS Fargate

Akshay2191:
Currently, the Agent fails to start in AWS Fargate environments. This occurs because Fargate abstracts the host filesystem, causing traditional container ID retrieval methods—which rely on specific **/proc/self/mountinfo** formats or the presence of standard Docker files—to return empty values. Without a valid Container ID, the Agent cannot unique identify itself to the NGINX One console and terminates.

As a fix, the Agent now detects the Fargate runtime and utilizes the **AWS Metadata Endpoint** to reliably retrieve the container ID.

### DIFF
--- a/pkg/host/info.go
+++ b/pkg/host/info.go
@@ -223,33 +223,25 @@ func (i *Info) releaseInfo(ctx context.Context, osReleaseLocation string) (*v1.R
 
 // containerID returns the container ID of the current running environment.
 func (i *Info) containerID(ctx context.Context) (string, error) {
-	var (
-		containerIDMount string
-		errMount         error
-	)
+	var errs error
 
-	containerIDMount, errMount = containerIDFromMountInfo(i.mountInfoLocation)
-	if containerIDMount != "" {
+	// Try to get container ID from mount info first
+	if containerIDMount, err := containerIDFromMountInfo(i.mountInfoLocation); err == nil && containerIDMount != "" {
 		return uuid.NewMD5(uuid.NameSpaceDNS, []byte(containerIDMount)).String(), nil
+	} else if err != nil {
+		errs = errors.Join(errs, err)
 	}
 
+	// Try to get container ID from ECS metadata if available
 	if metadataURI := os.Getenv(ecsMetadataEnvV4); metadataURI != "" {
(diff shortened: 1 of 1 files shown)


pkg/host/info.go
```diff
@@ -203,9 +222,34 @@ func (i *Info) releaseInfo(ctx context.Context, osReleaseLocation string) (*v1.R
...
+
+			return "", errEcs
+		}
+	}
+
+	if errMount != nil {
+		return "", errMount
+	}
+
+	return "", errors.New("container ID not found")
```
dhurley:
```suggestion
	var errs error
	
	// Try to get container ID from mount info first
	if containerIDMount, err := containerIDFromMountInfo(i.mountInfoLocation); err == nil && containerIDMount != "" {
		return uuid.NewMD5(uuid.NameSpaceDNS, []byte(containerIDMount)).String(), nil
	} else if err != nil {
		errs = errors.Join(errs, err)
	}

	// Try to get container ID from ECS metadata if available
	if metadataURI := os.Getenv(ecsMetadataEnvV4); metadataURI != "" {
		if cid, err := i.containerIDFromECS(ctx, metadataURI); err == nil && cid != "" {
			return uuid.NewMD5(uuid.NameSpaceDNS, []byte(cid)).String(), nil
		} else if err != nil {
			errs = errors.Join(errs, err)
		}
	}

	return "", errs
```

I think this function could be simplified to this to ensure that all errors are logged if something goes wrong

pkg/host/info.go
```diff
@@ -367,3 +412,30 @@ func mergeHostAndOsReleaseInfo(
...
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("metadata endpoint returned status %d", resp.StatusCode)
```
dhurley:
```suggestion
		return "", fmt.Errorf("metadata endpoint %s returned status %d", uri, resp.StatusCode)
```

PR #1498, opened 2026-01-23
Fix nginx-debug binary not found

Akshay2191:
Currently Agent v2 is not able to discover the agent process if it is running in debug mode. Due to which NIM is not able to register and gets trapped in an infinite backoff loop while trying to find the NGINX master process.

This change should fix the issue.

### DIFF
--- a/src/core/environment.go
+++ b/src/core/environment.go
@@ -606,6 +606,7 @@ func (env *EnvironmentType) Processes() (result []*Process) {
 		cmd, _ := p.CmdlineWithContext(ctx)
 
 		if env.isNginxProcess(name, cmd) {
+			log.Infof("-------------------- %v", name)
 			nginxProcesses[pid] = p
 		}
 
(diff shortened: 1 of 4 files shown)


src/core/environment.go
```diff
@@ -656,7 +656,12 @@ func (env *EnvironmentType) Processes() (result []*Process) {
 }
 
 func (env *EnvironmentType) isNginxProcess(name string, cmd string) bool {
-	return name == "nginx" && !strings.Contains(cmd, "upgrade") && strings.HasPrefix(cmd, "nginx:")
+	base := filepath.Base(name) // strip path if any
+
+	// Accept both the normal nginx binary and the debug binary
+	isNginxBinary := base == "nginx" || base == "nginx-debug"
+
+	return isNginxBinary && !strings.Contains(cmd, "upgrade") && strings.HasPrefix(cmd, "nginx:")
```
dhurley: when in NGINX is in debug I think the `cmd` would start with `{nginx-debug} nginx:`

PR #1522, opened 2026-02-12
fix: Ensure stream closed

mattdesmarais:
### Proposed changes

This PR fixes the file upload logic in `FileServiceOperator` to ensure that when sending files over 1MB (using chunked streaming), the agent closes the gRPC stream and waits for the NGINX One Console response only after all file chunks have been sent.

### Details

- Previously, the stream might not have been properly closed after sending all chunks, potentially causing incomplete uploads or missing server responses.
- Now, after all chunks are sent in `sendFileUpdateStreamChunks`, the agent calls `CloseAndRecv()` on the stream client, ensuring the stream is closed and the server response is received before proceeding.

### Impact

- Ensures reliable file uploads for large files.
- Prevents issues with incomplete or unacknowledged uploads.

+ @nginx-nickc


aphralG (approved):

aphralG:
recheck

PR #1551, opened 2026-03-04
Resolve pinned dependencies

Akshay2191:
### Proposed changes

I have fixed all the Pinned Dependencies which are doable, the remaining ones are in Test Docker file.
These test Dockerfiles use dynamic base images (${BASE_IMAGE}) by design and cannot be pinned without breaking test flexibility.

I have also added a change in the workflow so that we can run the dependencies test on any branch(for testing purpose).


craigell (approved):

aphralG (approved):

Akshay2191:
I will open a new PR for this.

PR #1565, opened 2026-03-19
Add Renovate config for dev-v2 branch

Akshay2191:
Adding Renovate config for dev-v2 branch


craigell (approved):

aphralG:
I could be wrong but I think we can just add dev-v2 to the list of branches in the config in main

Akshay2191:
You are right, even I was not sure how this will work out of dev-v2 branch as the renovate runs through default branch only.

PR #1576, opened 2026-03-25
Add security violations filter processor, and plug it in to the default OTLP logs pipeline when TcplogReceiver is configured

kamalchaturvedi:
### Proposed changes

**COMMIT 1:**

Add a new security violations filter processor that validates the first message as a one-time gate — on the very first log record, it checks:
    - The body is a string (not int, bytes, etc.).
    - The body has exactly 28 pipe-separated fields matching the secops-dashboard-log profile format.
    
    If either check fails, the gate closes permanently — all subsequent messages are dropped with zero overhead (early return before any iteration) until the OTel collector is restarted. If it passes, the gate opens permanently and all future string-bodied records flow through.

**COMMIT 2:**

Add support for NAP V5 (containerized mode) by verifying syslog IP configured against docker0 interface IP as well (along with 127.0.0.1)

**COMMIT 3:**

Plug security violation filter processor into the default logs pipeline, replacing the:
    - logsgzipprocessor: This processor was introduced for NGINX One for a deprecated project, which is not in scope anymore.
    - securityviolationsprocessor: This processor was converting the sylog security violation into a JSON body log record format, and then forwarding it. This deterministix JSON log body conversion is an overhead for the agent to process for every single log record.

### Verification

- Default OTEL pipeline now, once TcplogReceiver is configured by the agent (when `app_protect_security_log` directive has `syslog:server=127.0.0.1:1514` configured)

```
  pipelines:
    metrics/default:
      receivers:
        - hostmetrics
      processors:
        - batch/default_metrics
        - resource/default
      exporters:
        - otlp/default
    logs/default:
      receivers:
        - tcplog/nginx_app_protect
      processors:
        - securityviolationsfilter/default
        - batch/default_logs
        - resource/default
      exporters:
        - otlp/default
```

- In the first log the processor encounters, if valid number of CSV separated fields are sent as per the log-profile configured, the violation is forwarded to the management plane (as per the batch settings). Any subsequent violation is also forwarded.

- In the first log the processor encounters, if the expected CSV separated fields are not sent as per the log-profile configured , the violation is filtered out. Any subsequent violation is also filtered out, and not forwarded by agent.

Here is the error log that can be verified in the `/var/log/nginx-agent/opentelemetry-collector-agent.log` file for that error case:
```
2026-03-25T21:44:42.233Z	error	securityviolationsfilterprocessor/processor.go:102	Security violation log does not appear to be CSV format. Ensure the NAP logging profile uses the secops-dashboard-log format. All security violation logs will be dropped until the collector is restarted.	{"resource": {"service.instance.id": "3ff23ae7-7713-4745-bc28-cfc2198f8eb4", "service.name": "otel-nginx-agent", "service.version": "v3.8.0"}, "otelcol.component.id": "securityviolationsfilter/default", "otelcol.component.kind": "processor", "otelcol.pipeline.id": "logs/default", "otelcol.signal": "logs", "expected_fields": 27, "actual_fields": 1}
github.com/nginx/agent/v3/internal/collector/securityviolationsfilterprocessor.(*securityViolationsFilterProcessor).ConsumeLogs.(*securityViolationsFilterProcessor).ConsumeLogs.ResourceLogsSlice.All.func2.(*securityViolationsFilterProcessor).ConsumeLogs-range1.(*securityViolationsFilterProcessor).ConsumeLogs.(*securityViolationsFilterProcessor).ConsumeLogs.ResourceLogsSlice.All.func2.(*securityViolationsFilterProcessor).ConsumeLogs-range1.ScopeLogsSlice.All.func5.(*securityViolationsFilterProcessor).ConsumeLogs.(*securityViolationsFilterProcessor).ConsumeLogs.ResourceLogsSlice.All.func2.(*securityViolationsFilterProcessor).ConsumeLogs-range1-range4.func6.1
```
### Checklist


internal/config/config.go
```diff
@@ -1257,6 +1252,70 @@ func resolvePipelines() Pipelines {
 	}
 }
 
+func enforceSafeLogBatchProcessors(col *Collector) {
```
dhurley: Not sure if this function should be here.  Hardcoding the values of the batch processor in the code means that we can't change them in the future. Also means if a user wants to send logs to a different exporter they cant use their own log batch processor as it would be overridden by this.

internal/collector/securityviolationsfilterprocessor/factory.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) F5, Inc.
```
dhurley, line 1: Could you add a small README to this folder just to explain what the filter does?

internal/datasource/config/nginx_config_parser.go
```diff
@@ -103,9 +104,50 @@ type createAPIDetailsParams struct {
 func NewNginxConfigParser(agentConfig *config.Config) *NginxConfigParser {
 	return &NginxConfigParser{
```
dhurley, line 111: After calling the `getDocker0IP` function could you add a debug log message like this?
Could help us in the future when debugging.

```
docker0IP := getDocker0IP()
if docker0IP != "" {
    slog.Debug("Running in a docker environment, docker0 interface IP address found")
}

return &NginxConfigParser{
    agentConfig: agentConfig,
    docker0IP:   docker0IP,
}
```

internal/config/config.go
kamalchaturvedi: Hmm. Trying to see what we can do to ensure customer does not set huge values for `batch` processor when talking to N1C. 
In our testing, we found that larger values lead to `rpc error: code = ResourceExhausted desc = grpc: received message larger than max` message returned, which is a permanent error for otel collector (export request is not retried by it).

Any suggestions to handle it safely ?

internal/collector/securityviolationsfilterprocessor/factory.go
kamalchaturvedi: Added in latest commits

internal/datasource/config/nginx_config_parser.go
kamalchaturvedi: Added

internal/config/config.go
dhurley: In our documentation, we do warn users that if they are customising their OTel configuration, overriding default processors can cause issues with sending telemetry to their management plane. https://docs.nginx.com/nginx-one-console/agent/configure-otel-metrics/#example-usage

In the future, we would like to implement a mechanism where the management plane can configure the otel collector remotely so that even if a user overrides a default processor the management plane can undo it by sending their desired configuration when agents connects.
This would allow management planes to change these settings in the future without requiring updates to agent code.

kamalchaturvedi: So do you suggest we remove this safety check, or is it fine as is ?

dhurley: I would suggest removing this check

kamalchaturvedi: Okay removed the commit that introduced that check.

dhurley (approved):

PR #1641, opened 2026-04-21
Improvements to command plugin and service

dhurley:
### Proposed changes

There are a few places where things can be improved in the command plugin and service which are detailed below.

1. The handleSubscribeError function was updated to always reconnect if there is any error not just an unavailable error.
2. A lock was added to the agent config as this would change during runtime and multiple functions could read this while its being updated.
3. The requests in progress check was removed. This check could lead to a scenario where a request never completes when a connection reset is in progress. The check was unnecessary anyways since if there is any gRPC error while processing a request it will eventually recovery once the new connection is established.
4. The monitorSubscribeChannel goroutine exits if an invalid request comes from an auxiliary server. This was fixed to drop the request and continue monitoring the channel.
5. There was a potential deadlock in the config apply request queue when multiple messages were being added to the channel.

### Checklist


internal/command/command_plugin.go
```diff
@@ -157,6 +157,13 @@ func (cp *CommandPlugin) Reconfigure(ctx context.Context, agentConfig *config.Co
 	return err
 }
 
+func (cp *CommandPlugin) getConfig() *config.Config {
```
aphralG:
```suggestion
func (cp *CommandPlugin) config() *config.Config {
```

internal/command/command_service.go
```diff
@@ -634,3 +613,10 @@ func (cs *CommandService) connectCallback(
 		return response, nil
 	}
 }
+
+func (cs *CommandService) getAgentConfig() *config.Config {
```
aphralG:
```suggestion
func (cs *CommandService) agentConfig() *config.Config {
```

internal/command/command_service.go
```diff
@@ -501,41 +485,35 @@ func (cs *CommandService) receiveCallback(ctx context.Context) func() error {
...
-		}
+	cs.isConnected.Store(false)
 
-		return nil
-	}
+	slog.ErrorContext(ctx, fmt.Sprintf("Failed to %s. "+
+		"Trying create connection rpc again", errorMsg), "error", err)
 
-	slog.ErrorContext(ctx, "Failed to "+errorMsg, "error", err)
+	_, connectionErr := cs.CreateConnection(ctx, cs.resource)
```
aphralG, line 500: If we get a subscribe error does it always mean we need a new connection ?

internal/command/command_service.go
```diff
@@ -347,72 +324,76 @@ func (cs *CommandService) handleConfigApplyResponse(
...
-	for i := range indexOfConfigApplyRequest {
-		newResponse := response
+	// Advance the queue past the matched (now handled) entry.
+	cs.configApplyRequestQueue[instanceID] = cs.configApplyRequestQueue[instanceID][indexOfConfigApplyRequest+1:]
+	slog.DebugContext(ctx, "Removed config apply requests from queue", "queue", cs.configApplyRequestQueue[instanceID])
 
-		newResponse.GetMessageMeta().MessageId = id.GenerateMessageID()
+	var pending *mpi.ManagementPlaneRequest
+	if len(cs.configApplyRequestQueue[instanceID]) > 0 {
+		pending = cs.configApplyRequestQueue[instanceID][len(cs.configApplyRequestQueue[instanceID])-1]
```
aphralG: Could the pending name be a bit more descriptive ?

dhurley: sure

internal/command/command_service.go
dhurley: There could be other scenarios besides the server being unavailable where we might want to send a CreateConnection message again. Like we might reestablish our gRPC connection but the management plane has lost its state and doesn't know about us, so a CreateConnection would be required to update the state. 
There are probably other error scenarios where calling CreateConnection again isn't required but it wont do any harm. The management plane can just ignore the requests and respond if it already has that information. We could call CreateConnection only on a certain list of error types but seeing as calling CreateConnection again wont cause any harm I think its simpler to just call CreateConnection again for all errors.

internal/config/flags.go
```diff
@@ -42,7 +42,6 @@ var (
 	ClientGRPCMaxFileSizeKey               = pre(ClientRootKey) + "grpc_max_file_size"
 	ClientGRPCFileChunkSizeKey             = pre(ClientRootKey) + "grpc_file_chunk_size"
 	ClientGRPCMaxParallelFileOperationsKey = pre(ClientRootKey) + "grpc_max_parallel_file_operations"
-	ClientGRPCConnectionResetTimeoutKey    = pre(ClientRootKey) + "grpc_connection_reset_timeout"
```
sjberman, line 45: Is removing this a breaking change?

dhurley: thats a good point. I don't think anyone would be using this config option as we haven't addressed it in our docs but it might be safer to just deprecate it instead of removing.

Akshay2191 (approved):

aphralG (approved):

craigell (approved):

PR #1655, opened 2026-04-23
Bug/fix tls error default fallback

Akshay2191:
Currently, if TLS credentials are misconfigured or unreadable, we silently fall back to insecure credentials without notifying the user.

With this change, if TLS is configured but the certificates are invalid, we log an error and do not add any transport credentials, ensuring the user is aware that their TLS configuration is not working.


internal/grpc/grpc_test.go
```diff
@@ -162,6 +166,25 @@ func Test_GetDialOptions(t *testing.T) {
 			expected:    8,
 			createCerts: false,
 		},
+		{
+			name: "Test 7: TLS configured with invalid certs - no default fallback, token still added",
+			agentConfig: &config.Config{
+				Client: types.AgentConfig().Client,
```
aphralG, line 175: Could a check be added to see if the log is outputted using a log buffer 


```
		logBuf := &bytes.Buffer{}
		stub.StubLoggerWith(logBuf)
```

craigell (approved):

aphralG (approved):

PR #1666, opened 2026-05-06
Resolve pinned dependencies

Akshay2191:
I have fixed all the Pinned Dependencies which are doable, the remaining ones are in Test Docker file. 
These test Dockerfiles use dynamic base images (${BASE_IMAGE}) by design and cannot be pinned without breaking test flexibility so have added them to the ignore list.
I have also added a change in the workflow so that we can run the dependencies test on any branch(for testing purpose).


.github/workflows/scorecards.yml
```diff
@@ -12,6 +12,10 @@ on:
     branches:
       - main
       - dev-v2
+  pull_request:
+    branches:
+      - main
```
aphralG, line 17: why is this only done for main ?

Akshay2191: Thanks for pointing out.

aphralG (approved):

craigell (approved):

PR #1690, opened 2026-05-18
fix: ensure delta metric values are never negative

valyria257:
### Proposed changes

If nginx reset the values of the counters used for these metric values,
then the gauge values will be negative because they are calculated
based on the current value minus the previous value.
    
Since these gauges are tracking counts, it doesn't make sense for them
to be negative. The current value should always be greater than the
previous because it is based on a counter, unless it's been reset to 0.
    
Thus, this will set the gauge values to the current value, most likely
0, if the delta is negative.

### Checklist


internal/collector/nginxplusreceiver/record/http_metrics.go
```diff
@@ -28,7 +28,7 @@ func (hm *HTTPMetrics) RecordHTTPMetrics(stats *plusapi.Stats, now pcommon.Times
 	hm.mb.RecordNginxHTTPRequestsDataPoint(now, int64(stats.HTTPRequests.Total), "", 0)
 
 	// Request Count
-	requestsDiff := int64(stats.HTTPRequests.Total) - int64(hm.PreviousHTTPRequestsTotal)
+	requestsDiff := max(int64(stats.HTTPRequests.Total)-int64(hm.PreviousHTTPRequestsTotal), 0)
```
UnwashedMeme: When I did this in our old nim-agent I introduced a func `delta`, though prometheus calls it [`increase`](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase) and that's probably a better name to show that it can only be positive.

```
// delta calculates the delta value(difference) from monotonically increasing counters.
// delta watches for resets-- if the current value is less than previous value, we take the current value.
func delta[T uint64 | int64](current T, previous T) T {
	if current >= previous {
		return current - previous
	}
	return current
}
```

I prefer named functions, even super simple ones as it carries intent better than repeated "complex" expressions-- this one isn't _that_ complex but still.

It also returns more accurate data, if the counter goes  `0, 38, 24, 50, 70` the resulting increases are either `38, 0, 26, 20`  or `38, 24, 26, 20`.  We don't know whether the counter climbed higher than 38 before resetting, but we don't have to discard the 24 that we know of in the third data point.

valyria257: done!

UnwashedMeme (approved):

internal/collector/nginxplusreceiver/record/http_metrics.go
```diff
@@ -112,3 +112,13 @@ func (hm *HTTPMetrics) RecordHTTPLimitMetrics(stats *plusapi.Stats, now pcommon.
 		)
 	}
 }
+
+// increase calculates the delta value (difference) from monotonically increasing counters.
+// If the current value is less than previous value, in the case of a rest, we take the current value.
```
UnwashedMeme:
```suggestion
// If the current value is less than previous value, in the case of a reset, we take the current value.
```

craigell (approved):

aphralG (approved):

PR #1694, opened 2026-05-25
Added renovate.json from main into dev-v2

Akshay2191:
Adding renovate.json file to dev-v2 branch.


Akshay2191:
Renovate is already running on dev-v2 so this is not required.

PR #1712, opened 2026-06-03
Integration tests improvements

craigell:
### Proposed changes

This pull request reduces the amount of time to run integration tests and adds reliability to runs. 

### Checklist


test/helpers/test_containers_utils.go
```diff
@@ -496,8 +502,33 @@ func ExtractFileFromContainer(
 	containerPath string,
 ) string {
 	tb.Helper()
-	fileContent, err := testContainer.CopyFileFromContainer(ctx, containerPath)
+
```
aphralG, line 518: Could assert.Eventually be used instead ?

karensantana (approved):

.github/workflows/ci.yml
```diff
@@ -58,6 +58,10 @@ jobs:
           run: go install github.com/goreleaser/nfpm/v2/cmd/nfpm@${{ env.NFPM_VERSION }}
         - name: Fix golang dependency permissions
           run: chmod -R 0755 ~/go/pkg/mod ~/.cache/go-build
+        - name: Trim Go build cache
+          run: |
+            go clean -cache
```
Akshay2191: This cleans the full cache, not sure if that is what we are trying to achieve here as the next step is storing it. If we want to trim just the test cache it might be good to use "go clean -testcache" instead.

craigell: I noticed that the cache is getting  high to the threshold so this is a precautionary step.

Akshay2191: Instead of manually managing actions/cache and running go clean -cache (which wipes the build cache to 0 MB before saving and forces full recompilations), we can leverage the built-in caching in actions/setup-go.
According to the [official setup-go documentation](https://github.com/actions/setup-go#caching-package-dependencies) **cache: true** automatically manages both downloaded modules and compiled build objects to keep build times fast while preventing cache bloat.

Akshay2191 (approved):

PR #1757, opened 2026-06-25
Fix agent not logging collector logs to stdout by default in containers

karensantana:
### Proposed changes

When running in a container the agent collector logs are being written to disk instead of stdout. This change fixes this behavior so if a path is not specified in the config, `stdout` is used by default in containers.

### Checklist


craigell:
Think we can append a integration test for this in an existing suite since we use containers for testing.

craigell (approved):

karensantana:
> Think we can append a integration test for this in an existing suite since we use containers for testing.

Perfect, I'll add that. 👍

craigell (approved):

Akshay2191 (approved):

PR #1773, opened 2026-06-30
memoryscraper cleanup

Akshay2191:


internal/collector/containermetricsreceiver/internal/scraper/memoryscraper/internal/cgroup/memory.go
```diff
@@ -104,19 +104,19 @@ func (ms *MemorySource) VirtualMemoryStatWithContext(ctx context.Context) (*mem.
...
 
 	cgroupStat.Total = memoryLimitInBytes
-	cgroupStat.Available = memoryLimitInBytes - usedMemory
+	cgroupStat.Available = saturatingSub(memoryLimitInBytes, usedMemory)
 	cgroupStat.Used = usedMemory
 	cgroupStat.Cached = memoryStat.cached
 	cgroupStat.Shared = memoryStat.shared
 	cgroupStat.UsedPercent = usedMemoryPercent
-	cgroupStat.Free = memoryLimitInBytes - usedMemory
+	cgroupStat.Free = saturatingSub(memoryLimitInBytes, usedMemory)
```
karensantana, line 119: comment(non-blocking): I know this was not introduced in this PR, however it is interesting that `Free` and `Available` are holding the same value (and the metrics tests are passing). According to AI these values are not the same.  🤔

karensantana (approved):

craigell (approved):

internal/collector/containermetricsreceiver/internal/scraper/memoryscraper/internal/cgroup/memory.go
Akshay2191: Good catch. Free currently duplicates Available (limit - (usage - cache)). It should be computed from total usage (limit - usage) while Available can remain cache-adjusted. Existing tests pass because they validate current behavior; they should be updated alongside the fix.

PR #1774, opened 2026-06-30
propagate RunE errors to main

Akshay2191:


cmd/agent/main.go
```diff
@@ -52,5 +52,6 @@ func main() {
 	err := app.Run(ctx)
 	if err != nil {
 		slog.ErrorContext(ctx, "NGINX Agent exiting due to error", "error", err)
+		os.Exit(1)
```
karensantana, line 55: question(non-blocking): I was wondering, do we prefer exiting immediately  instead of signaling for a graceful shutdown? I see we have a graceful shutdown handler already (Line 32).

Akshay2191: Good question — this error path is pre-startup (before the long-running message loop begins), so there’s no in-flight work to gracefully drain.
We intentionally use os.Exit(1) here to surface startup failure correctly (instead of a clean/cancelled exit).

karensantana: Perfect, that makes sense, thanks for explaining! 👍

karensantana (approved):

craigell (approved):

PR #1837, opened 2026-07-23
fix: reject non-NGINX processes, classify nginx master by cmdline role instead of PPID (dev-v2)

ShubhenduSinghF5:
EnvironmentType.Processes() inferred IsMaster by checking whether a candidate nginx process's PPID was itself in the discovered nginx-process set — any nginx-named process whose parent was not another nginx process was flagged as a master.

This breaks when another tool injects a helper process that borrows the nginx: cmdline convention and runs under PID 1. A concrete case observed in the field: Dynatrace OneAgent injects a [nginx: OneAgent companion process] (name nginx, ppid 1), which the PPID heuristic classifies as a second master. Downstream, UpdateNginxDetailsFromProcesses keys nginxDetailsMap by the derived nginxId, both "masters" collide on the same key, and Go's non-deterministic map iteration decides which PID the agent sends the reload SIGHUP to. When the impostor wins, nginx -s reload silently no-ops while the agent still reports success.

Align v2's nginx master-process classification with the approach already used in NGINX Agent v3 [pkg/nginxprocess.Process.IsMaster] identify the master by the nginx-reported role prefix on the process command line rather than by parent-PID inference.

### Proposed changes

Add [isNginxMaster(cmd string) bool] — mirrors the v3 implementation, checks for the nginx: master / {nginx-debug} nginx: master cmdline prefix.
Use it in Processes() in place of the PPID-based inference.
[isNginxProcess] (the collection filter) is unchanged; only role classification moves from PPID to cmdline.

### Checklist

### DIFF
--- a/src/core/environment.go
+++ b/src/core/environment.go
@@ -597,12 +597,8 @@ func (env *EnvironmentType) Processes() (result []*Process) {
 		user, _ := nginxProcess.UsernameWithContext(ctx)
 		ppid, _ := nginxProcess.PpidWithContext(ctx)
 		cmd, _ := nginxProcess.CmdlineWithContext(ctx)
-		isMaster := false
 
-		_, ok := nginxProcesses[ppid]
-		if !ok {
-			isMaster = true
-		}
+		isMaster := isNginxMaster(cmd)
 
 		var exe string
 		if isMaster {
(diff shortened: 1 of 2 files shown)


ShubhenduSinghF5:

balakoteswar (approved):

devbgv (approved):

PR #1876, opened 2026-08-11
Unit test coverage expansion

craigell:
### Proposed changes

Increasing the unit test coverage of existing and recently added functionality. 

### Checklist


internal/config/mapper_test.go
```diff
@@ -303,9 +295,167 @@ func TestMapper_ToAgentConfigLogProto(t *testing.T) {
...
+				Auth:   nil,
+			},
+			wantHost: "127.0.0.1",
+			wantPort: 8888,
+			wantType: mpi.ServerSettings_SERVER_SETTINGS_TYPE_GRPC,
+			wantAuth: false,
+		},
+	}
+
+	for _, tt := range tests {
```
karensantana, line 358: issue: I believe there is a potential concurrency issue in this block of code due to shared test table data (tt) being accessed within a t.Parallel() call.
More safe approach:
`for _, tt := range tests {
    tt := tt // capture
    t.Run(tt.name, func(t *testing.T) {....`

internal/config/mapper_test.go
```diff
@@ -303,9 +295,167 @@ func TestMapper_ToAgentConfigLogProto(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			result := ToAuxiliaryCommandServerProto(tt.cmd)
+			require.NotNil(t, result)
```
karensantana, line 362: question: If `tt.cmd.Server == nil` will `require.NotNil(t, result)` succeed?

craigell: This is the auxillary command server, so it should work since it's not the main command server.

internal/config/mapper_test.go
craigell: I can remove the parallelism for this section of tests if we believe it could cause an issue.

internal/config/mapper_test.go
karensantana: Thanks for clarifying!

internal/config/mapper_test.go
karensantana: In my opinion, there is no need to remove the parallelism, we could just add this to prevent any issues:
```
for _, tt := range tests {
    tt := tt // Capture loop variable
    t.Run(tt.name, func(t *testing.T) {
        t.Parallel()
```

karensantana: You can disregard the entire comment, this was indeed an issue, but for previous versions of Go, this is safe to go as it is, here is a blog post that explain it:
https://medium.com/@krisguttenbergovitz/go-1-22s-loop-variable-fix-solving-a-decade-old-gotcha-and-modern-concurrency-pitfalls-69aa4eb0b8a1

karensantana (approved):

Akshay2191 (approved):

PR #1888, opened 2026-08-18
feat: add certificate expiry metric to agent for NGINXaaS 

vivki:
### Motivation

Platform engineers managing NGINXaaS deployments need alerting before certificates expire. This metric integrates into the existing monitoring stack and includes labels (`subject.common_name`, `file_path`, `serial_number`, `public_key_algorithm`) to easily identify the expiring cert.

nginx-agent already discovers every certificate NGINX uses during config parsing. This change exports that data as a time-series metric, giving operators a simple alert on `nginx.ssl.certificate.expiry` without additional tooling.

The receiver is separate from nginx/nginxplus receivers because it covers a distinct concern (TLS hygiene vs. traffic metrics) and can be enabled or disabled independently.

### Acceptance Criteria

- [x] **Performance:** the cert receiver parses each cert once at startup (`Start()`); `Scrape()` only emits cached data with no per-interval file I/O. Renewed certs are picked up by restarting the receiver on the next config apply.
- [x] **Security:** no private keys or raw cert bytes are transmitted; the only metadata exposed is `file_path`, `public_key_algorithm`, `serial_number`, and `subject.common_name`
- [x] **Opt-in:** disabled by default; enabled by adding `certificate_metrics` under `collector.receivers` in nginx-agent.yaml. Scrape interval is configurable via `collection_interval` (default 15s).

### Sequence Diagram

```mermaid
sequenceDiagram
    participant MP as Management Plane
    participant Agent as nginx-agent
    participant Plugin as OTel Collector Plugin
    participant Receiver as Certificate Receiver
    participant Scraper as Certificate Scraper

    MP->>Agent: Push NGINX config (with certs)
    Agent->>Agent: Parse config, discover ssl_certificate files
    Agent->>Plugin: NginxConfigUpdateTopic(NginxConfigContext)

    Plugin->>Plugin: Check CertificateMetrics != nil
    Plugin->>Plugin: updateCertificateReceivers()
    Note over Plugin: Extract cert paths from Files via CertificateMeta
    Note over Plugin: Always restart receiver to pick up renewals
    Plugin->>Plugin: Restart OTel Collector

    Plugin->>Receiver: Start()
    Receiver->>Scraper: Start(ctx)
    loop For each cert file path
        Scraper->>Scraper: cert.LoadCertificate(path)
        Scraper->>Scraper: Cache parsed x509.Certificate
    end

    loop Every collection_interval (default 15s)
        Plugin->>Scraper: Scrape(ctx)
        Scraper->>Scraper: Read cached certs (no file I/O)
        loop For each cached cert
            Scraper->>Scraper: Compute time-to-expiry
            Scraper->>Scraper: Emit nginx.ssl.certificate.expiry gauge
        end
        Scraper-->>Plugin: pmetric.Metrics
    end

    MP->>Agent: Push updated config (cert renewed in-place)
    Agent->>Plugin: NginxConfigUpdateTopic
    Plugin->>Plugin: updateCertificateReceivers() -> restart
    Plugin->>Receiver: Shutdown() -> Start()
    Receiver->>Scraper: Start(ctx)
    Note over Scraper: Re-reads cert files, refreshes cache

```

### Implementation

Adds a `certificate` OTel receiver that emits `nginx.ssl.certificate.expiry`: a gauge of seconds until each cert expires (negative when expired).

- **Scraper:** uses `cert.LoadCertificate` (single leaf cert per file, matching the management plane's model); certs are parsed once at `Start()` and cached in memory. `Scrape()` emits cached data with no per-interval file I/O.
- **Attributes:** `file_path`, `public_key_algorithm`, `serial_number` (decimal), `subject.common_name`
- **Resource attribute:** `instance.id`
- **Gated on:** presence of `certificate_metrics` key under `collector.receivers` (`Receivers.CertificateMetrics != nil`); disabled by default
- **Configurable interval:** `collection_interval` under `collector.receivers.certificate_metrics` (default 15s)
- **Restart behaviour:** collector always restarts the cert receiver on config apply, ensuring in-place cert renewals are picked up. If certs are rotated externally without a config push, stale data persists until the next reload.

### Commits

| # | Summary |
|---|---------|
| 1 | Add stability annotations to existing receiver metadata |
| 2 | Define certificate receiver metadata.yaml and config types |
| 3 | Generate boilerplate via mdatagen |
| 4 | Implement scraper, factory, wiring, and tests |
| 5 | Gate on `certificate_metrics` config presence |


internal/collector/otel_collector_plugin.go
```diff
@@ -32,9 +33,10 @@ import (
 )
 
 const (
-	maxTimeToWaitForShutdown  = 30 * time.Second
-	defaultCollectionInterval = 1 * time.Minute
-	filePermission            = 0o600
+	maxTimeToWaitForShutdown      = 30 * time.Second
+	defaultCollectionInterval     = 1 * time.Minute
+	defaultCertCollectionInterval = 15 * time.Second
```
UnwashedMeme: Does `defaultCertCollectionInterval` need to be shorter than the `defaultCollectionInteval`? I don't expect the certificate data to change very rapidly.

internal/collector/otel_collector_plugin.go
```diff
@@ -542,6 +544,10 @@ func (oc *Collector) checkForNewReceivers(ctx context.Context, nginxConfigContex
 		slog.DebugContext(ctx, "NAP logs feature disabled", "enabled_features", oc.config.Features)
 	}
 
+	if oc.config.IsFeatureEnabled(pkgConfig.FeatureCertificates) {
+		reloadCollector = reloadCollector || oc.updateCertificateReceivers(nginxConfigContext)
```
UnwashedMeme, line 547: I'm not familiar with the "Feature" system in here. Looking at it we have a feature for metrics. Is certificate metrics a separate feature?  

I had been assuming that  it was just if  `CertificateReceivers` config info wasn't in nginx-agent.conf then that receiver wouldn't be loaded.  Maybe something like getting rid of `defaultCertCollectionInterval` and only if oc.config.Collector.Receivers.CertScrapeInterval is nonzero do we add that receiver.  Then in order to use it I have to add to my agent config a block like

```yaml
collector:
  receivers:
    CertScrapeInterval: 1m   # this is what activates it, without it, no cert scraping.
```

internal/collector/certificatereceiver/scraper.go
```diff
@@ -0,0 +1,127 @@
...
+	if c.cfg.InstanceID != "" {
+		c.rb.SetInstanceID(c.cfg.InstanceID)
+	}
+
+	for _, path := range c.cfg.CertFilePaths {
+		if err := c.refreshCacheEntry(path); err != nil {
+			slog.WarnContext(ctx, "Failed to load certificate file, skipping", "path", path, "error", err)
+
+			continue
+		}
```
UnwashedMeme: I don't think we re-check here on the per-scrape interval. I think that could violate both our performance a/c (at most one re-parse) and our correctness a/c.  If the file changes on disk (new mtime) that doesn't mean it has been loaded into NGINX yet (config reload not yet happened or failing). We want the metric data to reflect what NGINX has loaded, not what the file on disk is.

I think the only time we read new file data is Start/updateCertificateReceivers -- right when agent is saying "the config has been reloaded, these are the new file paths.

internal/collector/otel_collector_plugin.go
```diff
@@ -763,6 +769,77 @@ func (oc *Collector) doesTcplogReceiverAlreadyExist(listenAddress string) bool {
...
+	slices.Sort(sortedA)
+	slices.Sort(sortedB)
+
+	return slices.Equal(sortedA, sortedB)
+}
+
+func certFilePathsFromFiles(files []*v1.File) []string {
+	var paths []string
+	for _, f := range files {
+		if _, ok := f.GetFileMeta().GetFileType().(*v1.FileMeta_CertificateMeta); ok {
```
UnwashedMeme: We talked some about using the data here (despite/because it shows only the first cert to match the management plane).

I suspect you're running into issues getting that data through the config+templating?

internal/collector/otel_collector_plugin.go
vivki: This makes sense. I'll change the implementation to gate on interval specification. Thanks!

internal/collector/certificatereceiver/scraper.go
vivki: Just thinking through this change, if certs are parsed once at Start() and cached in memory, Scrape() only emits the cached values. If the file content changes (e.g., cert renewed in-place) but the path stays the same, the scraper will keep reporting the original cert's metadata from when Start() ran. It won't detect or reload the updated file. Agent team did specifically mention file change despite file path being unchanged as a concern. Think a solution through

vivki: I think the simplest solution is to have the cert receiver restart on every reload.

internal/collector/otel_collector_plugin.go
valyria257: This would follow closer to the existing patterns:

The nginx-agent config the user provides would look something like:

```
collector:
  receivers:
    certificate_metrics:
      collection_interval: 30s
```

Then, the otel collector template would have something like this: 

```
{{- if ne .Receivers.CertificateMetrics nil }}
  certificate_metrics:
    {{- if .Receivers.CertificateMetrics.CollectionInterval }}
    collection_interval: {{ .Receivers.CertificateMetrics.CollectionInterval }}
    {{- end }}
{{- end }}
```

So, if `certificate_metrics` is not set, then the resulting otel collector component won't exist in the config. If `certificate_metrics` is set, but with no collection_interval, then the default will be used.

internal/collector/otel_collector_plugin.go
vivki: Yep, passing CertificateMeta through the config+templating gets awkward since the OTel receiver config is flat YAML and the proto objects are nested/typed. It seems simpler to pass file paths through the template and let the scraper read them once at Start, as you suggested in your other comment. If the collector restarts on every reload/successful config apply, the cache stays "fresh."

internal/collector/otel_collector_plugin.go
vivki: Defined certificate receiver as below:

```
{{- range .Receivers.CertificateReceivers }}
{{- if gt (len $.Receivers.CertificateReceivers) 1 }}
  certificate/{{- .InstanceID -}}:
{{- else }}
  certificate:
{{- end}}
    instance_id: "{{- .InstanceID -}}"
    {{- if .CollectionInterval }}
    collection_interval: {{ .CollectionInterval }}
    {{- end }}
    {{- if .CertFilePaths }}
    cert_file_paths:
      {{- range .CertFilePaths }}
      - "{{- . -}}"
      {{- end }}
    {{- end }}
{{- end }}
```

This block should be handle the collection interval nil check with {{- if .CollectionInterval }}.

internal/collector/certificatereceiver/scraper.go
vivki: Done; updated the cert receiver based on your feedback--thank you.

Removed mtime-based re-parsing: certs are now parsed once at Start() and Scrape() only emits cached values. No per-interval file reading.

Altered certificate receiver to always restart on reload/config apply unconditionally returns true. This means in-place cert renewals (same path, new content) get picked up on the next config apply without needing mtime checks.

Akshay2191: The OTel collector plugin already receives `NginxConfigUpdateTopic` which carries the full `NginxConfigContext` including `CertificateMeta` (already parsed, including `SerialNumber`) 

Rather than the current unconditional restart, `updateCertificateReceivers` could compare the serial numbers of the incoming certs against what the receiver is currently tracking. By doing this a restart is only triggered when a serial number changes or a path is added/removed — covering both path changes and in-place cert renewal. When nothing has changed, no restart happens and the existing cache continues serving the metric. This requires no new topic subscriptions and no changes to the scraper itself.

internal/collector/otel_collector_plugin.go
UnwashedMeme: In the end all of this becomes labels a flat set of strings with known keys per filepath. It seems like we should be able to shuffle it through?

Can we make the config in internal/collector/certificatereceiver/config.go:

```golang
type Config struct {
	InstanceID                     string                        `mapstructure:"instance_id"`
	CertFilePaths                  map[string]CertInfo           `mapstructure:"cert_file_paths"`
	MetricsBuilderConfig           metadata.MetricsBuilderConfig `mapstructure:",squash"`
	scraperhelper.ControllerConfig `mapstructure:",squash"`
}
type CertInfo struct {
	CommonName   string `mapstructure:"common_name"`
	SerialNumber string `mapstructure:"serial_number"`
	// ...
}
```

And then when we're templating is something vaguelly like:
```
    cert_file_paths:
      {{- range $path, $data := .CertFilePaths }}
        {{ $path }}: {{ $data | as_yaml}}
      {{- end }}
```

internal/collector/certificatereceiver/scraper.go
vivki: Hi Akshay, thanks for the feedback. I made some changes based on your and Nathan's suggestions. 

1. CertMeta (serial, common name, algorithm, expiry) is now extracted from the existing CertificateMeta proto and passed through the OTel config template. The scraper reads it from config directly and no longer calls LoadCertificate.
2. updateCertificateReceivers now uses maps.Equal on a map[string]CertMeta (path as key, serial number in value) so it only restarts when cert paths or serial numbers actually change. (Technically this means changes in any cert meta value will trigger a restart.)

internal/collector/otel_collector_plugin.go
vivki: Pasting comment I made above describing updates following feedback:

1. CertMeta (serial, common name, algorithm, expiry) is now extracted from the existing CertificateMeta proto and passed through the OTel config template. The scraper reads it from config directly and no longer calls LoadCertificate.
2. updateCertificateReceivers now uses maps.Equal on a map[string]CertMeta (path as key, serial number in value) so it only restarts when cert paths or serial numbers actually change. (Technically this means changes in any cert meta value will trigger a restart.)

internal/collector/certificatereceiver/scraper_test.go
```diff
@@ -0,0 +1,144 @@
...
+		CertMeta:             nil,
+	}
+	scraper := newTestScraper(t, cfg)
+
+	metrics, err := scraper.Scrape(context.Background())
+	require.NoError(t, err)
+	require.Equal(t, 0, metrics.ResourceMetrics().Len(), "Should return empty metrics when no certs configured")
+}
+
+func TestScrape_MultipleCerts(t *testing.T) {
```
craigell, line 117: Can an additional expired cert be included? It'll cover the scraper with both valid and expired certs.

internal/collector/otel_collector_plugin.go
UnwashedMeme: `oc.config.Collector.Receivers.CertificateMetrics.CollectionInterval` isn't a pointer, so it's default value will be a zero Duration.

`{{- if .CollectionInterval }}` is going to have a non-nil value (0).  I think that will be false,  have we checked that specifically?

If the config doesn't specify a collection interval, then the default from `certificatereceiver.defaultCollectInterval` will be used. That's currently 15s which I think is too aggressive.  This is a metric whose value rarely changes and the default for all the other metrics is 1m.

Can we make the certificatereceiver's default collection interval match the rest of the metrics?

internal/collector/otel_collector_plugin.go
```diff
@@ -542,6 +543,10 @@ func (oc *Collector) checkForNewReceivers(ctx context.Context, nginxConfigContex
 		slog.DebugContext(ctx, "NAP logs feature disabled", "enabled_features", oc.config.Features)
 	}
 
+	if oc.config.Collector.Receivers.CertificateMetrics != nil {
```
UnwashedMeme, line 546: During a reload can the `oc.config.Collector.Receivers.CertificateMetrics` have gone from not-nil to nil? 

Is an elseif clause like:
```golang
	} else if oc.config.Collector.Receivers.CertificateReceivers != nil {
		oc.config.Collector.Receivers.CertificateReceivers = nil
		reloadCollector = true
	}
```
a good idea to catch the case of "we used to have certificate receivers, but the customer has turned that off so we want to be sure to remove those receivers?"

internal/collector/certificatereceiver/scraper_test.go
craigell: Is there a test where we update the expired cert to be valid?

internal/collector/otel_collector_plugin.go
vivki: > {{- if .CollectionInterval }} is going to have a non-nil value (0). I think that will be false, have we checked that specifically?

Yes, {{- if .CollectionInterval }} is false when CollectionInterval is zero (Go templates treat zero Duration as false). So it won't render the line, and the receiver uses its own defaultCollectInterval. Added TestTemplateWrite_ZeroCollectionIntervalOmitted confirming this.

> Can we make the certificatereceiver's default collection interval match the rest of the metrics?

(How dare you question the original A/C?) Done; set default interval for cert receiver to 1m.

> During a reload can the oc.config.Collector.Receivers.CertificateMetrics have gone from not-nil to nil?

Good point, I added the else if clause. Used len > 0 instead of != nil since it's a slice.

Pushed changes to address feedback in newest commit. Thanks again

internal/collector/certificatereceiver/scraper_test.go
vivki: Thanks for pointing that out. Added an expired cert to TestScrape_MultipleCerts, and added TestScrape_RenewExpiredCert to cover both cases. (pushed to latest commit)

internal/collector/otel_collector_plugin.go
vivki: Pasting response in another comment here:

> {{- if .CollectionInterval }} is going to have a non-nil value (0). I think that will be false, have we checked that specifically?

Yes, {{- if .CollectionInterval }} is false when CollectionInterval is zero (Go templates treat zero Duration as false). So it won't render the line, and the receiver uses its own defaultCollectInterval. Added TestTemplateWrite_ZeroCollectionIntervalOmitted confirming this.

> Can we make the certificatereceiver's default collection interval match the rest of the metrics?

(How dare you question the original A/C?) Done; set default interval for cert receiver to 1m.

internal/collector/otelcol.tmpl
```diff
@@ -271,7 +293,7 @@ service:
 
   pipelines:
     {{- range $pipelineName, $pipeline := .Pipelines.Metrics }}
-      {{- if or (ne $.Receivers.HostMetrics nil) (ne $.Receivers.ContainerMetrics nil) (gt (len $.Receivers.OtlpReceivers) 0) (gt (len $.Receivers.NginxReceivers) 0) (gt (len $.Receivers.NginxPlusReceivers) 0) }}
+      {{- if or (ne $.Receivers.HostMetrics nil) (ne $.Receivers.ContainerMetrics nil) (gt (len $.Receivers.OtlpReceivers) 0) (gt (len $.Receivers.NginxReceivers) 0) (gt (len $.Receivers.NginxPlusReceivers) 0) (gt (len $.Receivers.CertificateReceivers) 0) }}
```
UnwashedMeme, line 296: (tiny nit)
Does templating treat zero length list as false?  Coudl this just be `or $.Receivers.HostMetrics $.Receivers.ContainerMetric $.Receivers.OtlpReceivers ....`

or maybe `or ... (gt (add (len $.Receivers.OtlpReceivers) (len  $.Receivers.NginxReceivers) ... ))

UnwashedMeme (approved):

internal/collector/certificatereceiver/config.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const defaultCollectInterval = 1 * time.Minute
+
+type Config struct {
+	InstanceID                     string                        `mapstructure:"instance_id"`
+	CertMeta                       map[string]CertMeta           `mapstructure:"cert_meta"`
+	MetricsBuilderConfig           metadata.MetricsBuilderConfig `mapstructure:",squash"`
+	scraperhelper.ControllerConfig `mapstructure:",squash"`
+}
```
Akshay2191, line 23: **Not a blocker** - but `CertMeta` is now defined in two places: `internal/config/types.go` and `internal/collector/certificatereceiver/config.go`. They have identical fields but there is no compile-time enforcement keeping them in sync. If a field is added to one and missed in the other, the data silently drops from the metric. If the OTel receiver is intentionally kept self-contained, then that's fine — but it would be worth adding a comment on both structs pointing to the other. If self-containment is not a hard requirement, then receiver's `config.go` could import and use `config.CertMeta`.

internal/collector/otelcol.tmpl
vivki: I _think_ it would pass as true for non-nil but empty. (Though it's an edge case that shouldn't happen, in practice these fields are either nil or populated. len > 0 feels like clutching a pitchfork, but I can be sure of its protection?)

And at the end of the day I'm hewing more towards "don't poke the bear/existing code pattern"

internal/collector/certificatereceiver/config.go
vivki: Good point. I don't think self-containment is necessary, so I updated `CertMeta` to be imported from `config` as you suggested.

Akshay2191 (approved):

craigell (approved):

